PR #34, opened 2015-02-28
Remove unused HTML attribute.

whyjustin:
Reformat included. View non-whitespace changes https://github.com/sonatype/nexus-book/pull/34/files?w=1.

### DIFF
--- a/site/book-template.html
+++ b/site/book-template.html
@@ -16,72 +16,70 @@
 </head>
 
 <body>
-<div dashboard-header="">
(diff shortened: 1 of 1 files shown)


mosabua:
Looks good to me. We have to port over to the Nexus book as well. Should we track via a JIRA issue? Do we want to port to other branches? If it affects look and feel we probably should... wdyt @jwayman

whyjustin:
> If it affects look and feel we probably should

This should not affect anything, the HTML attribute is cruft.

PR #36, opened 2015-03-19
Server correction

joedragons:
I believe this correction is right, please check.


mosabua:
;-) thx .. I will push it live soon

PR #37, opened 2015-03-20
Suggested strike+

joedragons:
The testguide specifically says you cannot assign privileges to users, that you have to assign the privilidges to a role which then you can assign to a user.  My test of this showed this to be true.  While just a wording difference, I think not having privilege lessens the chance of confusion and IMO is better than the alternative which would be to explain this verbosely.

### DIFF
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -1628,7 +1628,7 @@ implicit capture in the Staging Suite as documented in <<staging>>.
 [[security-roles]]
 === Managing Roles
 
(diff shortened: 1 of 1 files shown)


joedragons:
Meh I saw a typo too, adding it in hoping this is accepted.

joedragons:
Note, other places in the doc CRUD is written as C.R.U.D.  I did not adjust that as it does not make much difference to me (moreso that the first use is spelled out what it is) but it's currently not consistent.

PR #39, opened 2015-04-01
Version font consistancy

joedragons:
I noticed that version 1.2 had ++ markup and version 2.0 does not.  I removed the markup but am doing a PR.  If you disagree, please consider then adding it to 2.0 and making this consistent the reverse way.


mosabua:
Generally when referencing values from the command line or display I use + or ` (they are equivalent in the result). When referencing labels and buttons and such from the UI I use ' . So in this case .. I will `

PR #50, opened 2015-05-01
new UI semantics documented and updated

mosabua:
https://issues.sonatype.org/browse/NEXUS-8514


mosabua:
@djsauble can you please review this so we can establish a common naming of things..

mosabua:
CI build running at http://bamboo.s/browse/BOOK-NEXUS17-1

mosabua:
Here is a rendered version http://bamboo.s/browse/BOOK-NEXUS17-1/artifact/STAGE/site/3.0/reference/using-sect-intro.html

djsauble:
Will review this by EOD.

djsauble:
Why “menu icon” instead of “toolbar icon”?

https://github.com/sonatype/nexus-book/pull/50/files#diff-f247b168877cbbd993d8041315e4527bR33

You refer to the header as the “main toolbar” elsewhere, so I’d expect to see that language when referring to the “Browse”, “Administration”, and “User” icons.

chapter-using-nexus-ui.asciidoc
```diff
@@ -45,9 +45,9 @@ Main toolbar::
 The top of the page contains the header with a number of elements
 starting on the left with the Nexus logo:
 
-Nexus logo and version;; The Nexus logo and version display differs for
```
djsauble: What’s the difference between a “display” and a “label”?

Looks like you use “label” to refer to links elsewhere. Is that a deliberate choice?

chapter-using-nexus-ui.asciidoc
```diff
@@ -78,13 +78,13 @@ the main menu on the left.. In addition the sign out button
 will be displayed.
 
 
-Help button image:figs/web/ui-help-button-icon.png[scale=50];; Clicking the
- help button opens up the help menu.  It contains a link to specific
- help about the currently active feature view. The 'About' menu items
- displays a window with details about the Nexus version as well as
```
djsauble: Might be worth distinguishing between new browser windows and modal dialog windows. We use both in the UI.

chapter-using-nexus-ui.asciidoc
```diff
@@ -45,9 +45,9 @@ Main toolbar::
 The top of the page contains the header with a number of elements
 starting on the left with the Nexus logo:
 
-Nexus logo and version;; The Nexus logo and version display differs for
-{oss} and {pro} and allows you to know what
-version of Nexus you are accessing at a glance.
+Nexus logo and version;; The Nexus logo and version display differs
```
djsauble: What’s the difference between a “display” and a “label”?

Looks like you use “label” to refer to links elsewhere. Is that a deliberate choice?

chapter-using-nexus-ui.asciidoc
```diff
@@ -78,13 +78,13 @@ the main menu on the left.. In addition the sign out button
...
- help button opens up the help menu.  It contains a link to specific
- help about the currently active feature view. The 'About' menu items
- displays a window with details about the Nexus version as well as
- license and copyright information. The 'Documentation', 'Support' and
- 'Issue Tracker' link to the respective sections on the Sonatype
- website.
+Help button image:figs/web/ui-help-button-icon.png[scale=50];;
+ Clicking the help button opens up the help menu.  It contains a link
+ to specific help about the currently active feature view. The 'About'
+ menu items displays a window with details about the Nexus version as
```
djsauble: May be worth distinguishing between new browser windows and modal dialog windows. We use both in the UI.

chapter-using-nexus-ui.asciidoc
```diff
@@ -78,13 +78,13 @@ the main menu on the left.. In addition the sign out button
...
+ well as license and copyright information. The 'Documentation',
+ 'Knowledge base', 'Community', 'Issue Tracker' and 'Support' items
```
djsauble: You use **menu item** to describe “About”, and **items** to describe “Knowledge base”, “Community”, “Issue Tracker”, and “Support”. Should be consistent here.

chapter-using-nexus-ui.asciidoc
```diff
@@ -78,13 +78,13 @@ the main menu on the left.. In addition the sign out button
...
+ link to the respective sections on the Sonatype website.
```
djsauble: You use **sections** to describe “areas of the Nexus UI” and “pages on the documentation site”. Possibly redundant.

chapter-using-nexus-ui.asciidoc
```diff
@@ -102,22 +102,28 @@ display of the respective feature view in the feature view panel.
...
-<<fig-ui-example>> shows a typical user
-interface appearance of Nexus with the 'Repositories' feature view in
-the feature view panel. It shows a list of repositories with the
-current selection 'Central' highlighted. The panels underneath the list show
-details for the selected list item. 
+the main menu initially displays the 'Welcome' feature view. It
+changes display based on your selected item in the main menu.
+
+<<fig-ui-example>> shows a typical user interface appearance of Nexus
+with the 'Users' feature view in the feature view panel. It shows a
```
djsauble: Curious. Why “feature view” instead of just “feature”? Is the latter redundant in some way?

chapter-using-nexus-ui.asciidoc
```diff
@@ -21,8 +21,8 @@ using Nexus to avoid an unsatisfactory user experience.
```
joedragons: I know our support says IE9-11 but I heard NL back-supported a bunch of customers who are locked to IE8 so I'm nervous saying it's blocked unless that's true.  Just putting that out there (and aware not a change made in this revision=)).

chapter-using-nexus-ui.asciidoc
joedragons: RE: "The Nexus logo and version display differs"
In case Daniel's comment doesn't spur a change in this, the logo also does not change from OSS to PRO currently, so unless that's a todo, I think this may be misleading.

chapter-using-nexus-ui.asciidoc
```diff
@@ -55,7 +55,7 @@ menu section on the left of the user interface. The contents of the
 menu will depend on your assigned user privileges.
 
 Administration button image:figs/web/ui-admin-button-icon.png[scale=50];;
- The browse button allows to toggle the 'Administration' menu items in
+ The administration button allows to toggle the 'Administration' menu items in
```
joedragons: Defer to @djsauble , but I don't like the word "toggle" here (or above in browse).  IMO that means on/off in one menu option (and I believe webster's definition agrees) which this is not, it's 2 buttons and 2 menus.  Suggest switching (for example, maybe you can come up with better) "to toggle" to "a switch to".

joedragons:
Tangential (not changed in this PR) but I feel like the example for refresh is only needed because of the semi-confusing way the explainatory sentence is written (ie it says "all views" but really it's just the views you're on).  Consider striking example and changing the first sentence to something like "The refresh button is a global refresh button for the menus and views displayed at the time the refresh is done."  Probably could be polished/better.

chapter-using-nexus-ui.asciidoc
```diff
@@ -78,13 +78,13 @@ the main menu on the left.. In addition the sign out button
 will be displayed.
```
joedragons: Extra period in line 77 ("left..").

chapter-using-nexus-ui.asciidoc
joedragons: Note, @djsauble corrected a plaurality here as well=)  "menu items displays" should be "menu item displays" or "menu items display".  Incidentally, removing the word menu doesn't correct this.

chapter-using-nexus-ui.asciidoc
```diff
@@ -78,13 +78,13 @@ the main menu on the left.. In addition the sign out button
...
 
-Help button image:figs/web/ui-help-button-icon.png[scale=50];; Clicking the
- help button opens up the help menu.  It contains a link to specific
- help about the currently active feature view. The 'About' menu items
- displays a window with details about the Nexus version as well as
- license and copyright information. The 'Documentation', 'Support' and
- 'Issue Tracker' link to the respective sections on the Sonatype
- website.
+Help button image:figs/web/ui-help-button-icon.png[scale=50];;
+ Clicking the help button opens up the help menu.  It contains a link
```
joedragons: Minor: I thought it weird that everything else is in left/right order except for help (which is currently positioned before sign in, in PRO).  Maybe intentional.  2 cents.

chapter-using-nexus-ui.asciidoc
```diff
@@ -92,7 +92,7 @@ The main menu on the left contains either the 'Browse', the
 'Administration' or the 'User' menu items. The exact list of available
 menu items depends on the current user's assigned privileges. E.g.,
 the 'Browse' menu as visible in <<fig-ui-overview-admin>> includes the
-'Upload' section, which is not available to anonymous users by
+'Security' section, which is not available to anonymous users by
 default.  The panel itself can be horizontally collapsed and expanded
```
joedragons: Minor: This is not a great example IMO, since Anonymous have no admin privilidges (we even state that) so would never see that list.  Consider switching anonymous to Deployment users?

chapter-using-nexus-ui.asciidoc
```diff
@@ -102,22 +102,28 @@ display of the respective feature view in the feature view panel.
 Feature View Panel::
```
joedragons: Minor: Just of note, this is currently referenced 7 times on this page before it's described what it is.  I figured it out but I am also not a new user.  I am not sure if anything can be done about this without somewhat of an overhaul or if it's worth it.  Wanted to make a note since I noticed.

chapter-using-nexus-ui.asciidoc
```diff
@@ -102,22 +102,28 @@ display of the respective feature view in the feature view panel.
...
+<<fig-ui-example>> shows a typical user interface appearance of Nexus
+with the 'Users' feature view in the feature view panel. It shows a
+list of users.  
 
 [[fig-ui-example]]
-.Typical Example Nexus Interface with Repository List and Details
+.Typical Example Nexus Interface with a List
 image::figs/web/ui-example.png[scale=40]
 
+Clicking on a row in the list, switches the feature view to an item
```
joedragons: Believe possession ('s) of item is missing here (ala "item's").

joedragons:
+1 besides non-minor/suggestion notes.

chapter-using-nexus-ui.asciidoc
djsauble: I just tried opening Nexus 3 in IE8, and it stalls on loading.

Didn’t realize the experience was completely broken on IE8 (I thought we took you to a browser compatibility warning screen, and you could choose to proceed or not with your current browser).

chapter-using-nexus-ui.asciidoc
djsauble: Agree. The button is being toggled, not the menu. Perhaps instead:

“The administration button displays the Administration menu items in the main menu section on the left of the user interface…”

djsauble:
+1, minor suggestions about consistency

chapter-using-nexus-ui.asciidoc
joedragons: Well that concerns me too, but not sure if it should.  Maybe I'm just not in a loop.  Thanks for checking.

mosabua: This is from Nexus 2 .. I will leave this as is assuming that you are going to fix it somehow and then file a documentation subtask to update here..

joedragons: Not to be contradictory but NX2 reads "Older versions such as Microsoft Internet Explorer 7 or earlier are not supported and actively blocked from using Nexus to avoid an unsatisfactory user experience. Internet Explorer 8 works up to Nexus 2.8 and is not supported for newer releases." so this has changed.  I wouldn't have said anything if it was the same.  OTOH what is there matches the current NX3 experience per @djsauble so I agree with your logic.

chapter-using-nexus-ui.asciidoc
mosabua: Reworded. I think a logo change might happen in the future (currently debated) but it doesnt matter with the new wording. For your reference in Nexus 2 the logo image is different because the string is part of the logo. Now its all text. I would be surprised if the rebranding effort will not force us to use the new names like Nexus Pro and Nexus OSS ..

chapter-using-nexus-ui.asciidoc
mosabua: Changed to switch. Imho toggle is equivalent as a word to switch to but if you prefer switch.. thats fine by me.

chapter-using-nexus-ui.asciidoc
mosabua: Fixed

chapter-using-nexus-ui.asciidoc
mosabua: That changed from the last version we documented. Fixed.

chapter-using-nexus-ui.asciidoc
mosabua: I changed to just dialog. Where do we use new windows btw?

chapter-using-nexus-ui.asciidoc
mosabua: All fixed.

chapter-using-nexus-ui.asciidoc
mosabua: Fixed.

chapter-using-nexus-ui.asciidoc
mosabua: Reworded to dialog and window

chapter-using-nexus-ui.asciidoc
mosabua: Synced

chapter-using-nexus-ui.asciidoc
mosabua: Feature view because Jason said thats what its called. And I agree.. a feature is more than just a single view in the UI.

chapter-using-nexus-ui.asciidoc
mosabua: I am leaving this for now. You will have to fix the behaviour and we can adjust doc then,.

chapter-using-nexus-ui.asciidoc
mosabua: Cahnged to switch

chapter-using-nexus-ui.asciidoc
mosabua: That changed in the UI. Adjusted now in docs.

chapter-using-nexus-ui.asciidoc
mosabua: Fixed

chapter-using-nexus-ui.asciidoc
mosabua: True .. do we need a list first maybe?  ... I thought not since it is all close together..

chapter-using-nexus-ui.asciidoc
mosabua: I thought I hadnt changed it at all. I guess I did after all.

PR #51, opened 2015-05-01
Updated tool configs to new CMA repos

mosabua:
https://issues.sonatype.org/browse/NEXUS-8530

At this stage this chapter referenced the nexus-book-examples repo. I think we will have to create a Nexus 3 branch there now and update all the examples to work against Nexus 3 with CMA. That is also going to be a good test.

Please let me know if you agree on this approach and if so I can get this work included as part of this issue and respectively also update the PR with new wording.

### DIFF
--- a/chapter-maven.asciidoc
+++ b/chapter-maven.asciidoc
@@ -35,7 +35,7 @@ anchor:ex-maven-nexus-simple[Listing: Configuring Maven to Use a Single Nexus Gr
       <!--This sends everything else to /public -->
       <id>nexus</id>
       <mirrorOf>*</mirrorOf>
-      <url>http://localhost:8081/content/groups/public</url>
+      <url>http://localhost:8081/repository/maven-public/</url>
     </mirror>
   </mirrors>
   <profiles>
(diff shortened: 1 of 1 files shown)


mosabua:
Rendered output at http://bamboo.s/browse/BOOK-NEXUS18-1/artifact/STAGE/site/3.0/reference/config.html

joedragons:
Tangentially, I think line 164 ".Listing: Minimal Aether Configuration in an Ant file" has an extra . before Listing.  This is not present in NX2.11 book tho does not appear to have been changed in this PR.

joedragons:
Also tangentially (and does not appear to be changed by this PR tho it is in the reviewed matieral), I personally believe having the TOC first on http://bamboo.s/browse/BOOK-NEXUS18-1/artifact/STAGE/site/3.0/reference/config.html is a sure fire way to have 90% of people never read the text that follows it.  I know this is typical doc setup but just a thought.  This is a deviance from NX2 book where there was a seperate Introduction "chapter".

Anyway, back on topic, I am +1 to the section for coherancy but I didn't check all those formats.  I believe that speaks to the plan you are suggesting (redoing example section) which I am also +1 to.

PR #52, opened 2015-05-04
Remove redundant letter

joedragons:
I believe either the letter a is redundant or there is a word missing.  Sending PR to assure I'm right and I picked the right fix.
I noticed this reviewing the NX3 book draft btw.


mrprescott:
+1 :)

PR #53, opened 2015-05-25
Word changes

joedragons:
For review:

First word, I believe might be the wrong word but not 100% confident about the word change.

Second word, I believe is wrong because the field before it is required so it's not instead.


mrprescott:
+1

PR #60, opened 2015-06-09
Rebuild Maven Repository Metadata task docs

mosabua:
https://issues.sonatype.org/browse/NEXUS-8515

### DIFF
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -367,6 +367,14 @@ can also take up storage space. Using this task you can tell Nexus to
 periodically purge this information. The setting 'Purge Items older
 than (days)' controls the age of the data to be deleted.
 
+Rebuild Maven Repository Metadata:: This task will rebuild the
+maven-metadata.xml files with the correct information and will also
+validate the checksums (.mh5/.sha1) for all files in the specified
+maven2-format repository. The 'Group Id;, 'Artifact Id' and
+'Base Version' parameters allow you to narrow down the section of the
+repository that will be repaired. Typically this task is run manually to
+repair a corrupted repository.
+
 ////
 Backup All Nexus Configuration Files:: This scheduled task will
 archive the contents of the +sonatype-work/nexus/conf+ directory.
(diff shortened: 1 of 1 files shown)


mosabua:
Rendered output at http://bamboo.s/browse/BOOK-NEXUS26-1/artifact/STAGE/site/3.0/reference/system.html#tasks

chapter-configuration.asciidoc
```diff
@@ -367,6 +367,14 @@ can also take up storage space. Using this task you can tell Nexus to
 periodically purge this information. The setting 'Purge Items older
 than (days)' controls the age of the data to be deleted.
 
+Rebuild Maven Repository Metadata:: This task will rebuild the
+maven-metadata.xml files with the correct information and will also
+validate the checksums (.mh5/.sha1) for all files in the specified
```
joedragons: ".md5" typo here?

chapter-configuration.asciidoc
```diff
@@ -367,6 +367,14 @@ can also take up storage space. Using this task you can tell Nexus to
...
+maven2-format repository. The 'Group Id;, 'Artifact Id' and
```
joedragons: Minor: I feel like other places you've been putting 'maven2' format...but I may be wrong.  The - caused me pause anyway, but I don't feel strongly.

joedragons:
+1 besides comments.
As far as keeping repo names lower case, I feel here it looks odd since every other word for Tasks is clearly caps.  Probably good to decide before we get more tasks though...only have 2 now so easy to make a "rule".  Could be something for the principles document one way or the other (@djsauble).

mosabua:
I agree with @jdillon and we should keep it consistent and use the lower case names for the repository formats everywhere. This reduces confusion for the users in the UI as well as readers of the docs and keeps it consistent. We just have to make sure we update the UI as there are currently inconsistencies and move to use 'maven2', 'nuget' and 'raw' only. cc @jdillon 

Maybe we should file a issue to get that work done at one sweep .. or we keep it on the list of todo items for the whole relabelling. I would rather get it done now though.

chapter-configuration.asciidoc
mosabua: fixed

chapter-configuration.asciidoc
mosabua: 'maven2' or so is used when it refers to something in the UI. Otherwise it is just a normal word. I removed format though since maven2 repository is easily understood and concise.

PR #61, opened 2015-06-11
Minimal install docs

mosabua:
https://issues.sonatype.org/browse/NEXUS-8822

### DIFF
--- a/chapter-installation.asciidoc
+++ b/chapter-installation.asciidoc
@@ -2,7 +2,7 @@
 == Installing and Running Nexus
 
 Nexus only requires a Java Runtime Environment and installation is a
-simple process. This chapter provides further details for the get
+simple process. This chapter provides further details to get
 started with Nexus and keep it running successfully in production
 deployments.
 
@@ -10,9 +10,9 @@ WARNING: Currently Nexus 3 is a pre-release application and running it
 for production usage is not recommended. The installation
 process is only suitable for evaluation and testing usage.
 
-=== JRE Prerequisites
+=== Java Prerequisite
 
-{oss} and {pro} only have one prerequisite, a Java Runtime Environment
+{nexus} only has one prerequisite, a Java Runtime Environment
 (JRE) compatible with Java 7 or Java 8 from Oracle. Nexus is most
 often run with the JRE that is bundled with a Java Development Kit
 (JDK) installation. We recommend to use the latest version of Java
@@ -29,7 +29,7 @@ Java HotSpot 64-Bit Server VM (build 25.45-b02, mixed mode)
 ----
 
 When multiple JDK or JRE versions are installed, you need to ensure
-the correct version is using by running the above command as the
+the correct version by running the above command as the
 operating system user, that is used to run Nexus.
 
 ////
@@ -89,14 +89,15 @@ version] for an evaluation.
 
 The following instructions are for installing {oss} or {pro} as a
 stand-alone server. The Nexus bundle combines the Nexus web
-application with a Eclipse Jetty server and supplementary files like
+application with an Eclipse Jetty server and supplementary files like
 startup scripts.
 
 Installing Nexus is straightforward. Unpack the Nexus bundle archive
-in a directory. If you are installing Nexus on a local workstation to
-give it a test run, you can install it in your home directory or
-wherever you like. Nexus doesn't have any hard coded directories. It
-will run from any directory. If you downloaded the ZIP
+in a directory, to which you have full access. If you are installing
+Nexus on a local workstation to give it a test run, you can install it
+in your home directory or wherever you like. Nexus doesn't have any
+hard coded directories. It will run from any directory. If you
+downloaded the ZIP archive, run:
 
 ----
 $ unzip nexus-3.0.0-b2015061001-bundle.zip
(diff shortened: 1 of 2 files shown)


mosabua:
Rendered output at http://bamboo.s/browse/BOOK-NEXUS27-1/artifact/STAGE/site/3.0/reference/install.html

mosabua:
New build coming shortly.

mosabua:
Here is latest rendered output http://bamboo.s/browse/BOOK-NEXUS27-3/artifact/STAGE/site/3.0/reference/install.html

mosabua:
One more update ;-) http://bamboo.s/browse/BOOK-NEXUS27-4/artifact/STAGE/site/3.0/reference/install.html

chapter-installation.asciidoc
```diff
@@ -1,17 +1,40 @@
...
-Java Runtime Environment (JRE) compatible with Java 7. Nexus is most
-often run with the JRE that is bundled with a Java Development Kit
-(JDK) installation, and it can be run with Oracle's JDK for Java 7. To
-download the Oracle JDK, go to
-http://www.oracle.com/technetwork/java/javase/downloads/index.html . At
-a minium Java 7u2 is required, but we recommend to use the latest available
-version.
+WARNING: Currently Nexus 3 is a pre-release application and running it
+for production usage is not recommended. The installation
+process is only suitable for evaluation and testing usage.
```
joedragons: Minor: I do not see the need for the second sentence.  The installation process should be the same no matter what usage.  If left, it reads awkwardly to me, but I know what you mean.

chapter-installation.asciidoc
```diff
@@ -65,83 +85,79 @@ Pro Support landing page].
...
 
 ----
-$ sudo cp nexus-2.9.0-02-bundle.tar.gz /usr/local
-$ cd /usr/local
-$ sudo tar xvzf nexus-2.9.0-02-bundle.tar.gz
-$ ln -s nexus-2.9.0-02 nexus
+$ sudo cp nexus-3.0.0-b2015061001-bundle.tar.gz /opt
+$ cd /opt
+$ sudo tar xvzf nexus-3.0.0-b2015061001-bundle.tar.gz
+$ ln -s nexus-3.0.0-b2015061001 nexus
```
joedragons: Minor: I am unsure the point of the ln unless it's for reference later (that'd surprise me as you don't rename the linux directories).  Could be covered later or my inexpertise but wanted to make a note while I was here/remembered.

chapter-installation.asciidoc
```diff
@@ -65,83 +85,79 @@ Pro Support landing page].
...
 avoid problems with Windows file registry virtualization. If you plan
 to run Nexus as a specific user you could install into the
 +AppData\Local+ directory of that users home directory. Otherwise
-simply go with e.g., +C:\nexus+ or something similar.
+simply use e.g., +C:\nexus+ or something similar.
 
-The Nexus installation directory +nexus-2.9.0-02+ or
-+nexus-professional-2.9.0-02+ has a sibling directory named
+The Nexus installation directory +nexus-3.0.0-b2015061001+ or
++nexus-professional-3.0.0-b2015061001+ has a sibling directory named
```
joedragons: Minor: Recommend just removing the directory names.  Saves you from updating them each time and not sure the value added (also these can be changed).

chapter-installation.asciidoc
```diff
@@ -65,83 +85,79 @@ Pro Support landing page].
...
 of Nexus.
 
-By default, this directory is always a sibling to the Nexus
-installation directory. If you installed Nexus in the +/usr/local+
+By default, this directory is a sibling to the Nexus
+installation directory. If you installed Nexus in the +/opt+
 directory it would also contain a +sonatype-work+ subdirectory with a
 nested +nexus+ directory containing all of the content and
-configuration. The location of the +sonatype-work+ directory can be
+configuration.
```
joedragons: I feel like this paragraph is redundant.  You stated in the previous it's a subling.  The later sentences just seem to be describing what a sibling is.  If my comment seems weird, I am likely missing the intent of this paragraph.

joedragons:
+1, minor notes

I agree with Jeffry's sentiment we remove pro for now.  We did it pretty much everywhere else as you know we had several discussions;)

chapter-installation.asciidoc
mosabua: Well.. the reason I am saying this is that there is no installing as a service or running in the background or installer or anything. You think I can still toss it?

chapter-installation.asciidoc
mosabua: Its to make it the same for each release. So that it can be referenced from outside scripts under the same location. Kind of secondary now.. and might change later.

chapter-installation.asciidoc
mosabua: Its better to be explicit in the examples.. changing them is an easy search and replace and we have done that for each release for a long time.

mosabua:
Pro removed.

chapter-installation.asciidoc
joedragons: I see.  You're more familiar with those things than me, I defer.

chapter-installation.asciidoc
mosabua: Reworded to make it clearer.

chapter-installation.asciidoc
joedragons: That is fair, but above, don't you rename it to just "nexus"?  I feel like there's two example threads running here and you bounce back and forth.

PR #62, opened 2015-06-24
replaced icons for editions with plain text

mosabua:
https://issues.sonatype.org/browse/NEXUS-8513


mosabua:
This is trivial so I am merging without review.

PR #63, opened 2015-07-10
Updated version strings to 2.11.4-01

mosabua:
https://issues.sonatype.org/browse/NEXUS-8901


jwayman:
Was there really no changes other than versions? If so, +1 from me - might be nice to get a +1 from another member of the team.

mosabua:
I looked at the linked JIRA issues and there is nothing from what I can tell. Stuart and Jason agree as well.

PR #64, opened 2015-07-14
revise reverse proxy section

peterlynch:


rseddon:
Looks really good for the most part.  A few small nits:

> "Nexus is essentially a sophisticated web application"  

I think you should probably should leave "essentially" out of that?

> Using a https://en.wikipedia.org/wiki/Reverse_proxy[reverse proxy]
+in front of Nexus is not typically required or necessary.

I think this should be "using a reverse proxy in front of Nexus to enhance performance is not typically required or necessary".

Also, in the section below that describing why you might want a reverse proxy the two most common reasons are not listed:

* To run nexus on a restricted port
* Using the reverse proxy to server SSL

Although they are both well documented below that.

mosabua:
Excellent. We should merge this after the minor changes suggested by Rich and get it live with the 2.11.4 release. I can review in detail tonight before or after you merge..

peterlynch:
> I think this should be "using a reverse proxy in front of Nexus to enhance performance is not typically required or necessary".

It is is never required for performance.

peterlynch:
@mosabua possible toget a more useful failure messages from CI build?

http://bamboo.s/browse/BOOK-NEXUS30-6

mosabua:
You can check by building it locally or commit the addition of -v to the a2x call. I cant look atm since I cant connect to the VPN here. Anyway... I am happy to fix it up and push it tonight as long as you are happy with the text. Just let me know with a comment that you are done and I can do the rest.

jwayman:
I don't mind making verbose the stock setting for these things (on the lifecycle).

mosabua:
Either is fine by me.. feel free to commit that..

peterlynch:
@mosabua Jeff filled me in on --verbose option. Setting up the develop environment failed for me sith some strange error from mac ports dblatex install. I think next time I will just commit the --verbose option to a2x to see what is wrong

mosabua:
Renders fine. Imho we can push this live as soon as 2.11.4 is release since these version changes are already in master as well.

PR #65, opened 2015-07-16
Adds installation commands for OSX and Homebrew

djsauble:
screen shot 2015-07-16 at 11 56 10 am

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -39,13 +39,29 @@ them or contact us directly if you need something.
 
 ==== How to Build the Book
 
-I knew you'd ask that, and here's the simple answer:
+I knew you'd ask that, and here's the simple answer (assumes Homebrew and OSX):
 
-* Install asciidoc
-* Install dblatex and the docbook xsl style sheets
 * Install python
+ $ brew install python
+* Install asciidoc
+ $ brew install asciidoc
+* Install the docbook xsl style sheets
+ $ brew install docbook
+* Install a LaTeX distribution (Download 2.5GB package from https://www.tug.org/mactex/)
+* Install dblatex
+...
+$ sudo easy_install pip
+$ pip install dblatex
+...
 * Install imagemagick
-* run ./build.sh
+...
+$ brew link libpng
+$ brew install imagemagick
+...
+* Run the build
+ $ ./build.sh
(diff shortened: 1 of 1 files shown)


mosabua:
Using brew is a lot more problematic to install then using macports. We should document both if desired..

kellyrob99:
Instructions for macports would be appreciated for sure :)

djsauble:
Happy to let someone else take a stab at the Macports documentation. I use Homebrew myself, so that’s what I documented.

mosabua:
This is already documented on the wiki. It has not been here on purpose since the instructions apply to the Nexus and Nexus Lifecycle docs as well as the Maven books. Adding it here seems inefficient and the wrong place.

mosabua:
wdyt @jwayman

djsauble:
If it’s documented elsewhere, we should at least include a link to that in the readme here. I got stuck trying to figure this out, so clearly the docs could do a better job at pointing you to more detailed installation instructions.

mosabua:
True... I will let @jwayman sort this out or I can work on it when we are back. 

Just for reference

https://docs.sonatype.com/display/ProdMgmt/Asciidoc+%28Book%29+Publishing+Process

https://docs.sonatype.com/display/ProdMgmt/Functional+User+Documentation

https://docs.sonatype.com/display/ProdMgmt/Help+Content

Imho the install instructions should be somewhere there and potentially improved there and maybe a link to it added here. To some degree we however need a public space since the Maven books as well as the Nexus docs are CC licensed and open to the public.

jwayman:
I don't fully parse this thread. However, I don't think we should be in the
business of maintain instructions for contributors (outside of Sonatype
employees) to build these books locally. Besides the trouble with getting
it running... what's the actual issue with how it;s documented?

Jeff

mosabua:
I just think the docs should be in a central location for all our books and documentation.. otherwise it seems just fine.

jwayman:
+1 to central location. -1 to including in a readme. Which is technically
where it is now... that could use some improvement.

mosabua:
I am closing this as the content should be in the respective wiki pages since they apply to multiple of our books and not in this specific readme.

mosabua:
It would be great if you could add it there @djsauble

PR #67, opened 2015-07-21
Corrected some text typos.

andham:
Signed-off-by: Anders Hammar <anders@hammar.net>


joedragons:
+1 tho makes me wonder (didn't check) if any of these areas should have better markup vs just caps correctness.

mosabua:
Thanks @andham 
@joedragons consistent usage of caps like @andham fixed it should be enough.

I will push this live after the merge.

mosabua:
Live now.

PR #70, opened 2015-07-29
Integration with VS (16.8) text updates

joedragons:
I was rereviewing this against VS2015 and do not believe we updated some of the words when we updated the screenshot for VS2013.
No further updates need to be done against VS2015 but added some clarification overall.


mosabua:
When we merge this, we have to also cherry-pick or port to the nexus-3.0.x branch

joedragons:
@mosabua Agreed.

PR #76, opened 2015-09-16
NEXUS-9041 fix https error and add docker push instruction.

jwayman:


chapter-docker.asciidoc
```diff
@@ -182,20 +182,34 @@ private images you have pushed to your hosted repositories. In addition it inclu
...
 
 [[docker-access]]
-=== Accessing Docker Images
+=== Pulling/Pushing Docker Images
 
-Downloading Docker images from Nexus can be performed with the `pull` command. The only necessary addition are the
-hostname or IP address of the Nexus server as well as the Docker connector port for the repository or repository group
-to download from:
+Downloading Docker images from (pulling), or uploading to (pushing), Nexus can be performed with the Docker `pull` or
+`push` command (respectively). The only necessary addition are the hostname or IP address of the Nexus server as well
```
kellyrob99: command here should be pluralized, and I would drop the 'respectively' as it's fairly obvious already

chapter-docker.asciidoc
```diff
@@ -182,20 +182,34 @@ private images you have pushed to your hosted repositories. In addition it inclu
...
+----
+docker <command> <nexus-hostname>:<repository-port>/<namespace>/<image>:<tag>
+----
+
 The preferred setup is to proxy a relevant source of public/private images you want to use (Docker Hub being the most
 common choice), configure one or more hosted repositories to contain your own images, and expose these repositories
 through one Docker group repository.
-Examples for pulling various images from a Nexus server running on the host `nexus.example.com` and exposing a Docker
+
+Examples for p various images from a Nexus server running on the host `nexus.example.com` and exposing a Docker
```
kellyrob99: typo here "for p"

joedragons: "p" word cutoff here?  Or leftover 'p'?

chapter-docker.asciidoc
```diff
@@ -210,15 +224,21 @@ therefore show up as `library/ubuntu` and `library/postgres` in Nexus.
 
 After a successful `pull` you can start the container with `run`.
 
-[[docker-push]]
-=== Pushing Docker Images to Docker Hosted Repositories
-
-TBD - docker push and so on, what about credentials, ...
+If we made changes to each of these images, and then wanted to push them up to our Nexus repository. It might look
```
kellyrob99: So in order to do this you need to 'tag' the image so this might better read as:
```
If we build or modify an image and then want to push them up to our Nexus repository, it might look something like this:
```

chapter-docker.asciidoc
```diff
@@ -210,15 +224,21 @@ therefore show up as `library/ubuntu` and `library/postgres` in Nexus.
...
 ----
-docker <command> <nexus-hostname>:<repository-port>/<namespace>/<image>:<tag>
+docker push nexus.example.com:18443/ubuntu:my_update
+docker push nexus.example.com:18443/bitnami/node:old
+docker push nexus.example.com:18443/postgres:9.5
 ----
+
+Now, these updated images would be ready to be pulled by anyone with access to the repository.
+
+NOTE: Prior to push, and depending on your Nexus configuration, login credentials may be required before a push or pull
```
joedragons: Am curious to others thoughts but afaik this will always be required unless they're saved in docker, has nothing to do with nexus config (ie no matter what you do with nexus, it'll always be required).  If you strike "Nexus" I believe this statement is equally if not more true.

chapter-docker.asciidoc
```diff
@@ -210,15 +224,21 @@ therefore show up as `library/ubuntu` and `library/postgres` in Nexus.
...
-=== Pushing Docker Images to Docker Hosted Repositories
-
-TBD - docker push and so on, what about credentials, ...
+If we made changes to each of these images, and then wanted to push them up to our Nexus repository. It might look
+something like this:
 
-NOTE: All commands using the Docker client against Nexus follow a common structure:
 ----
-docker <command> <nexus-hostname>:<repository-port>/<namespace>/<image>:<tag>
+docker push nexus.example.com:18443/ubuntu:my_update
```
kellyrob99: Add:
```
docker tag <imageId> nexus.example.com:18443/ubuntu:my_update
```

chapter-docker.asciidoc
kellyrob99: should mention that these are credentials for a User on Nexus, as opposed to DockerHub for instance

chapter-docker.asciidoc
joedragons: I think "addition" should also be plauralized (to additions).  It's followed up with "are" and multiple things.

joedragons:
+1 besides notes=)  Thanks!

chapter-docker.asciidoc
joedragons: Perhaps move the word "Nexus" from before configuration to before login?

chapter-docker.asciidoc
jwayman: Fixed respectively. Not convinced on the plural. "Or" should stand as primarily singular. Though additions in the next sentence surely should be. I could be grammatically swayed.

chapter-docker.asciidoc
jwayman: Fixed.

chapter-docker.asciidoc
jwayman: Fixed.

chapter-docker.asciidoc
jwayman: If Nexus Anonymous access it enabled (maybe this is gone with 3). That's the scenario I was thinking of.

joedragons: Currently (to my knowledge) you cannot push anonymously in NX3.  The anonymous role doesn't come with it and you cannot (currently) edit the permissions.  I suppose that may change.  I feel like it's safer to (re)move the word based around how NX3 is now.  It shouldn't make anything untrue in the future.  2 cents, but others may know more=)

chapter-docker.asciidoc
jwayman: I changed the examples up a little please review for accuracy.

kellyrob99:
+1

PR #82, opened 2015-09-28
Removal of sonatype-work

joedragons:
Work on https://issues.sonatype.org/browse/NEXUS-9223 - removal of sonatype-work references from NX3 documentation.

Bamboo: http://bamboo.s/browse/BOOK-NEXUS44

### DIFF
--- a/chapter-installation.asciidoc
+++ b/chapter-installation.asciidoc
@@ -149,18 +149,6 @@ to run Nexus as a specific user you could install into the
 +AppData\Local+ directory of that users home directory. Otherwise
 simply use e.g., +C:\nexus+ or something similar.
 
-The Nexus application directory +nexus-3.0.0-b2015061001+ requires
-another directory named +sonatype-work+. This directory contains all
-of the repository and configuration data for Nexus and is stored
-outside of the Nexus installation directory to make it easier to
-upgrade to a newer version of Nexus.
-
-By default, this directory is a sibling to the Nexus
-installation directory. If you installed Nexus in the +/opt+
-directory it would also contain a +sonatype-work+ subdirectory with a
-nested +nexus+ directory containing all of the content and
-configuration. 
-
 ////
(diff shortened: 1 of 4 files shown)


chapter-installation.asciidoc
```diff
@@ -149,18 +149,6 @@ to run Nexus as a specific user you could install into the
 +AppData\Local+ directory of that users home directory. Otherwise
 simply use e.g., +C:\nexus+ or something similar.
 
-The Nexus application directory +nexus-3.0.0-b2015061001+ requires
-another directory named +sonatype-work+. This directory contains all
-of the repository and configuration data for Nexus and is stored
```
mosabua: Should be reworded to talk about the nested data folder instead and then about how to configure it to be somewhere else..

joedragons: @mosabua Won't that be covered in the blobstore section?  Or maybe I'm missing what you want here.

joedragons:
I see some references in chapter-migrating.asciidoc but it does not look like that's live in NX3 yet.  Rather than blindly stripping or rewriting that chapter I am leaving those alone.

chapter-installation.asciidoc
mosabua: I am updating this as well as reviewing and merging and publishing of the whole enchilada ;-)

chapter-introduction.asciidoc
```diff
@@ -233,9 +233,7 @@ a single repository group, and you should download a copy of the free
 Nexus book - 'Repository Management with Nexus'. Once you've
 familiarized yourself with {oss}, you can easily upgrade to
 {pro} by downloading and installing Nexus
-Pro. Nexus stores all of your repository data and
-configuration in a directory named +sonatype-work+, which is separate
```
mosabua: Just fyi .. that chapter is not yet included and might never be ;-)

PR #83, opened 2015-09-30
NEXUS-9281 Documenting Max Metadata Age

joedragons:
And a typo fix.

### DIFF
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -884,9 +884,14 @@ attempt to locate an component in a remote repository. Nexus will
 periodically retest the remote repository and unblock the repository
 once it becomes available.
 
-Maximum component age:: Tells Nexus what that maximum age of a
+Maximum component age:: Tells Nexus what the maximum age of a
 component is, before it retrieves a new version from the remote
-repository.  
+repository.
+
+Maximum metadata age:: Tells Nexus how long to cache metadata before
+rechecking the remote repository.  If the metadata is component metadata,
+it uses the longer of this value and 'Maximum component age' before
+rechecking.
 
 ////
 tbd once fixed..
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -884,8 +884,8 @@ attempt to locate an component in a remote repository. Nexus will
 periodically retest the remote repository and unblock the repository
 once it becomes available.
 
-Maximum component age:: When Nexus receives a request for a component, it
-only retrieves a new version from the remote repository when the existing
+Maximum component age:: When the proxy receives a request for a component, it
+does not request a new version from the remote repository until the existing
 component is older than 'Maximum component age'.
 
 Maximum metadata age:: Nexus retrieves metadata from the remote repository.
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -888,10 +888,10 @@ Maximum component age:: Tells Nexus what the maximum age of a
 component is, before it retrieves a new version from the remote
 repository.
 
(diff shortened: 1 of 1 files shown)


chapter-configuration.asciidoc
```diff
@@ -884,9 +884,14 @@ attempt to locate an component in a remote repository. Nexus will
 periodically retest the remote repository and unblock the repository
 once it becomes available.
 
-Maximum component age:: Tells Nexus what that maximum age of a
+Maximum component age:: Tells Nexus what the maximum age of a
 component is, before it retrieves a new version from the remote
-repository.  
+repository.
+
+Maximum metadata age:: Tells Nexus how long to cache metadata before
```
mrprescott: perhaps, "Tells Nexus how long it should use cached metadata before rechecking.." since the caching doesn't necessarily end after this time period.

chapter-configuration.asciidoc
```diff
@@ -884,9 +884,14 @@ attempt to locate an component in a remote repository. Nexus will
...
+rechecking the remote repository.  If the metadata is component metadata,
```
mrprescott: I'm now wondering if we define metadata anywhere, it seems we're drawing a distinction between two things, "metadata" and a particular type of metadata, without having explained what metadata is or what the other types might be.

mrprescott:
+1 w/ minor comments

mrprescott:
This PR got me wondering whether we need to start talking about Nexus as 'Nexus RM'.

chapter-configuration.asciidoc
joedragons: That is fair.  I could leave this generic but people may be confused.  I was, then @cstamas set me straight.

jdillon:
"Tells Nexus" seems like poor language to use to describe a setting.

chapter-configuration.asciidoc
```diff
@@ -884,9 +884,14 @@ attempt to locate an component in a remote repository. Nexus will
 periodically retest the remote repository and unblock the repository
 once it becomes available.
 
-Maximum component age:: Tells Nexus what that maximum age of a
+Maximum component age:: Tells Nexus what the maximum age of a
 component is, before it retrieves a new version from the remote
-repository.  
+repository.
```
peterlynch: age defined as what? minutes/seconds/ms - is the unit defined elsewhere in the UI and do we deliberately exclude it here?

joedragons:
@jdillon @mrprescott any better?

chapter-configuration.asciidoc
```diff
@@ -884,9 +884,14 @@ attempt to locate an component in a remote repository. Nexus will
 periodically retest the remote repository and unblock the repository
 once it becomes available.
 
-Maximum component age:: Tells Nexus what that maximum age of a
-component is, before it retrieves a new version from the remote
-repository.  
+Maximum component age:: When Nexus receives a request for a component, it
+only retrieves a new version from the remote repository when the existing
+component is older than 'Maximum component age'.
```
peterlynch: Doesn't only retrieve a new component if the component is actually different?

chapter-configuration.asciidoc
joedragons: It specifically says "(in minutes)" in the UI.  I can add it here as well, but seemed redundant to me.

chapter-configuration.asciidoc
joedragons: True (and tested) but I'm inclined to "solve" this by putting in a negative tense rather than list all the conditions that Nexus needs to make to retreive a component=)  Let me know if you like the rewrite better.

joedragons:
Squash merged 8ae16df with nexus-3-M6.

PR #90, opened 2015-10-09
NEXUS-9022 Documentation of proxy repository index button

joedragons:
https://issues.sonatype.org/browse/NEXUS-9022

Not sure if we want to say more or not.  Speak your piece or hold your peace?=)


mosabua:
Good enough for now imho. Once index downloads remote are happening we will have to elaborate more. But we can deal with that later .. +1

PR #92, opened 2015-10-21
NEXUS-9337 Adjusted references regarding SSL creation

joedragons:
* Adjusted references from custom.properties to org.sonatype.nexus.cfg
(NEXUS-9337)
* Removed the example link to 2.5 from 5.8.4 as there is no example
there (just a list of files).  If an example is added, we could addback
this link, currently I believe deceiving.
* Tightened some spacing to match styleguide

https://issues.sonatype.org/browse/NEXUS-9337

CI: http://bamboo.s/browse/BOOK-NEXUS52

### DIFF
--- a/chapter-installation.asciidoc
+++ b/chapter-installation.asciidoc
@@ -7,14 +7,14 @@
 Nexus only requires a Java Runtime Environment and installation is a simple process. This chapter provides further
 details to get started with Nexus and keep it running successfully in production deployments.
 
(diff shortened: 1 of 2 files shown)


mosabua:
+1

In the future please don't reformat paragraphs to different format unless you are also changing the content. This enables much faster reviews.

joedragons:
Acknowledged.

PR #94, opened 2015-10-30
NEXUS-9320 minor language update.

jwayman:


jwayman:
Addressed comments.

chapter-docker.asciidoc
```diff
@@ -1,5 +1,5 @@
 [[docker]]
-== Docker Container Registry Support
+== Docker Repository
```
mosabua: imho this should use Registry as this is the native name used in the Docker ecosystem. A Docker repository is actually the images in a registry so this is kind of wrong the way we are naming this now.

https://hub.docker.com/explore/

However talked to @jwayman about how this change came about and I will not restart the discussion. Just wanted to leave a comment to clarify that I disagree with this naming.

kellyrob99: @mosabua Whichever way you twist it, the nomenclature does not mesh with what exists in NX already and my 2 cents says we go with ours and note the discrepancy rather than start renaming NX bits.

jwayman: Yeah this is a toss up. Perhaps we can still find a middle ground. Why not put Private Docker Registry in parentheses?

mosabua: Imho the target audience is docker users .. and for them its a registry. But lets not worry. That ship has sailed. Just wanted to mention my concerns.

jwayman: I think we should use data to solve this. Which as it happens aligns with the push to use "Private Docker Registry" but also aligns with a push away from plural.

Below are the results for search volume among the various  options

keyword_planner_ _google_adwords

FWIW this is still a change from the original title, so a change, to some degree was needed.

@mosabua I think it might be worth coming up with a process on how we name these headings, as these can (and do translate to SEO value). We should always put the user first, but sometimes that can be supported by trending and search volume data - it's just a smarter way and will take out the guess work.

I'll make the appropriate changes, and reopen the PR.

PR #99, opened 2015-11-05
Documentation for blob store administration

mosabua:
Already uses the new wording from https://github.com/sonatype/nexus-internal/pull/300

https://issues.sonatype.org/browse/NEXUS-9386

Rendered output at http://bamboo.s/browse/BOOK-NEXUS59-2/artifact/STAGE/site/3.0/reference/admin.html#admin-repository-blobstores


mosabua:
I am merging this now since we need it live.

PR #101, opened 2015-11-16
NEXUS-9443: Update documentation to reflect single-button "Invalidate cache" action for repositories

fjmilens3:
https://issues.sonatype.org/browse/NEXUS-9443

(We need to wait on https://issues.sonatype.org/browse/NEXUS-9182 to merge.)


joedragons:
Nice writeup, super clear.

PR #102, opened 2015-11-16
Rewrite based on NEXUS-9374 decision

joedragons:
In m6 you need to login before you can perform any actions against hosted/proxy repositories.  This invalidates the docker authentication section as written as you are no longer prompted (NEXUS-9374).
Creating a PR rather than committing directly since a larger revamp.

This section was also adjusted to "size 114" tho probably don't notice since almost totally rewritten.

### DIFF
--- a/chapter-docker.asciidoc
+++ b/chapter-docker.asciidoc
@@ -190,6 +190,10 @@ authentication is persisted in `~/.docker/config.json` and reused for any subseq
 repository.  Individual login operations must be performed for each repository and repository group you want to
 access in an authenticated manner.
 
+TIP: Specifically when planning to push to a repository a pre-emptive login operation is advisable as it removes
+the need for use interaction and is therefore suitable for continuous integration server setups and the
+automations scenarios.
+
 === Accessing Docker Repositories
 
 You can browse Docker repositories in the Nexus user interface and inspect the components and assets and their
--- a/chapter-docker.asciidoc
+++ b/chapter-docker.asciidoc
@@ -173,16 +173,22 @@ TIP: Check out this repository configuration demonstrated in link:https://www.yo
 
 === Authentication
 
-Before performing any command against a Nexus Docker hosted or proxy repository, you must perform a docker login 
-command such as follows:
+If access to a repository requires the user to be authenticated, `docker` queries the user for the username,
+password and email address and persists it in `~/.docker/config.json`.  Typically this is required when
+<<anonymous, anonymous access>> to the repository manager is disabled or the operation requires authentication. An
+example is a `push` operation that publishes an image to the repository.
+
+The authentication can be configured in a separate step using the `docker login` command for the desired
+repository or repository group:
 
 ----
 docker login <nexus-hostname>:<repository-port>
 ----
 
-Provide a Nexus username and password as well as an email address. This authentication is persisted in 
-`~/.docker/config.json` and reused for any subsequent interactions against that repository.  Individual logins 
-must be performed for each hosted and proxy repository.
+Provide your repository manager credentials of username and password as well as an email address. This
+authentication is persisted in `~/.docker/config.json` and reused for any subsequent interactions against that
+repository.  Individual login operations must be performed for each repository and repository group you want to
+access in an authenticated manner.
 
 === Accessing Docker Repositories
 
--- a/chapter-docker.asciidoc
+++ b/chapter-docker.asciidoc
@@ -173,9 +173,16 @@ TIP: Check out this repository configuration demonstrated in link:https://www.yo
 
 === Authentication
 
-The first invocation of any command against a Nexus Docker connector and therefore a Docker repository or repository
-group triggers a login request. Provide the Nexus username and password as well as an email address to Docker. This
-authentication is persisted in `~/.docker/config.json` and reused for any subsequent interaction.
+Before performing any command against a Nexus Docker hosted or proxy repository, you must perform a docker login 
+command such as follows:
+
+----
+docker login <nexus-hostname>:<repository-port>
+----
+
+Provide a Nexus username and password as well as an email address. This authentication is persisted in 
+`~/.docker/config.json` and reused for any subsequent interactions against that repository.  Individual logins 
+must be performed for each hosted and proxy repository.
 
 === Accessing Docker Repositories
 


chapter-docker.asciidoc
```diff
@@ -173,9 +173,16 @@ TIP: Check out this repository configuration demonstrated in link:https://www.yo
...
+Before performing any command against a Nexus Docker hosted or proxy repository, you must perform a docker login 
+command such as follows:
+
+----
+docker login <nexus-hostname>:<repository-port>
+----
+
+Provide a Nexus username and password as well as an email address. This authentication is persisted in 
+`~/.docker/config.json` and reused for any subsequent interactions against that repository.  Individual logins 
+must be performed for each hosted and proxy repository.
```
mosabua: Thats not true.. only for each port you are accessing..

joedragons: Unless I mistested, for whatever reason you don't need to login to groups.  That's why I was specific.

mosabua: That doesnt make sense. I would understand if you only have to login to whatever you need to publish to and can read on your own. But then your reference to proxy repositories is wrong above. Wdyt @kellyrob99 or @adreghiciu ?

mosabua:
According to @kellyrob99 the group should behave the same. I think we have to verify this and potentially file a bug and fix it. I will try to do that unless @joedragons beats me to it ;-)

joedragons:
Removed review label til we're sure.  Thanks.

mosabua:
So @joedragons and @kellyrob99 .. with my setup I can run

docker search 192.168.1.108:18443/postgres

searching in a group as well as a proxy repo without any login and it works just fine. I would expect that behaviour to be honest since anonymous access is allowed. So imho the first statement above is wrong. Login is not required for any command.

When I do a search against an empty hosted repo I get a 404. I would expect to get a empty search result instead.

If I turn off anonymous access and do a pull I get

$ docker pull 192.168.1.108:18443/hello-world
Using default tag: latest
Pulling repository 192.168.1.108:18443/hello-world
Could not reach any registry endpoint

Thats okay imho. 

However when I login it still wont let me pull

docker login 192.168.1.108:18443
Username: admin
Password:
Email: manfred@sonatype.com
WARNING: login credentials saved in /Users/manfred/.docker/config.json
Login Succeeded
[/Users/manfred] $docker pull 192.168.1.108:18443/hello-world
Using default tag: latest
Pulling repository 192.168.1.108:18443/hello-world
Could not reach any registry endpoint

Something really weird is going on and it does not make sense how this all works . E.g. I can still run a docker pull command and get asked for the credentials so I do not have to use login as you wrote @joedragons .

To be honest I would just change the wording to say that you can use docker login in the docs you changed. And then we have to create a bug about how anonymous access works with docker and so on and test that and potentially file bugs to fix.

mosabua:
1. The doc is definitely wrong. I can pull from a group and proxy without authenticating just fine.
2. I can also search in both of them just fine.
3. If I search in a hosted repo I get 404:

docker search 192.168.1.108:18444/centos
Error response from daemon: Unexpected status code 404

this is imho a bug. It should instead return a empty search result.

mosabua:
If I turn off anonymous access..

4. I can search and will be quoted to login and it then works fine. I do NOT need to use docker login

[/Users/manfred] $docker search 192.168.1.108:18443/test

Please login prior to search:
Username: admin
Password:
Email: manfred@sonatype.com
WARNING: login credentials saved in /Users/manfred/.docker/config.json
Login Succeeded
NAME                                DESCRIPTION                                     STARS     OFFICIAL   AUTOMATED
freewil/bitcoin-testnet-box         private, bitcoin, testnet-in-a-box              4                    [OK]
...

mosabua:
I can OPTIONALLY login in first and then subsequent commands will work. That login works for group, proxy and hosted.

docker login 192.168.1.108:18443
Username: admin
Password:
Email: m@t.c
WARNING: login credentials saved in /Users/manfred/.docker/config.json
Login Succeeded
[/Users/manfred] $docker search 192.168.1.108:18443/blue
NAME                             DESCRIPTION                                     STARS     OFFICIAL   AUTOMATED
codeskyblue/docker-gogs          docker image for gogs(the git source code ...   31                   [OK]
juanluisbaptiste/bigbluebutton   BigBlueButton Web Conferencing System           6                    [OK]

mosabua:
So in summary the doc needs to be changed to say that you can optionally use the login command. And we might want to fix the 404 that happens with hosted repos.

mosabua:
Also note that I have not tested pushing ... e.g. to a group or a proxy. That should be invalid, but I will leave that to you @joedragons

kellyrob99:
@mosabua You should only get a 404 searching hosted repos if V1 api is disabled. Pretty simple answer here: V2 does not have a search endpoint so we didn't impl it.

mosabua:
@kellyrob99 with V1 enabled it works as expected. So no need for a bug imho @joedragons

joedragons:
FYI you are not prompted when you docker push, so I believe this is still not accurate as written.  That's the whole nature of the NEXUS-9374 ticket.  Sorry this got derailed by my belief that you got prompted for proxy.  I am adjusting the testguide to match the proxy related findings but that was not the cause of this rewrite.
@mosabua I am curious if you saw otherwise because it may speak to a NEXUS-9374 intermittancy/workaround or maybe not even a bug.

mosabua:
Well .. I was prompted and wrote the docs following what I experienced.. but I did not do any push

joedragons:
Right, the example given is to push and is not true and won't be til NEXUS-9374 is fixed as far as I can tell.  "An example is a `push` operation that publishes an image to the repository."
That's my concern and why I made the adjustment to start.  I can try again if you like?

mosabua:
The sentence it meant to mean that you need to be authenticated for push.. not that you will be prompted. And that is correct imho

joedragons:
Then I think this is (still) deceivingly written.  Sentence A says docker will prompt for auth.  Sentence B says when you'll need auth.  Sentence C gives an example of when you'll need auth which then infers need for Sentence A.  And even if you disagree, Sentence A is still not happening becase of the bug.  

"If access to a repository requires the user to be authenticated, docker queries the user for the username, password and email address and persists it in ~/.docker/config.json. Typically this is required when anonymous access to the repository manager is disabled or the operation requires authentication. An example is a push operation that publishes an image to the repository."

This is no better written than the original Sentence A.

"The first invocation of any command against a Nexus Docker connector and therefore a Docker repository or repository group triggers a login request."

It is generalized but in the instances where it does not work that does not matter/help.

mosabua:
@joedragons I will revisit today and fix it up.

PR #103, opened 2015-11-18
Nexus update based on legality of Docker name usage.

jwayman:
The changes in this pull request reflect a recent update to how we should be talking about Docker per their specified guidelines.

### DIFF
--- a/chapter-docker.asciidoc
+++ b/chapter-docker.asciidoc
@@ -1,68 +1,68 @@
 [[docker]]
-== Private Docker Registry
+== Private Registry for Docker
 {inoss}
 
 [[docker-introduction]]
 === Introduction
 
-https://www.docker.com/[Docker] containers and their usage have revolutionized the way applications and the underlying
-operating system are packaged and deployed to development, testing and production systems.
+https://www.docker.com/[Docker] containers and their usage have revolutionized the way applications and the
+underlying operating system are packaged and deployed to development, testing and production systems.
 
-The creation of the http://opencontainers.org/[Open Container Initiative], and the involvement of a large number of
-stakeholders, guarantees that the ecosystem of tools around the lightweight containers and their usage will continue to
-flourish.
+The creation of the http://opencontainers.org/[Open Container Initiative], and the involvement of a large number
+of stakeholders, guarantees that the ecosystem of tools around the lightweight containers and their usage will
+continue to flourish.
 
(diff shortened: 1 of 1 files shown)


chapter-docker.asciidoc
```diff
@@ -1,68 +1,68 @@
...
+TIP: A configured context-path for the Nexus user interface does not affect the Docker connector URLs. E.g. if
+your Nexus instance is configured to be available at http://localhost:8081/nexus instead of the default root
+context http://localhost:8081/, the URLs for your Docker repositories will still only use the configured port for
+the repository and omit the context path in the URL. This is a side-effect of the the fact that Docker does not
 support context paths in the registry API.
 
-
 [[docker-registry-api]]
-=== Docker Registry API
+=== Support for ßDocker Registry API
```
joedragons: I see a weird symbol here (ßDocker).  Will take a screen in case it's not pasting.  I am looking via github.com.
screen shot 2015-11-18 at 9 27 07 am

joedragons:
+1 assuming you were looking and trusting you know the legality (I do not=)).

chapter-docker.asciidoc
```diff
@@ -1,68 +1,68 @@
...
-images within your organization in a hosted repository. Users can then launch containers based on those images,
-resulting in a completely private Docker registry with all the features available in the Nexus Repository Manager.
+This allows you to reduce time and bandwidth usage for accessing Docker images in a registry as well as share
+your images within your organization in a hosted repository. Users can then launch containers based on those
+images, resulting in a completely private Docker registry with all the features available in the Nexus Repository
+Manager.
 
 [[docker-ssl-connector]]
-=== SSL and Docker Connector Configuration
+=== SSL and Connector Configuration for Docker
```
mosabua: We might have to rename this in the user interface as well @jwayman cc @DarthHater

chapter-docker.asciidoc
mosabua: This is a "scharfes s" ... a sharp s .. which is equivalent to ss and used in German ;-)

chapter-docker.asciidoc
mosabua: Filed https://issues.sonatype.org/browse/NEXUS-9456

mosabua:
Applied with separate commit

PR #104, opened 2015-11-18
NEXUS-9448: Update Nexus 3 documentation to include "Purge Unused Maven Snapshots" task

fjmilens3:
https://issues.sonatype.org/browse/NEXUS-9448

### DIFF
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -323,10 +323,10 @@ execution. In addition, any incomplete Docker uploads that have been orphaned by
 up whenever the task executes.
 
 Purge Unused Components and Assets:: This task can be used to remove components and assets in Docker proxy
-repositories. Any component that has not been requested in the configured amount of days will be purged.
+repositories. Any component that has not been requested in the configured number of days will be purged.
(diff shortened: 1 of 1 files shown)


chapter-configuration.asciidoc
```diff
@@ -325,6 +325,9 @@ up whenever the task executes.
 Purge Unused Components and Assets:: This task can be used to remove components and assets in Docker proxy
 repositories. Any component that has not been requested in a configured amount of days, will be purged.
 
+Purge Unused Maven Snapshots:: This task can be used to remove unused snapshots from Maven repositories. Any snapshot
```
fjmilens3: This is liberally borrowed from the above description of the other purge task. I'm unsure if I should liven my text up, or aim for consistency.

joedragons: I think borrowing is fine. I prefer "the configured amount of days" but like you said you borrowed=) I also think adding "When run, any snapshop that has not..." reads nicer but again you're borrowing=)

joedragons:
+1 w/ minor note.

PR #105, opened 2015-11-24
NEXUS-8449 - Update jetty xml configuration options as it pertains to https

peterlynch:
Depends on https://github.com/sonatype/nexus-internal/pull/333

### DIFF
--- a/chapter-security.asciidoc
+++ b/chapter-security.asciidoc
@@ -441,7 +441,7 @@ the truststore file as trusted.
 
 {inall}
 
-Providing access to the user interface and content via HTTPS is a recommended best practice for any deployment.
+Providing access to the user interface and content via HTTPS is a best practice.
 
 You have two options:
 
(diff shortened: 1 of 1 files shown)


mosabua:
As agreed with @peterlynch I will review, merge and add a generic section about usage of NEXUS_HOME ... no other reviews necessary at this stage.

peterlynch:
@mosabua somewhat related, this is what I believe is a concise and clear outline of important directory terms with regards to Nexus 2

https://support.sonatype.com/hc/en-us/articles/213464308-Understanding-Nexus-Installation-Directories

mosabua:
Following that guide assumes that there is a shared top level nexus directory in which you find the application and the data (sonatype-work). From what I have seen this is often not the case and certainly not required.

mosabua:
I will also have to check what names are used in the installer to be consistent..

peterlynch:
lets not let the NEXUS_HOME related changes block this PR

mosabua:
I am not... I just got a lot of other things to do as well and we cant publish this whole set for a bit anyway. I am going to merge and update this all today.

PR #109, opened 2015-12-10
fix hidden anchor link titles

peterlynch:
An example url:

http://books.sonatype.com/nexus-book/3.0/reference/security.html#ssl-inbound

Before:

image

After:

image

### DIFF
--- a/site/css/clmBook.css
+++ b/site/css/clmBook.css
@@ -266,6 +266,15 @@ div.content + div.title {
 .header-centered {
     text-align: center;
 }
+/* align anchored links to below the fixed navbar */
+*[id]:before {
+  display: block;
+  content: " ";
+  margin-top: -55px;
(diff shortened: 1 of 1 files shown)


joedragons:
Nice, that's bugged me for a while.

peterlynch:
Please apply this to the other Nexus 2.x branches as needed.

mosabua:
This is awesome @peterlynch .. I will roll this into all the branches for the 2.12 and rebranding roll out. 

@jwayman and @annrollo should port this over to the Nexus IQ Server documentation as well.

mosabua:
Btw @jwayman .. you should include that in a commit that does the navigation refactor across branches.. that way you can cherry pick across easily and do it in one sweep. I am just doing it for 2.8+ here now.

PR #112, opened 2016-01-12
added description for Purge Orphaned API Keys task, NEXUS-9619

mosabua:
See https://issues.sonatype.org/browse/NEXUS-9619

### DIFF
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -337,9 +337,9 @@ execution. In addition, any incomplete Docker uploads that have been orphaned by
 up whenever the task executes.
 
 Purge Orphaned API Keys:: This scheduled tasks deletes old, unused API keys. These keys are generated, for
-example, when using the User Token feature or NuGet repositories. A key becomes unused, when the user account is
-deleted is deleted. The task purges these orphaned API keys and should be scheduled to run regularly to remove
-these redundant keys.
+example, when using the User Token feature or publishing to NuGet repositories. A key becomes unused, when the
+user account is deleted. The task purges these orphaned API keys and should be scheduled to run regularly to
+remove these redundant keys.
 
 Purge Unused Components and Assets:: This task can be used to remove components and assets in proxy repositories.
 Any component that has not been requested in the configured number of days will be purged.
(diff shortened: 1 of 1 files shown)


mosabua:
The only question I have is ... why do we need to delete these keys. I assume they can not be used anymore at all anyway since the user account associated with the key is deleted. Or is that wrong? @mrprescott 

If they can be used.. isn't that a security issue? And if they cant be used... is it really just simple housekeeping to use this task .. like the empty trash task basically but for keys?

chapter-configuration.asciidoc
```diff
@@ -336,12 +336,18 @@ task. You can configure the minimum age of incomplete uploads to be purged and h
 execution. In addition, any incomplete Docker uploads that have been orphaned by a Nexus restart will be cleaned
 up whenever the task executes.
 
+Purge Orphaned API Keys:: This scheduled tasks deletes old, unused API keys. These keys are generated, for
+example, when using the User Token feature or NuGet repositories. A key becomes unused, when the user account is
+deleted is deleted. The task purges these orphaned API keys and should be scheduled to run regularly to remove
```
joedragons: You have "is deleted" twice here.

chapter-configuration.asciidoc
```diff
@@ -336,12 +336,18 @@ task. You can configure the minimum age of incomplete uploads to be purged and h
...
+these redundant keys.
+
 Purge Unused Components and Assets:: This task can be used to remove components and assets in proxy repositories.
 Any component that has not been requested in the configured number of days will be purged.
 
 Purge Unused Maven Snapshots:: This task can be used to remove unused snapshots from Maven repositories. Any
 snapshot that has not been requested in the configured number of days will be purged.
 
+
```
joedragons: Minor: Not sure why this newline but it seems the end result consumes it and it only shows one.

joedragons:
+1 besides above comment.  Not sure the answer to your Q; I've assumed a cleanup task but there may be deeper reaching implications.

chapter-configuration.asciidoc
mosabua: fixed

chapter-configuration.asciidoc
mosabua: In asciidoc multiple whitespace lines just signal new paragraph... just like in latex and many other markup languages..

PR #117, opened 2016-01-21
[OSSRH-1992] Change credentials to static string to avoid breaking clients

djsauble:
Issue: https://issues.sonatype.org/browse/OSSRH-19926

Related: https://github.com/sonatype/nexus-book/pull/119

This string should never change, regardless of branding, as third-party clients depend on it. As a result, I’ve changed it from the {pro} macro to a static string.

### DIFF
--- a/chapter-maven.asciidoc
+++ b/chapter-maven.asciidoc
@@ -465,12 +465,13 @@ These minimal settings allow sbt to download the declared dependencies.
 To deploy build outputs to a Nexus repository with the +publish+
 task, user credentials can be declared in the +build.sbt+ file:
 
-// The credentials string should never change, as third-party clients depend on it
 [subs="attributes"]
 ----
 credentials += Credentials("Sonatype Nexus Repository Manager",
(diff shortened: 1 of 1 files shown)


mosabua:
The help text should NOT be in the code snippet though. You want to fix that up or should I do it? It also needs to go into the 3x branch and probably be published.

djsauble:
@mosabua Is that better?

I’ll file a PR for 3.x as well.

mosabua:
Haha ... I would add it after the code snippet and remove the comment sign (//)

mosabua:
You could make it a TIP:

djsauble:
Moved.

mosabua:
Since you break up the sentence.. you have to change to

And ...

djsauble:
kk, look good? I’ve mirrored the changes in https://github.com/sonatype/nexus-book/pull/119 too.

mosabua:
Ship it! +1

PR #123, opened 2016-01-29
added default admin username and password 

mosabua:
to various appropriate places in docs

https://issues.sonatype.org/browse/NEXUS-9743

### DIFF
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -9,11 +9,11 @@ This chapter covers all aspects of configuring {pro} and {oss}. Specifically the
 button image:figs/web/ui-admin-button-icon.png[scale=50] in the main toolbar.
 
 Many of the features shown in this section are only available to administrative users and are available in
-{oss}.
+{oss}. Some features are only available in {pro}.
 
 TIP: The default user for accessing these features has the username 'admin' and the password 'admin123'. 
 
-Some features are only available in {pro}. The 'Administration' menu contains the following sections:
+The 'Administration' menu contains the following sections:
 
 ////
 
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -9,7 +9,11 @@ This chapter covers all aspects of configuring {pro} and {oss}. Specifically the
 button image:figs/web/ui-admin-button-icon.png[scale=50] in the main toolbar.
 
 Many of the features shown in this section are only available to administrative users and are available in
-{oss}. Some features are only available in {pro}. The 'Administration' menu contains the following sections:
+{oss}.
+
+TIP: The default user for accessing these features has the username 'admin' and the password 'admin123'. 
+
+Some features are only available in {pro}. The 'Administration' menu contains the following sections:
 
 ////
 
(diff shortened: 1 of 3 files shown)


chapter-configuration.asciidoc
```diff
@@ -9,7 +9,11 @@ This chapter covers all aspects of configuring {pro} and {oss}. Specifically the
 button image:figs/web/ui-admin-button-icon.png[scale=50] in the main toolbar.
 
 Many of the features shown in this section are only available to administrative users and are available in
-{oss}. Some features are only available in {pro}. The 'Administration' menu contains the following sections:
+{oss}.
```
mrprescott: The 'only' makes this sentence confusing.

chapter-installation.asciidoc
```diff
@@ -208,8 +208,21 @@ image::figs/web/ui-overview-anonymous.png[scale=45]
 While we use +localhost+ throughout this documentation, you may need to use the IP Loopback Address of
 +127.0.0.1+, the IP address or the DNS hostname assigned to the machine running the repository manager.
 
+The user interface shows the features available to an anonymous user. The repository manager installation includes
+an administrative user with full access with the username 'admin' and the password 'admin123'. You can sign in
```
mrprescott: ..with full access. Its username is 'admin'..

chapter-configuration.asciidoc
mosabua: Reordered into the above sections so it makes sense in context.

mrprescott: Actually, this whole chunk is confusing. There seem to be four ideas:
* Many features in this section are available to admin users
* Many features are in {oss}, but some are only in {pro}
* Here's the default admin user credentials
* Here's what's in the 'Aministration' menu

I think it would be clearer to explain them in order. As it stands, ideas 1 and half of idea 2 are welded together into a sentence. We then have idea 3, then the last half of idea 2, but it's in the same paragraph as idea 4 and makes them seem unduly related.

chapter-installation.asciidoc
mosabua: Fixed

chapter-configuration.asciidoc
```diff
@@ -9,7 +9,11 @@ This chapter covers all aspects of configuring {pro} and {oss}. Specifically the
 button image:figs/web/ui-admin-button-icon.png[scale=50] in the main toolbar.
 
 Many of the features shown in this section are only available to administrative users and are available in
-{oss}. Some features are only available in {pro}. The 'Administration' menu contains the following sections:
+{oss}. Some features are only available in {pro}.
```
mrprescott: Many of the features shown in this section are available to administrative users.  Some are available in {oss}, while others are {pro} features.

chapter-configuration.asciidoc
mosabua: I already reordered.. will revisit with these ideas.

chapter-configuration.asciidoc
joedragons: I don't see the point of mentioning oss myself.  There arn't any features that are just OSS right?  Did someone suddenly get confused what documentation they were reading?=)
But maybe there is=)

joedragons: @mosabua faster than me=)

chapter-configuration.asciidoc
mosabua: I changed to use the {inall} macro and remove the sentence altogether. That makes it more consistent with the other chapters.

mosabua:
Received +1 from @mrprescott on hipchat.

PR #125, opened 2016-02-01
More bower related documentation

mosabua:
Created for @dulanism to update .. 

Rendered results at http://bamboo.s/browse/BOOK-NEXUS87

### DIFF
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -70,23 +70,21 @@ not yet installed. If not already, upgrade your Bower package to version 1.5 or 
 
 You can install the downloaded Bower resolver package using npm globally with the command:
 ----
-npm install -g bower-nexus3-resolver-1.0.0-pre.tgz
+npm install -g bower-nexus3-resolver.tgz
 ----
-
-You can also build as part of a build of the entire project by specifying the +-Pinclude-bower-resolver+ option.
 
 [[bower-resolver-config]]
 === Configuring the Bower Resolver
 
 You will need to create a +.bowerrc+ file, which you can build in JSON. The JSON in the project's +.bowerrc+ file
-must reference the resolver you installed.
+must reference the repository you configured. Adjust or modify the file to include the repository you add.
 
 Configure your +.bowerrc+ file to reference the URL for the Bower registry in the repository 
 manager:
 ----
 {
     "registry"    : "http://localhost:8081/repository/your-bower-repo",
-    "resolvers"   : ["bower-nexus3-resolver-1.0.0-pre"]
+    "resolvers"   : ["bower-nexus3-resolver"]
 }
 ----
 
@@ -103,12 +101,13 @@ You may also install the resolver on a per-project basis instead by adding it as
 === Browsing and Searching Bower Repositories
 
 You can browse Bower repositories in the user interface inspecting the components and assets and their details as 
-documented in <<browse-browse>>.
+documented in <<browse-browse>>. Further searching can be done in the user interface as described in <<search-components>>. 
 
-Components can be searched in the user interface as described in <<search-components>>. A search finds all 
+////
+A search finds all 
 components and assets that are currently stored in the repository manager, either because they have been deployed 
 to a hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
-
+////
 
 ////
 /* Local Variables: */
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -12,16 +12,17 @@ the Bower registry and other public registries without incurring repeated downlo
 be proxied in the repository manager. The proxy allows {pro} and {oss} to grab artifacts from Bower.
 
 The Bower proxy is optimized for the front-end. It downloads a component (e.g. jQuery) just once when multiple 
-packages depend on the proxy, reducing page load. To use the Bower proxy to access registries through Nexus 3, 
+packages depend on the proxy, reducing page load. To use the Bower proxy to access registries through Nexus, 
 set up a custom URL resolver that plugs into the package manager and interacts with the repository manager. The 
 resolver is an API introduced in Bower version 1.5. It fetches artifacts and version information through the 
-Nexus 3 Repository Manager, then automatically searches and saves the artifacts in the 
-repository.
+repository manager, then automatically searches and saves the artifacts in the repository.
 
 [[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up Bower proxy repository inside of the repository manager. This enables the system to proxy non-Nexus Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers, you should proxy the registry hosted at http://bower.herokuapp.com/[http://bower.herokuapp.com].
+You can set up Bower proxy repository within Nexus to access a remote repository location. For example, you can 
+configure your Bower proxy to access the public Bower registry at 
+http://bower.herokuapp.com/[http://bower.herokuapp.com] or your own private registry.
 
 Enabling rewrite allows the Bower resolver to force itself through the Nexus to retrieve an artifact. This is enabled by default.
 
@@ -62,11 +63,11 @@ tbd
 [[bower-installation]]
 === Installing the Bower Resolver
 
-The resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to facilitate 
-using packages in your tools and workflows.
+The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
 
-Before installing the package, make sure you have npm installed locally. See <<npm>> to review configuration if 
-not yet installed. If not already, upgrade your Bower package to version 1.5 or higher before you proceed to the resolver download.
+Before installing resolver, make sure you have <<npm>> installed locally. Also, be sure to upgrade your Bower 
+package to version 1.5 or higher before you proceed to the resolver download.
 
 You can install the downloaded Bower resolver package using npm globally with the command:
 ----
@@ -76,10 +77,9 @@ npm install -g bower-nexus3-resolver.tgz
 [[bower-resolver-config]]
 === Configuring the Bower Resolver
 
-You will need to create a +.bowerrc+ file, which you can build in JSON. The JSON in the project's +.bowerrc+ file
-must reference the repository you configured. Adjust or modify the file to include the repository you add.
+You need a +.bowerrc+ JSON file to use the Bower resolver. You can create a global file or modify your existing global file. Include the registry URL in the file.
 
-Configure your +.bowerrc+ file to reference the URL for the Bower registry in the repository 
+For example, here's a +.bowerrc+ file that references the URL for the Bower registry in the repository 
 manager:
 ----
 {
@@ -88,7 +88,7 @@ manager:
 }
 ----
 
-You may also install the resolver on a per-project basis instead by adding it as a +devDependency+ in your 
+You may also install the resolver on a per-project basis by adding it as a +devDependency+ in your 
 +package.json+:
 ----
 "devDependencies" : {
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -72,7 +72,7 @@ package to version 1.5 or higher before you proceed to the resolver installation
 
 You can install the downloaded Bower resolver package using npm globally with the command:
 ----
-npm install -g bower-nexus3-resolver.tgz
+npm install -g bower-nexus3-resolver
 ----
 
 [[bower-resolver-config]]
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -5,35 +5,34 @@
 [[bower-introduction]]
 === Introduction
 
-http://bower.io[Bower] is a package manager for front-end web development. Consumers of JavaScript 
-packages who routinely build projects need to keep track of and access remote packages. Using Bower 
-ensures front-end packages are up to date (or set to the specific versions you need). The package manager 
-installs the right versions of the packages you need and their dependencies. Bower can manage such components 
-as HTML, CSS, JavaScript, and images.
+http://bower.io[Bower] is a package manager for front-end web development. Consumers of JavaScript packages who
+routinely build projects need to keep track of and access remote packages. Using Bower ensures front-end packages
+are up to date (or set to the specific versions you need). The package manager installs the right versions of the
+packages you need and their dependencies. Bower can manage such components as HTML, CSS, JavaScript, and images.
 
-{pro} and {oss} support the Bower registry format for proxy repositories. This allows the repository manager 
-to take advantage of the packages in the Bower registry and other public registries without incurring repeated 
-downloads of packages, since they will be proxied in the repository manager. The proxy allows {pro} and {oss} 
-to grab artifacts from Bower. To use Bower with the Nexus 3 Repository Manager, you will need to install and configure a pluggable resolver that allows Bower to interact with the repository manager.
+{pro} and {oss} support the Bower registry format for proxy repositories. This allows the repository manager to
+take advantage of the packages in the Bower registry and other public registries without incurring repeated
+downloads of packages, since they will be proxied in the repository manager. The proxy allows {pro} and {oss} to
+grab artifacts from Bower. To use Bower with the Nexus 3 Repository Manager, you will need to install and
+configure a pluggable resolver that allows Bower to interact with the repository manager.
 
-Bower is optimized for the front-end. If multiple packages depend on a single package Bower will download a 
+Bower is optimized for the front-end. If multiple packages depend on a single package Bower will download a
 component (e.g jQuery) just once. This is known as a flat dependency graph and it helps reduce page load.
 
-[[bower-proxying]]
+[[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up a proxy repository inside of the repository manager. This enables the system 
-to proxy other Nexus Bower registries. To reduce duplicate downloads and improve download 
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up a proxy repository inside of the repository manager. This enables the system to proxy other Nexus
+Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers,
+you should proxy the registry hosted at http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower
+accesses this registry directly. You can also proxy any other registries you require.
 
-When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g. 
-tar.gz). 
+When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g.
+tar.gz).
 
-If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite 
-capabilities proxies the information directly from Bower’s registry without redirecting to the repository manager 
-to retrieve artifacts. The content will be as if navigating directly to the Bower registry in your 
-browser.
+If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite
+capabilities proxies the information directly from Bower’s registry without redirecting to the repository manager
+to retrieve artifacts. The content will be as if navigating directly to the Bower registry in your browser.
  
 Minimal configuration steps are:
 
@@ -41,7 +40,7 @@ Minimal configuration steps are:
 - Define URL for 'Remote storage' e.g. http://bower.herokuapp.com/[http://bower.herokuapp.com]
 - Select a 'Blob store' for 'Storage'
 
-[[bower-hosting]]
+[[bower-hosted]]
 === Hosting Bower Repositories
 
 A hosted Bower repository can be used to deploy your own as well as third-party components.
@@ -54,16 +53,27 @@ Minimal configuration steps are:
 - Define 'Name'
 - Select 'Blob store' for 'Storage'
 
+
+////
+[[bower-group]]
+=== Grouping Bower Repositories
+
+tbd
+
+////
+
 [[bower-installation]]
 === Installing Bower
 
-Before installing the package, make sure you have npm installed locally. See <<_introduction>> to review configuration if not yet installed and published.
+Before installing the package, make sure you have npm installed locally. See <<npm>> to review
+configuration if not yet installed and published.
 
-Bower works by fetching and installing packages from all over; the package manager automatically searches and 
-saves them in the repository. Bower keeps track of these packages in a manifest file, +bower.json+. Bower 
-provides hooks to facilitate using packages in your tools and workflows.
+Bower works by fetching and installing packages from all over; the package manager automatically searches and
+saves them in the repository. Bower keeps track of these packages in a manifest file, +bower.json+. Bower provides
+hooks to facilitate using packages in your tools and workflows.
 
-Now that your repositories are set up in Nexus 3, download the Bower package +.tgz+ file. Then install it with npm, globally:
+Now that your repositories are set up in the repository manager, download the Bower package +.tgz+ file. Then
+install it with npm, globally:
 ----
 npm install -g bower-nexus3-resolver-1.0.0-pre.tgz
 ----
@@ -73,7 +83,8 @@ You can also build as part of a build of the entire project by specifying the +-
 [[bower-resolver-config]]
 === Configuring the Bower Resolver
 
-You will need to create a +.bowerrc+ file, which you can build in JSON. The JSON in the project's +.bowerrc+ file must reference the resolver you installed.
+You will need to create a +.bowerrc+ file, which you can build in JSON. The JSON in the project's +.bowerrc+ file
+must reference the resolver you installed.
 
 Configure your +.bowerrc+ file to reference the URL for the Bower registry in the repository 
 manager:
@@ -91,6 +102,7 @@ You may also install the resolver on a per-project basis instead by adding it as
   "bower-nexus3-resolver" : "*"
 }
 ----
+
 
 [[bower-browse-search]]
 === Browsing and Searching Bower Repositories
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -67,8 +67,8 @@ tbd
 The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
 facilitate using packages in your tools and workflows.
 
-Before installing resolver, make sure you have <<npm>> installed locally. Also, be sure to upgrade your Bower 
-package to version 1.5 or higher before you proceed to the resolver download.
+Before installing resolver, make sure you have npm installed. Also, be sure to upgrade your Bower 
+package to version 1.5 or higher before you proceed to the resolver installation.
 
 You can install the downloaded Bower resolver package using npm globally with the command:
 ----
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -27,10 +27,10 @@ http://bower.herokuapp.com/[http://bower.herokuapp.com] or your own private regi
 Enabling rewrite allows the Bower resolver to force itself through Nexus to retrieve an artifact. This is enabled 
 by default.
 
-If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite
-capabilities proxies the information directly from the remote registry without redirecting to the repository 
-manager to retrieve artifacts. The content will be as if navigating directly to the Bower registry in your 
-browser.
+If deactivated, no rewrite of the URL occurs. As a result, the original component URL is exposed. Turning off 
+rewrite capabilities proxies the information directly from the remote registry without redirecting to the 
+repository manager to retrieve artifacts. The content will be as if navigating directly to the Bower registry in 
+your browser.
  
 Minimal configuration steps are:
 
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -6,29 +6,29 @@
 === Introduction
 
 http://bower.io[Bower] is a package manager for front-end web development. Consumers of JavaScript packages who
-routinely build projects need to keep track of and access remote packages. Using Bower ensures front-end packages
-are up to date (or set to the specific versions you need). The package manager installs the right versions of the
-packages you need and their dependencies. Bower can manage such components as HTML, CSS, JavaScript, and images.
+routinely build projects need to keep track of and access remote packages. {pro} and {oss} support the Bower 
+registry format for proxy repositories. This allows the repository manager to take advantage of the packages in 
+the Bower registry and other public registries without incurring repeated downloads of packages, since they will 
+be proxied in the repository manager. The proxy allows {pro} and {oss} to grab artifacts from Bower.
 
-{pro} and {oss} support the Bower registry format for proxy repositories. This allows the repository manager to
-take advantage of the packages in the Bower registry and other public registries without incurring repeated
-downloads of packages, since they will be proxied in the repository manager. The proxy allows {pro} and {oss} to
-grab artifacts from Bower.
-
-Bower is optimized for the front-end. If multiple packages depend on a single package Bower will download a
-component (e.g jQuery) just once. This is known as a flat dependency graph and it helps reduce page load.
+The Bower proxy is optimized for the front-end. It downloads a component (e.g. jQuery) just once when multiple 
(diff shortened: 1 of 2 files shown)


joedragons:
RE https://issues.sonatype.org/browse/NEXUS-9580 - note on proxy "enable rewrite" is now on/checked/enabled by default.

chapter-bower.asciidoc
```diff
@@ -5,35 +5,34 @@
...
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up a proxy repository inside of the repository manager. This enables the system to proxy other Nexus
+Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers,
+you should proxy the registry hosted at http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower
+accesses this registry directly. You can also proxy any other registries you require.
 
-When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g. 
-tar.gz). 
+When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g.
```
fjmilens3: I don't know if they care about it being a .tar.gz or not.  That's more of an implementation detail because of how Github et al work, and that's also how we intend to pattern hosting the files for hosted when we get there as well, because of that.  But I don't know if a Bower user would think of the artifact in .tar.gz terms or not?  (Thoughts, anyone?)

chapter-bower.asciidoc
```diff
@@ -5,35 +5,34 @@
...
 
-[[bower-proxying]]
+[[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up a proxy repository inside of the repository manager. This enables the system 
-to proxy other Nexus Bower registries. To reduce duplicate downloads and improve download 
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up a proxy repository inside of the repository manager. This enables the system to proxy other Nexus
```
fjmilens3: Not only does it enable the system to proxy other Nexus Bower registries, but the main reason for having the proxy is that it allows you to proxy non-Nexus Bower registries like the main one (that you've mentioned).

chapter-bower.asciidoc
```diff
@@ -54,16 +53,27 @@ Minimal configuration steps are:
...
 
-Bower works by fetching and installing packages from all over; the package manager automatically searches and 
-saves them in the repository. Bower keeps track of these packages in a manifest file, +bower.json+. Bower 
-provides hooks to facilitate using packages in your tools and workflows.
+Bower works by fetching and installing packages from all over; the package manager automatically searches and
+saves them in the repository. Bower keeps track of these packages in a manifest file, +bower.json+. Bower provides
+hooks to facilitate using packages in your tools and workflows.
 
-Now that your repositories are set up in Nexus 3, download the Bower package +.tgz+ file. Then install it with npm, globally:
+Now that your repositories are set up in the repository manager, download the Bower package +.tgz+ file. Then
```
fjmilens3: I might say something about "download the Nexus Bower resolver package .tgz file" or something explicitly.

chapter-bower.asciidoc
```diff
@@ -54,16 +53,27 @@ Minimal configuration steps are:
...
+tbd
+
+////
+
 [[bower-installation]]
 === Installing Bower
 
-Before installing the package, make sure you have npm installed locally. See <<_introduction>> to review configuration if not yet installed and published.
+Before installing the package, make sure you have npm installed locally. See <<npm>> to review
+configuration if not yet installed and published.
```
fjmilens3: I don't know if we need "and published" here.  Setting up npm on their workstation doesn't have anything to do with us publishing anything (or at that point, anything to do with Nexus at all, I don't think).

chapter-bower.asciidoc
```diff
@@ -5,35 +5,34 @@
...
-When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g. 
-tar.gz). 
+When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g.
+tar.gz).
 
-If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite 
-capabilities proxies the information directly from Bower’s registry without redirecting to the repository manager 
-to retrieve artifacts. The content will be as if navigating directly to the Bower registry in your 
-browser.
+If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite
```
fjmilens3: I like this section, to me it's probably the cleanest and shortest explanation for someone of what this thing does.

chapter-bower.asciidoc
```diff
@@ -5,35 +5,34 @@
...
+[[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up a proxy repository inside of the repository manager. This enables the system 
-to proxy other Nexus Bower registries. To reduce duplicate downloads and improve download 
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up a proxy repository inside of the repository manager. This enables the system to proxy other Nexus
+Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers,
+you should proxy the registry hosted at http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower
```
fjmilens3: I'm not sure if "By default Bower accesses this registry directly" is correct. I think you have to create a Bower proxy repo and manually key this in, it's not the default. If you mean that by default it doesn't have rewrites enabled, by default that option *is* enabled with one of @adreghiciu's more recent PRs (and either way should probably be rolled into the general description about rewrites and what that means).

chapter-bower.asciidoc
dulanism: updated per comments.

chapter-bower.asciidoc
```diff
@@ -5,35 +5,31 @@
...
 
-[[bower-proxying]]
+[[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up a proxy repository inside of the repository manager. This enables the system 
-to proxy other Nexus Bower registries. To reduce duplicate downloads and improve download 
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up a proxy repository inside of the repository manager. This enables the system to proxy non-Nexus Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers,
```
fjmilens3: I think we need to be better with this line yet, but I'm not sure how to to do it. It's my understanding that we can proxy both Nexus Bower registries and non-Nexus ones, at least of the main Bower registry/private Bower registry variety (I don't know if we can proxy an Artifactory one or if we even intend to support that behavior). 

If we just put "proxy Bower registries" do we think that's too broad/misleading? @DarthHater?

dulanism:
Whadadya think about adding a "this is what happens with the rewritten url" section within "Configuring the Bower Resolver"? As written in https://docs.sonatype.com/display/Nexus/Bower+Testing#BowerTesting-Testingpackageswithrewritesenabled cc @fjmilens3 @DarthHater

fjmilens3:
I don't know how deep of an explanation I'd give because I'm not sure the user cares about how the endpoints work behind the scenes.  It might be worth saying that it's a custom resolver that uses the pluggable resolver API introduced in Bower 1.5, and that we have additional features added to Nexus that the resolver uses to fetch artifacts and version information through Nexus when rewrites are turned on.

(Speaking of which, do we have it listed in the docs that you need Bower 1.5 or above for sure for this to work at all?)

chapter-bower.asciidoc
```diff
@@ -5,35 +5,30 @@
...
+are up to date (or set to the specific versions you need). The package manager installs the right versions of the
+packages you need and their dependencies. Bower can manage such components as HTML, CSS, JavaScript, and images.
+
+{pro} and {oss} support the Bower registry format for proxy repositories. This allows the repository manager to
+take advantage of the packages in the Bower registry and other public registries without incurring repeated
+downloads of packages, since they will be proxied in the repository manager. The proxy allows {pro} and {oss} to
+grab artifacts from Bower.
+
+Bower is optimized for the front-end. If multiple packages depend on a single package Bower will download a
 component (e.g jQuery) just once. This is known as a flat dependency graph and it helps reduce page load.
```
joedragons: Is this "Bower" within the Nexus repository manager or just a general statement of truth?  If the latter, I'm +0 to including this, I'm not sure the value of telling what bower does.  If within NXRM, I'm not sure that's clear as written.

chapter-bower.asciidoc
joedragons: I would suggest redundantly saying a "Bower" proxy here.  I also think you can combine the first two sentences into one.  For example (adjust as you see fit):
"You can set up a Bower proxy repository which enables the system to proxy non-Nexus Bower registries.  This helps (t)o reduce dup..."

chapter-bower.asciidoc
dulanism: If i understand your question correctly @joedragons , I would conclude that I can edit/downplay what Bower is and does, and instead skew it toward how it proxies and resolves.

chapter-bower.asciidoc
```diff
@@ -5,35 +5,30 @@
...
-You can set up a proxy repository inside of the repository manager. This enables the system 
-to proxy other Nexus Bower registries. To reduce duplicate downloads and improve download 
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up a proxy repository inside of the repository manager. This enables the system to proxy non-Nexus Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers,
+you should proxy the registry hosted at http://bower.herokuapp.com/[http://bower.herokuapp.com].
 
-When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g. 
-tar.gz). 
+Rewrite capabilities are a default setting. This allows the Bower resolver to force itself through the Nexus to retrieve an artifact.
```
joedragons: Suggest switching this around and telling what it does and then that it's enabled.
For example (adjust as you see fit): "Enabling rewrite allows the Bower resolver to force itself through the Nexus to retrieve an artifact.  This is enabled by default."

Note, I also removed the word capabilities because we have a section with Capabilities and this is not one of them.

I also have some concerns we talk about the resolver before we describe what it is...that seems a potential source of confusion.  I am not sure how I feel about having the resolver section earlier however.  I'd be curious to how others feel about that.
It seems like something you'd do before you were creating proxies, though I have been setting it up second myself:P  On the other hand, I already know what it is, so wouldn't be confused here.
If you don't agree at least a link to the later section might help people who are wondering what this is a chance to read before adjusting the setting.

chapter-bower.asciidoc
joedragons: That's what I would do.
I do not see the need to tell people how Bower functions (outside of Nexus).
I'd gander the number of people that come to documentation trying to figure
out what format is best for their company is super small.  If people are
reading this, I'd say they're using Bower already.
Now in connection to Nexus yes the facts are good, but see no harm in
making those sentences clear.
2 cents.

chapter-bower.asciidoc
```diff
@@ -5,35 +5,30 @@
...
 
-If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite 
-capabilities proxies the information directly from Bower’s registry without redirecting to the repository manager 
-to retrieve artifacts. The content will be as if navigating directly to the Bower registry in your 
-browser.
+If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite
+capabilities proxies the information directly from Bower’s registry without redirecting to the repository manager
```
joedragons: Minor: Suggest switch "Bower's" with "the remote".  "Bower's" to me infers there's just one (like in the case of bower.io) or seems non-helpful since we're in the Bower section.

chapter-bower.asciidoc
```diff
@@ -53,17 +49,32 @@ Minimal configuration steps are:
...
+////
 
 [[bower-installation]]
-=== Installing Bower
+=== Installing the Bower Resolver
 
-Before installing the package, make sure you have npm installed locally. See <<_introduction>> to review configuration if not yet installed and published.
+To use Bower with the Nexus 3 Repository Manager, you will need to install and configure a custom resolver that 
+plugs into the package manager and interacts with the repository manager. The resolver is an API introduced in 
+Bower version 1.5. It fetches artifacts and version information through Nexus 3 Repository Manager, then
```
joedragons: While true, I'm not convinced this highlights they need Bower v1.5+ to have this work.  Be curious to others thoughts.  This may be enough if we have a KB article or something but seems buried and not prounounced to me otherwise.

I also don't think we need to say "3" since this is the "3" documentation and it's not included in the "2" documentation.

chapter-bower.asciidoc
joedragons: I just (re)remembered that in the resolver config, you need the repositories name.  So this order is probably right.  Still may be worth a link tho.

chapter-bower.asciidoc
joedragons: Okie, I see this is highlighted more clearly further down.

chapter-bower.asciidoc
```diff
@@ -53,17 +49,32 @@ Minimal configuration steps are:
...
 
-Bower works by fetching and installing packages from all over; the package manager automatically searches and 
-saves them in the repository. Bower keeps track of these packages in a manifest file, +bower.json+. Bower 
-provides hooks to facilitate using packages in your tools and workflows.
+The resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to facilitate 
+using packages in your tools and workflows.
 
-Now that your repositories are set up in Nexus 3, download the Bower package +.tgz+ file. Then install it with npm, globally:
+Before installing the package, make sure you have npm installed locally. See <<npm>> to review configuration if 
+not yet installed. If not already, upgrade your Bower package to version 1.5 or higher before you proceed to the resolver download.
```
joedragons: As far as I know, configuring npm via Nexus is not needed to install this.  I would rephrase "See <<npm>>..." to something like "If you are using npm through Nexus, you will need to make sure you have a proxy to http://registry.npmjs.org/ to deal with the dependancies in the resolver.  See <<npm>> for more details on configuration of npm within Nexus."

The other steps they need to do is make sure they're logged in and have the Realm enabled but if they have not done that, they are really not using npm through Nexus.  My proposed rewrite is assuming they've done some minimal work and/or should read the entire npm chapter=)

chapter-bower.asciidoc
joedragons: OTOH, you have downloading and configuring seperately, so maybe reorganizing would work.

chapter-bower.asciidoc
```diff
@@ -53,17 +49,32 @@ Minimal configuration steps are:
...
+
+When your repositories are set up in the repository manager, download the Bower resolver package +.tgz+ file. 
+Then install it with npm, globally:
```
joedragons: As hinted above, as far as I know there's no requirement to have repositories setup before they install the resolver.  You can probably simplify these sentences to "You can install the downloaded Bower resolver package using npm globally with the command:"

chapter-bower.asciidoc
```diff
@@ -53,17 +49,32 @@ Minimal configuration steps are:
...
 ----
 npm install -g bower-nexus3-resolver-1.0.0-pre.tgz
```
joedragons: Unless @fjmilens3 made some discoveries yesterday or we're renaming it, this doesn't match what we had initially planned to distribute to Cigna.
Potentially you could also add the word "like" to my recommended sentence switch above, which would then hopefully prompt them to use the correct package name if literally following along=)

joedragons:
While not changed Line 82 reads "You can also build as part of a build of the entire project by specifying the +-Pinclude-bower-resolver+ option."  I don't think anyone will (currently) be building NX3 and think this line can be struck as internal.  @fjmilens3 can probably correct me if I'm wrong.

chapter-bower.asciidoc
```diff
@@ -73,14 +84,15 @@ You can also build as part of a build of the entire project by specifying the +-
 [[bower-resolver-config]]
 === Configuring the Bower Resolver
 
-You will need to create a +.bowerrc+ file, which you can build in JSON. The JSON in the project's +.bowerrc+ file must reference the resolver you installed.
+You will need to create a +.bowerrc+ file, which you can build in JSON. The JSON in the project's +.bowerrc+ file
```
joedragons: They can create or if they have an existing, they can modify it.
I suggest changing to something like: "You will need a .bowerrc file to use the Bower resolver.  You can create a global one or modify your existing global one or create one on a per project basis, depending on how you have Bower configured."
I do not feel strongly about saying it can be built in JSON since we give them what they need to put in and again I feel this is something they would(should?) kinda know.  Add it back in if you like etc=)

chapter-bower.asciidoc
fjmilens3: +1. I think we may want to manually rename the .tgz before uploading so that it matches to avoid confusion, but I still like the idea of adding this change.

chapter-bower.asciidoc
```diff
@@ -73,14 +84,15 @@ You can also build as part of a build of the entire project by specifying the +-
...
+must reference the resolver you installed.
```
joedragons: This is actually not totally true (see further comment below).  However as important you must reference the repository you configured, so it can be adjusted to that or modified to include that.

chapter-bower.asciidoc
```diff
@@ -73,14 +84,15 @@ You can also build as part of a build of the entire project by specifying the +-
...
 
 Configure your +.bowerrc+ file to reference the URL for the Bower registry in the repository 
 manager:
 ----
 {
     "registry"    : "http://host:port/repository/your-bower-repo",
-    "resolvers"   : "bower-nexus3-resolver-1.0.0-pre"
+    "resolvers"   : ["bower-nexus3-resolver-1.0.0-pre"]
```
joedragons: Double checked with @fjmilens3 and the versioning "-1.0.0-pre" is not correct here.  It's just "resolvers"   : [ "bower-nexus3-resolver" ]

fjmilens3:
Regarding @joedragons' comment, I thought we were going to strike out the whole part about building the resolver and the Maven builds and all that (in fact, I thought we already had?).

chapter-bower.asciidoc
fjmilens3: Correct.

chapter-bower.asciidoc
fjmilens3: What about "the proxied Bower registry"?

chapter-bower.asciidoc
fjmilens3: I agree here. I'd assume that if they're reading this, they already know what Bower is and what they want to do with it. I think I assumed that this was in here more for marketing reasons than technical ones?

chapter-bower.asciidoc
joedragons: Also good=)

chapter-bower.asciidoc
fjmilens3: I think this was added because @dulanism is stating that you need to have npm installed locally on your machine to be able to install the resolver. 

I do like the note about if you're using npm through Nexus that you need to have it proxying the npm registry, though. Even though we don't have that in place yet for the resolver, when we ship we'll have to, and that seems like it could be an easy mistake for someone to make at a company where they try to keep their developer team hermetically-sealed.

chapter-bower.asciidoc
fjmilens3: No technical requirement, no, but the resolver can't work without the repositories being set up (you'll either get errors or get erroneous 404s). I think we still want to explain how the resolver works after we've explained how the repos work, but I can definitely agree with shortening that as you suggest.

chapter-bower.asciidoc
joedragons: *nods* In brief, my concern is as written (I believe) we say "You need npm" then "If you don't have npm, see our Nexus npm section".  But the Nexus npm section has a LOT more detail than they need to do this.  They can just download npm (not even using Nexus) and be good to go.

joedragons:
More things noticed but not changed specifically by this PR.

Minor: Line 114 reads "Components can be searched in the user interface as described in <<search-components>>." however (as noted in the following sentence) both components and assets can be done in this way.  Suggest minor rewrite to something like "Further searching can be done in the user interface..."

Minor: Line 116 mentions hosted ("to a hosted repository") which we have struck other references to because it's not completely ready yet.  I believe this can be left as it is true (in general for search) however wanted to mention since I noticed in case an oversight and anyone disagrees.

joedragons:
I think it was, I think this is residual.

chapter-bower.asciidoc
fjmilens3: @joedragons, I'm pretty sure that proxying a Nexus Bower repo is exactly what `NexusBowerResolver` is intended to do.

joedragons: Fair enough.  I'm good with broad and tightening if there's confusion later but I defer to @DarthHater .

chapter-bower.asciidoc
```diff
@@ -5,35 +5,30 @@
...
-[[bower-proxying]]
+[[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up a proxy repository inside of the repository manager. This enables the system 
-to proxy other Nexus Bower registries. To reduce duplicate downloads and improve download 
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up a proxy repository inside of the repository manager. This enables the system to proxy non-Nexus Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers,
+you should proxy the registry hosted at http://bower.herokuapp.com/[http://bower.herokuapp.com].
```
joedragons: npm doc section says "By default npm accesses this registry directly."  I believe this is true here of Bower, so adding that kinda note (By default Bower accesses this registry directly.") might add some emphasis discussed desired above.

chapter-bower.asciidoc
dulanism: Correct

chapter-bower.asciidoc
dulanism: Corrected

chapter-bower.asciidoc
```diff
@@ -5,34 +5,29 @@
...
+routinely build projects need to keep track of and access remote packages. {pro} and {oss} support the Bower 
+registry format for proxy repositories. This allows the repository manager to take advantage of the packages in 
+the Bower registry and other public registries without incurring repeated downloads of packages, since they will 
+be proxied in the repository manager. The proxy allows {pro} and {oss} to grab artifacts from Bower.
+
+The Bower proxy is optimized for the front-end. It downloads a component (e.g. jQuery) just once when multiple 
+packages depend on the proxy, reducing page load. To use the Bower proxy to access registries through Nexus 3, 
+set up a custom URL resolver that plugs into the package manager and interacts with the repository manager. The 
+resolver is an API introduced in Bower version 1.5. It fetches artifacts and version information through the 
+Nexus 3 Repository Manager, then automatically searches and saves the artifacts in the
```
joedragons: I don't think you need to mention "3" here.  I mentioned this in another place I believe.

chapter-bower.asciidoc
```diff
@@ -5,34 +5,29 @@
...
-
-[[bower-proxying]]
+http://bower.io[Bower] is a package manager for front-end web development. Consumers of JavaScript packages who
+routinely build projects need to keep track of and access remote packages. {pro} and {oss} support the Bower 
+registry format for proxy repositories. This allows the repository manager to take advantage of the packages in 
+the Bower registry and other public registries without incurring repeated downloads of packages, since they will 
+be proxied in the repository manager. The proxy allows {pro} and {oss} to grab artifacts from Bower.
+
+The Bower proxy is optimized for the front-end. It downloads a component (e.g. jQuery) just once when multiple 
+packages depend on the proxy, reducing page load. To use the Bower proxy to access registries through Nexus 3,
```
joedragons: I don't think you need to mention "3" here. I mentioned this in another place I believe.

chapter-bower.asciidoc
```diff
@@ -5,34 +5,29 @@
...
+repository.
+
+[[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up a proxy repository inside of the repository manager. This enables the system 
-to proxy other Nexus Bower registries. To reduce duplicate downloads and improve download 
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up Bower proxy repository inside of the repository manager. This enables the system to proxy non-Nexus Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers, you should proxy the registry hosted at http://bower.herokuapp.com/[http://bower.herokuapp.com].
```
joedragons: Still not convinced about this second sentence, since you probably can proxy Nexus Bower registries...but if you remove "non-Nexus" then it basically is redundant with the first sentence. Maybe unclear what's trying to be said here.

chapter-bower.asciidoc
dulanism: Slashing "3".

chapter-bower.asciidoc
dulanism: I was using the same verbiage from other proxying sections of the Nexus Book. I can simplify the text to highlight that the Bower proxy can access both public and private registries. Henceforth this proposed revision:
"You can set up Bower proxy repository within Nexus to access a remote repository location. For example, you can configure your Bower proxy to access the public Bower registry at 
http://bower.herokuapp.com or a private registry at 
https://www.npmjs.com/package/private-bower."

chapter-bower.asciidoc
dulanism: Slashed "3" here too

chapter-bower.asciidoc
joedragons: I like that better, except I'd change the "a private registry at" to "a private registry using" or just "your own private registry" (no example).  Reason: the first is a link to the repo, the second is a link to something you use to create the repo.

chapter-bower.asciidoc
dulanism: Updated.

chapter-bower.asciidoc
```diff
@@ -53,34 +49,42 @@ Minimal configuration steps are:
...
 ----
 
-You can also build as part of a build of the entire project by specifying the +-Pinclude-bower-resolver+ option.
-
 [[bower-resolver-config]]
 === Configuring the Bower Resolver
 
-You will need to create a +.bowerrc+ file, which you can build in JSON. The JSON in the project's +.bowerrc+ file must reference the resolver you installed.
+You will need to create a +.bowerrc+ file, which you can build in JSON. The JSON in the project's +.bowerrc+ file
+must reference the repository you configured. Adjust or modify the file to include the repository you add.
```
joedragons: @dulanism Thanks.  For your future awareness, this will probably eventually be group and/or we may want to add a section telling how they can do multiple proxies etc.  For now this should be good, I think.

dulanism: Noting this for the not-so-distant future. Thx @joedragons

chapter-bower.asciidoc
```diff
@@ -5,34 +5,29 @@
...
+You can set up Bower proxy repository inside of the repository manager. This enables the system to proxy non-Nexus Bower registries. To reduce duplicate downloads and improve download speeds for your developers and CI servers, you should proxy the registry hosted at http://bower.herokuapp.com/[http://bower.herokuapp.com].
 
-When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g. 
-tar.gz). 
+Enabling rewrite allows the Bower resolver to force itself through the Nexus to retrieve an artifact. This is enabled by default.
 
-If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite 
-capabilities proxies the information directly from Bower’s registry without redirecting to the repository manager 
-to retrieve artifacts. The content will be as if navigating directly to the Bower registry in your 
+If deactivated, no rewrite of the URL occurs. As a result, original component URL is exposed. Turning off rewrite
```
joedragons: Miss the word "the" before original?

chapter-bower.asciidoc
dulanism: Okay. Updated.

chapter-bower.asciidoc
```diff
@@ -5,34 +5,30 @@
...
-to proxy other Nexus Bower registries. To reduce duplicate downloads and improve download 
-speeds for your developers and CI servers, you should proxy the registry hosted at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com]. By default Bower accesses this registry directly. You can also proxy any other registries you require.
+You can set up Bower proxy repository within Nexus to access a remote repository location. For example, you can 
+configure your Bower proxy to access the public Bower registry at 
+http://bower.herokuapp.com/[http://bower.herokuapp.com] or your own private registry.
 
-When rewrite capabilities are activated the Bower resolver forces itself to Nexus to retrieve an artifact (e.g. 
-tar.gz). 
+Enabling rewrite allows the Bower resolver to force itself through the Nexus to retrieve an artifact. This is enabled by default.
```
DarthHater: Sentence reads odd at "through the Nexus".

dulanism: Whoops. Sounds like a song. Fixed @DarthHater

chapter-bower.asciidoc
```diff
@@ -53,38 +51,45 @@ Minimal configuration steps are:
...
+=== Installing the Bower Resolver
 
-Before installing the package, make sure you have npm installed locally. See <<_introduction>> to review configuration if not yet installed and published.
+The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
 
-Bower works by fetching and installing packages from all over; the package manager automatically searches and 
-saves them in the repository. Bower keeps track of these packages in a manifest file, +bower.json+. Bower 
-provides hooks to facilitate using packages in your tools and workflows.
+Before installing resolver, make sure you have <<npm>> installed locally. Also, be sure to upgrade your Bower
```
joedragons: Minor: I think this will show up fine in the website but in the pdf this shows as "you have Chapter 10 installed locally".

chapter-bower.asciidoc
```diff
@@ -53,38 +51,45 @@ Minimal configuration steps are:
...
 
-Before installing the package, make sure you have npm installed locally. See <<_introduction>> to review configuration if not yet installed and published.
+The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
 
-Bower works by fetching and installing packages from all over; the package manager automatically searches and 
-saves them in the repository. Bower keeps track of these packages in a manifest file, +bower.json+. Bower 
-provides hooks to facilitate using packages in your tools and workflows.
+Before installing resolver, make sure you have <<npm>> installed locally. Also, be sure to upgrade your Bower 
+package to version 1.5 or higher before you proceed to the resolver download.
```
joedragons: Minor: I think "download" should be "install"?  Even when we have the resolver in npm, it'd be installing it not downloading it.

joedragons:
+1 w/ final minor comments.  Thanks!!

chapter-bower.asciidoc
dulanism: Just saw that. I'm gonna remove the link/macro.

chapter-bower.asciidoc
```diff
@@ -53,38 +51,45 @@ Minimal configuration steps are:
...
 
-Now that your repositories are set up in Nexus 3, download the Bower package +.tgz+ file. Then install it with npm, globally:
+You can install the downloaded Bower resolver package using npm globally with the command:
 ----
-npm install -g bower-nexus3-resolver-1.0.0-pre.tgz
+npm install -g bower-nexus3-resolver.tgz
```
fjmilens3: Are we just going to send it over as bower-nexus3-resolver.tgz? I'm not sure what the best way to handle this is right now, because in the end we won't be distributing it like this.  If we're okay with that, I'm not going to worry about it because we can't make everything perfect yet.

fjmilens3:
+1 with one comment.

chapter-bower.asciidoc
dulanism: No. I'll remove it. Thx

PR #126, opened 2016-02-01
Added short config tips 

mosabua:
for for jvm memory, port, context and data directory.

https://issues.sonatype.org/browse/NEXUS-9707

Rendered results at http://bamboo.s/browse/BOOK-NEXUS88

This one is for you @aavritzer ;-)

### DIFF
--- a/chapter-installation.asciidoc
+++ b/chapter-installation.asciidoc
@@ -296,9 +296,9 @@ directory'. This can be achieved by updating the paths in the properties `karaf.
 Configuring the specifics of the repository manager runtime involves configuration for all components in various
 configuration files and startup scripts. This section details these and provides recipes for specific tasks.
 
-The startup of the JVM running the repository manager is managed via files in the `$install-dir/bin` directory within the
-installation.  The application startup is performed with the JVM configured via the `JAVA_HOME` environment
-variable and the configuration in the file `$install-dir/bin/nexus.vmoptions` :
+The startup of the JVM running the repository manager is managed via files in the `$install-dir/bin` directory
+within the installation.  The application startup is performed with the JVM configured via the `JAVA_HOME`
+environment variable and the configuration in the file `$install-dir/bin/nexus.vmoptions` :
 
 ----
 -Xms256M
@@ -334,15 +334,46 @@ context path and host.
 
 system.properties:: Configuration parameters used for the JVM and application start up.
 
-////
-TBD
-Maybe this belongs into the configuration chapter but here are some ideas
-Small recipes
-- configure Java runtime
-- configure context
-- configure port
-- configure data directory location
-////
+==== Updating Memory Allocation and other JVM Paramaters
+
+The default memory  allocation for the repository manager is a lower value of `768kb`, suitable for most usage
+patterns. As a Java application running on the JVM the repository manager is using JVM configuration parameters
+for numerous settings as part of the startup parameters for the JVM. These values are defined in the configuration file 
+ `$install-dir/bin/nexus.vmoptions`. Increased memory configuration can be set with e.g. :
+
+----
+-Xms512M
+-Xmx2G
+----
+
+Other JVM parameters such as GC algorithm can be configured in the same location.
+
+
+==== Changing HTTP Port and Context
+
+The default value for the HTTP port used to access the repository manager user interface and resources is
+`8081`. In addition, the root context is used. The user interface is therefore available at
+e.g. `http://localhost:8081/`. Port and context are configured in `$install-dir/etc/org.sonatype.nexus.cfg`.
+A suitable configuration to expose the user interface at `http://localhost:9081/repository/` is:
+
+----
+application-port=9081
+nexus-context-path=/repository/
+----
+
+==== Configuring Data Storage Location
+
+<<installation-archive,Bundle archive installation>> of the repository manager configures the location of the
+storage for binary blobs in the `data` directory, which is nested inside the application directory. The
+<<installation-installer,installer application>> allows the user to configure the location of this directory to be
+any path. The configuration of this folder is located in `$install-dir/bin/nexus.vmoptions`. For example, if you
+want to use the absolute path `/opt/repository/storage/`, you have to change to:
+
+----
+-Dkaraf.data=/opt/repository/storage
+-Dkaraf.instances=/opt/repository/storage/instances
+-Djava.io.tmpdir=/opt/repository/storage/tmp
+----
 
 ////
 /* Local Variables: */
--- a/chapter-installation.asciidoc
+++ b/chapter-installation.asciidoc
@@ -14,7 +14,7 @@ manager running successfully in production deployments.
 [[install-sect-downloading]]
 === Downloading
 
-{oss} can be downloaded from https://support.sonatype.com/forums/23079216[Sonatype]. Bundles are available for the
+{oss} can be downloaded from https://support.sonatype.com/forums/23079216[Sonatype]. Distributions are available for the
 64-bit versions for Apple OSX, Microsoft Windows and Unix/Linux. They contain all necessary resources to install and run
 the repository manager. You can download a plain archive file or an installer for your operating system of a specific release
 version.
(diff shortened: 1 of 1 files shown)


chapter-installation.asciidoc
```diff
@@ -334,15 +334,46 @@ context path and host.
...
+storage for binary blobs in the `data` directory, which is nested inside the application directory. The
+<<installation-installer,installer application>> allows the user to configure the location of this directory to be
+any path. The configuration of this folder is located in `$install-dir/bin/nexus.vmoptions`. For example, if you
+want to use the absolute path `/opt/repository/storage/`, you have to change to:
+
+----
+-Dkaraf.data=/opt/repository/storage
+-Dkaraf.instances=/opt/repository/storage/instances
+-Djava.io.tmpdir=/opt/repository/storage/tmp
+----
```
mosabua: Not sure if there is any other configuration. This seemed to work for me.

chapter-installation.asciidoc
```diff
@@ -296,9 +296,9 @@ directory'. This can be achieved by updating the paths in the properties `karaf.
 Configuring the specifics of the repository manager runtime involves configuration for all components in various
 configuration files and startup scripts. This section details these and provides recipes for specific tasks.
```
mosabua: Just reformatted..

chapter-installation.asciidoc
```diff
@@ -334,15 +334,46 @@ context path and host.
...
-Maybe this belongs into the configuration chapter but here are some ideas
-Small recipes
-- configure Java runtime
-- configure context
-- configure port
-- configure data directory location
-////
+==== Updating Memory Allocation and other JVM Paramaters
+
+The default memory  allocation for the repository manager is a lower value of `768kb`, suitable for most usage
```
mrprescott: Wonder if we should standardize on the jdk's names for these settings. I think we're tweaking 'initial heap size', rather than the memory allocation for the repository manager (which is not quite the same thing).

kellyrob99: I think 768kb is WAY too much memory ;)

chapter-installation.asciidoc
```diff
@@ -334,15 +334,46 @@ context path and host.
...
+A suitable configuration to expose the user interface at `http://localhost:9081/repository/` is:
+
+----
+application-port=9081
+nexus-context-path=/repository/
+----
+
+==== Configuring Data Storage Location
+
+<<installation-archive,Bundle archive installation>> of the repository manager configures the location of the
```
kellyrob99: This is a bit misleading, as the data directory covers all files used by the system, and only the default shipped blobstore. So calling it 'binary blobs' is not terribly correct. 
Perhaps: system configuration and runtime files, and the default location for binary blobs (with a reference perhaps to Blobstore config?)

kellyrob99:
+1 once comments are addressed

chapter-installation.asciidoc
mrprescott: This feels to me like it has one foot in NX2's layout. The <installDir>/data directory contains blobs, but also a ton of other dynamically updated data - Karaf's bundle cache, all of our orient DBs (for config, component metadata, etc.), logs and so on.

Tweaking karaf.data will move all of this stuff, not just the (default) storage for binary blobs.

mrprescott: Beaten to the punch.

chapter-installation.asciidoc
jdillon: 768m not kb!

chapter-installation.asciidoc
kellyrob99: @mrprescott We can both hit him - it's not frowned upon :+1:

chapter-installation.asciidoc
```diff
@@ -296,9 +296,9 @@ directory'. This can be achieved by updating the paths in the properties `karaf.
 Configuring the specifics of the repository manager runtime involves configuration for all components in various
 configuration files and startup scripts. This section details these and provides recipes for specific tasks.
 
-The startup of the JVM running the repository manager is managed via files in the `$install-dir/bin` directory within the
-installation.  The application startup is performed with the JVM configured via the `JAVA_HOME` environment
-variable and the configuration in the file `$install-dir/bin/nexus.vmoptions` :
+The startup of the JVM running the repository manager is managed via files in the `$install-dir/bin` directory
+within the installation.  The application startup is performed with the JVM configured via the `JAVA_HOME`
```
jdillon: the search order of the JVM is a bit more complex than JAVA_HOME and may not be used at all of JRE is bundled

chapter-installation.asciidoc
```diff
@@ -334,15 +334,46 @@ context path and host.
...
+==== Changing HTTP Port and Context
+
+The default value for the HTTP port used to access the repository manager user interface and resources is
+`8081`. In addition, the root context is used. The user interface is therefore available at
+e.g. `http://localhost:8081/`. Port and context are configured in `$install-dir/etc/org.sonatype.nexus.cfg`.
+A suitable configuration to expose the user interface at `http://localhost:9081/repository/` is:
+
+----
+application-port=9081
+nexus-context-path=/repository/
```
jdillon: /repository/ is a bad example, as you'd end up with urls like `http://localhost:8081/repository/repository/maven-central`

chapter-installation.asciidoc
jdillon: we'll hopefully be changing to avoid needing -Dkaraf.instances here, but for now its rquired.

chapter-installation.asciidoc
jdillon: and yes, should use the names that folks know... heap size.

chapter-installation.asciidoc
jdillon: I would really like to stop using "bundle" to refer to distributions or distribution archives.  This is already complex in terms of plugin bundles in NX2 and now more confusing in terms of osgi bundles in NX3.

chapter-installation.asciidoc
mosabua: Ack

chapter-installation.asciidoc
mosabua: haha.. didnt think of that. So true. How about I change to /storage/ or so. I dont want to use /nexus/ .. that seems bad.

jdillon: Folks will either want to leave the context alone (recommended) or change it to /nexus for legacy comfort, so I don't see why we'd invent something new

chapter-installation.asciidoc
mosabua: Hm... so how can it be  configured if not via JAVA_HOME? Btw. I just reformatted this existing paragraph but I might as well fix it.

jdillon: install4j has a search order it uses, one of which is checking for JAVA_HOME.  I would have to dig a bit to find out what the order is, though the script on mac/unix should inform, I dont' recall what it is on windows atm

chapter-installation.asciidoc
mosabua: All those are fixed. The 768kb refers to the 68k Assembler Edition of Nexus

chapter-installation.asciidoc
mosabua: or maybe /components/

jdillon: Any time I've ever seen anyone ask about the web context-path in NX3 was to change it to /nexus so I don't see why we need to make this more complex.  If you want for demo purposes use something generic like /custom-context-path.  Folks that want/need to change this will know what they want/need.  Folks that don't want/need this should just leave it along and trying to invent a new prefix could mislead folks into thinking they have to change it which they shouldn't unless they have very specific deployment requirements.

chapter-installation.asciidoc
mosabua: All done.

chapter-installation.asciidoc
mosabua: I will check it out.

chapter-installation.asciidoc
mosabua: So you are okay with changing the example to /nexus/ or do you want /custom-context-path/ ?

chapter-installation.asciidoc
mosabua: I am taking this out in this context and will create another issue that documents how to configure the JVM to be used in this same section.

mosabua: https://issues.sonatype.org/browse/NEXUS-9760

PR #133, opened 2016-02-10
Add Hosted and Group Repo entries to bower doc

dulanism:
Add Hosted and Group Repo entries to bower doc

CI: http://bamboo.s/browse/BOOK-NEXUS94
JIRA: https://issues.sonatype.org/browse/NEXUS-9579
https://issues.sonatype.org/browse/NEXUS-9581

### DIFF
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -45,7 +45,7 @@ Minimal configuration steps are:
 
 Creating a Bower hosted repository allows you to register packages in the repository manager. The hosted 
 repository acts as an authoritative location for these components. This effectively creates an asset that becomes 
-a pointer to an external URL (such as a git repository).
+git a pointer to an external URL (such as a git repository).
 
 To add a hosted Bower repository, create a new repository with the recipe 'bower (hosted)' as 
 documented in <<admin-repositories>>.
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -17,6 +17,43 @@ set up a custom URL resolver that plugs into the package manager and interacts w
 resolver is an API introduced in Bower version 1.5. It fetches artifacts and version information through the 
 repository manager, then automatically searches and saves the artifacts in the repository.
 
+[[bower-resolver-config]]
+=== Configuring the Bower Resolver
+
+You need a +.bowerrc+ JSON file to use the Bower resolver. You can create a new or modify an existing global 
+file. Include the registry URL in the file.
+
+Create a +.bowerrc+ file that points to the URL of a Bower group that will merge your proxy and hosted 
+repositories. The reference to the resolver will also be added, after installed. See the example, below.
+----
+{
+    "registry"    : "http://localhost:8081/repository/bower-group",
+	"resolvers"   : ["bower-nexus3-resolver"]
+}
+---- 
+
+You may also install the resolver on a per-project basis by adding it as a +devDependency+ in your 
++package.json+:
+----
+"devDependencies" : {
+  "bower-nexus3-resolver" : "*"
+}
+----
+
+[[bower-installation]]
+=== Installing the Bower Resolver
+
+The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
+
+Before installing resolver, make sure you have npm installed. Also, be sure to upgrade your Bower 
+package to version 1.5 or higher.
+
+You can install the downloaded Bower resolver package using npm globally with the command:
+----
+npm install -g bower-nexus3-resolver
+----
+
 [[bower-proxy]]
 === Proxying Bower Repositories
 
@@ -42,9 +79,8 @@ Minimal configuration steps are:
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components. You can register a 
-Bower package to install components (such as libraries or frameworks) against a hosted repository in Nexus. This 
-needs to be flexible to allow users to register against git repos
+A hosted Bower repository can be used to map to private Bower registries. You can register a Bower package to 
+install components (such as libraries or frameworks) against a hosted repository in Nexus.
 
 To create the Bower package run the command +bower register <packagename> <giturl>+ against a hosted repository. 
 Nexus will create an asset that point to the Git URL. A user will be able to browse Bower packages assets that are registered to a hosted repository.
@@ -62,10 +98,10 @@ Minimal configuration steps are:
 === Grouping Bower Repositories
 
 A repository group is the recommended way to expose all your Bower repositories from the repository manager to 
-your users, without needing any further client side configuration. A repository group allows you to expose the 
-aggregated content of multiple proxy and hosted repositories as well as other repository groups with one URL for 
+your users, with minimal additional client side configuration. A repository group allows you to expose the 
+aggregated content of multiple proxy and hosted repositories as well as other repository groups with one URL in 
 tool configuration. This is possible for Bower repositories by creating a new repository with the 'bower (group)' 
-recipe recipe as documented in <<admin-repositories>>.
+recipe as documented in <<admin-repositories>>.
 
 Minimal configuration steps are:
 
@@ -73,51 +109,17 @@ Minimal configuration steps are:
 - Select 'Blob store' for 'Storage'
 - Add Bower repositories to the 'Members' list in the desired order
 
-Bower groups can contain hosted or proxy bower repositories. Groups operate by order of selection. In other words, the first asset in a given group is accessed; and each subsequent group is accessed, accordingly. If there are two components of the same name, the first asset will be grabbed, and so on.
-
-
-[[bower-installation]]
-=== Installing the Bower Resolver
-
-The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
-facilitate using packages in your tools and workflows.
-
-Before installing resolver, make sure you have npm installed. Also, be sure to upgrade your Bower 
-package to version 1.5 or higher before you proceed to the resolver installation.
-
-You can install the downloaded Bower resolver package using npm globally with the command:
-----
-npm install -g bower-nexus3-resolver
-----
-
-[[bower-resolver-config]]
-=== Configuring the Bower Resolver
-
-You need a +.bowerrc+ JSON file to use the Bower resolver. You can create a global file or modify your existing global file. Include the registry URL in the file.
-
-For example, here's a +.bowerrc+ file that references the URL for the Bower registry in the repository 
-manager:
-----
-{
-    "registry"    : "http://localhost:8081/repository/your-bower-repo",
-    "resolvers"   : ["bower-nexus3-resolver"]
-}
-----
-
-You may also install the resolver on a per-project basis by adding it as a +devDependency+ in your 
-+package.json+:
-----
-"devDependencies" : {
-  "bower-nexus3-resolver" : "*"
-}
-----
+Bower groups can contain hosted or proxy repositories. Groups operate by order of selection. In other words, the 
+first member in a given group is accessed; and each subsequent group is accessed, accordingly. If there are two 
+components of the same name, the first will be grabbed, and so on.
 
 
 [[bower-browse-search]]
 === Browsing and Searching Bower Repositories
 
 You can browse Bower repositories in the user interface inspecting the components and assets and their details as 
-documented in <<browse-browse>>. Further searching can be done in the user interface as described in <<search-components>>. 
+documented in <<browse-browse>>. Further searching can be done in the user interface as described in 
+<<search-components>>. 
 
 ////
 A search finds all 
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -17,29 +17,6 @@ set up a custom URL resolver that plugs into the package manager and interacts w
 resolver is an API introduced in Bower version 1.5. It fetches artifacts and version information through the 
 repository manager, then automatically searches and saves the artifacts in the repository.
 
-[[bower-resolver-config]]
-=== Configuring the Bower Resolver
-
-You need a +.bowerrc+ JSON file to use the Bower resolver. You can create a new or modify an existing global 
-file. Include the registry URL in the file.
-
-Create a +.bowerrc+ file that points to the URL of a Bower group that will merge your proxy and hosted 
-repositories. The reference to the resolver will also be added, after installed. See the example, below.
-----
-{
-    "registry"    : "http://localhost:8081/repository/bower-group",
-	"resolvers"   : ["bower-nexus3-resolver"]
-}
----- 
-
-You may also install the resolver on a per-project basis by adding it as a +devDependency+ in your 
-+package.json+:
-----
-"devDependencies" : {
-  "bower-nexus3-resolver" : "*"
-}
-----
-
 [[bower-installation]]
 === Installing the Bower Resolver
 
@@ -83,7 +60,7 @@ A hosted Bower repository can be used to map to private Bower registries. You ca
 install components (such as libraries or frameworks) against a hosted repository in Nexus.
 
 To create the Bower package run the command +bower register <packagename> <giturl>+ against a hosted repository. 
-Nexus will create an asset that point to the Git URL. A user will be able to browse Bower packages assets that are registered to a hosted repository.
+Nexus will create an asset that point to the Git URL. A user will be able to browse Bower assets that are registered to a hosted repository.
 
 To add another hosted Bower repository, start a new repository with the recipe 'bower (hosted)' as 
 documented in <<admin-repositories>>.
@@ -115,17 +92,41 @@ components of the same name, the first will be grabbed, and so on.
 
 
 [[bower-browse-search]]
-=== Browsing and Searching Bower Repositories
+=== Browsing Bower Repositories and Searching Packages
 
 You can browse Bower repositories in the user interface inspecting the components and assets and their details as 
 documented in <<browse-browse>>. Further searching can be done in the user interface as described in 
 <<search-components>>. 
 
-////
-A search finds all 
-components and assets that are currently stored in the repository manager, either because they have been deployed 
-to a hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
-////
+Searching for Bower installed packages can be performed in the user interface as described in 
+<<search-components>>.This search will find all Bower modules that are currently stored in the repository 
+manager, either because they have been pushed to a hosted repository or they have been proxied from an upstream 
+repository and cached in the repository manager.
+
+
+[[bower-config]]
+=== Configuring Bower
+
+Installing Bower is done via npm. So, if you install npm first, installing Bower will work. But to do a complete 
+npm configuration for Nexus, review <<_introduction>> to set up and configure the package manager.
+
+Create a +.bowerrc+ file that points to the URL of a Bower group that will merge your proxy and hosted 
+repositories. You can create a new or modify an existing global file. The reference to the resolver will also be 
+added, after installed. See the example, below.
+----
+{
+    "registry"    : "http://localhost:8081/repository/bower-group",
+	"resolvers"   : ["bower-nexus3-resolver"]
+}
+---- 
+
+You may also install the resolver on a per-project basis by adding it as a +devDependency+ in your 
++package.json+:
+----
+"devDependencies" : {
+  "bower-nexus3-resolver" : "*"
+}
+----
 
 ////
 /* Local Variables: */
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -8,21 +8,20 @@
 http://bower.io[Bower] is a package manager for front-end web development. Consumers of JavaScript packages who
 routinely build projects need to keep track of and access remote packages. {pro} and {oss} support the Bower 
 registry format for proxy repositories. This allows the repository manager to take advantage of the packages in 
-the Bower registry and other public registries without incurring repeated downloads of packages, since they will 
-be proxied in the repository manager. The proxy allows {pro} and {oss} to grab artifacts from Bower.
+the Bower registry and other public registries without incurring repeated downloads of packages. The proxy allows {pro} and {oss} to grab artifacts from Bower.
 
-The Bower proxy is optimized for the front-end. It downloads a component (e.g. jQuery) just once when multiple 
-packages depend on the proxy, reducing page load. To use the Bower proxy to access registries through the 
-repository manager, set up a custom URL resolver that plugs into the package manager and interacts with the 
-repository manager. The resolver is an API introduced in Bower version 1.5. It fetches artifacts and version 
-information through the repository manager, then automatically searches and saves the artifacts in the repository.
+Bower is optimized for the front-end. It downloads a component (e.g. jQuery) just once even when a 
+single package depends on it, reducing page load. It uses a proxy to access remote registries with the help of a 
+custom URL resolver, which is dedicated to allowing integration of Bower with {pro} and {oss}. The resolver 
+is an API introduced in Bower version 1.5. Bower fetches artifacts and version information through the repository 
+manager, then automatically searches and saves the artifacts in the repository.
 
 
 [[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up Bower proxy repository within {pro} and {oss} to access a remote repository location. For example, 
-you can configure your Bower proxy to access the public Bower registry at 
+You can set up a Bower proxy repository within {pro} and {oss} to access a remote repository location. For 
+example, you can configure your Bower proxy to access the public Bower registry at 
 http://bower.herokuapp.com/[http://bower.herokuapp.com] or your own private registry.
 
 Enabling rewrite allows the Bower resolver to force itself through the repository manager to retrieve an 
@@ -45,7 +44,7 @@ Minimal configuration steps are:
 
 Creating a Bower hosted repository allows you to register packages in the repository manager. The hosted 
 repository acts as an authoritative location for these components. This effectively creates an asset that becomes 
-a pointer to an external URL (such as a git repository).
+a pointer to an external URL (such as a Git repository).
 
 To add a hosted Bower repository, create a new repository with the recipe 'bower (hosted)' as 
 documented in <<admin-repositories>>.
@@ -76,45 +75,59 @@ each subsequent group is accessed, accordingly. If there are two components of t
 grabbed.
 
 
-[[bower-config]]
-=== Configuring Bower
+[[bower-installation]]
+=== Installing Bower
 
 Installing Bower is done via npm. So, if you install npm first, you can run Bower. But to do a complete 
 npm configuration for {oss} and {pro}, review <<_introduction>> to set up and configure the package manager. 
-Also, be sure your Bower package to version 1.5 or higher.
+Also, be sure to update your Bower package to version 1.5 or higher.
 
 In order to proxy downloads through the repository manager, install the Bower resolver with this command:
 ----
 npm install -g bower-nexus3-resolver
 ----
 
-The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to 
-facilitate using packages in your tools and workflows.
+[[bower-configuration]]
+=== Configuring Bower
 
-You need a +.bowerrc+ JSON file to access the registry URLs. You can create a global file or modify the existing 
-one. Include the registry URL in the file. If, for example, you ever deploy a package to a single Bower 
-hosted repo, set up your +.bowerrc+. In this example you replace your group and hosted repository names with the 
-ones below. 
+Once you have set up your hosted and proxy repositories for Bower packages, and created a repository group to 
+merge them you can create a +.bowerrc+ JSON file in the command line to access registry URLs. Adding the 
+"resolvers" variable to the file enables Bower to fetch artifacts for the repository manager.
+
+Edit your +.bowerrc+ file for a read-only group search and resolver to retrieve artifacts.
+
+----
+{
+	"registry" : "http://localhost:8081/repository/bower-group",
+	"resolvers" : ["bower­-nexus3-­resolver"]
+}
+----
+
+To maintain a private registry, edit the registy variable.
+
+----
+{
+	"registry" : "http://localhost:8081/repository/private-repo",
+	"resolvers" : ["bower­-nexus3-­resolver"]
+}
+----
+
+To register a package with Bower edit the +.bowerrc+ with a register variable. It acts as a write operation since
+it has to go out to the repository manager and associate the package with a particular URL. 
+
+In this example edit your file with a registry object to search a repository group and host a registered 
+package. 
 ----
 {
     "registry" : {
         "search" : [ 
             "http://localhost:8081/repository/bower-group"
         ],
-        "register" : "http://admin:admin123@localhost:8081/repository/bower-hosted"
+        "register" : "http://localhost:8081/repository/bower-hosted"
    },
    "resolvers" : [ "bower-nexus3-resolver" ]
 }
----- 
-
-Duplicating the resolver value is not necessary. Once configured it becomes accessible in the +.bowerrc+ located 
-in your global file.
-
-You have the option to configure multiple search places, but you will only be able to use one register URL. In 
-other words you can only use one hosted repository at a time. 
-
-For example, you might have two hosted repositories (e.g. Hosted1 and Hosted2) and want to push to Hosted1, then 
-later to Hosted2. You would need to manually adjust the register line in the +bowerrc+ example, above.
+----
 
 If you desire to use a per-project +.bowerrc+, then in addition to the single hosted repository +.bowerrc+ you 
 must have a project-specific +.bowerrc+ that contains the following:
@@ -125,6 +138,8 @@ must have a project-specific +.bowerrc+ that contains the following:
    }
 ----
 
+TIP: There is no default global +.bowerrc+. You can configure Bower to register in a unique directory with your 
+own +.bowerrc+ and at the same time maintain a global one defined differently.
 
 [[bower-browse-search]]
 === Browsing Bower Repositories and Searching Packages
@@ -132,9 +147,9 @@ must have a project-specific +.bowerrc+ that contains the following:
 You can browse Bower repositories in the user interface inspecting the components and assets and their details, 
 as described in <<search-components>>.
 
-Searching for Bower installed packages can be performed in the user interface, too. This search will find all 
-Bower modules that are currently stored in the repository manager, either because they have been pushed to a 
-hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
+Searching for Bower packages can be performed in the user interface, too. It finds all packages that are 
+currently stored in the repository manager, either because they have been pushed to a hosted repository or they 
+have been proxied from an upstream repository and cached in the repository manager.
 
 ////
 /* Local Variables: */
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -106,7 +106,7 @@ grabbed, and so on.
 === Browsing Bower Repositories and Searching Packages
 
 You can browse Bower repositories in the user interface inspecting the components and assets and their details as 
-documented in <<browse-assets>>.
+documented in Browse Assets.
 
 Searching for Bower installed packages can be performed in the user interface as described in 
 <<search-components>>. This search will find all Bower modules that are currently stored in the repository 
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -38,28 +38,43 @@ Minimal configuration steps are:
 - Define URL for 'Remote storage' e.g. http://bower.herokuapp.com/[http://bower.herokuapp.com]
 - Select a 'Blob store' for 'Storage'
 
-////
+
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+A hosted Bower repository can be used to deploy your own as well as third-party components. You can register a 
+Bower package to install components (such as libraries or frameworks) against a hosted repository in Nexus. This 
+needs to be flexible to allow users to register against git repos
 
-To create another hosted Bower repository, add a new repository with the recipe 'bower (hosted)' as 
+To create the Bower package run the command +bower register <packagename> <giturl>+ against a hosted repository. 
+Nexus will create an asset that point to the Git URL. A user will be able to browse Bower packages assets that are registered to a hosted repository.
+
+To add another hosted Bower repository, start a new repository with the recipe 'bower (hosted)' as 
 documented in <<admin-repositories>>.
 
 Minimal configuration steps are:
 
 - Define 'Name'
 - Select 'Blob store' for 'Storage'
-////
 
-////
+
 [[bower-group]]
 === Grouping Bower Repositories
 
-tbd
+A repository group is the recommended way to expose all your Bower repositories from the repository manager to 
+your users, without needing any further client side configuration. A repository group allows you to expose the 
+aggregated content of multiple proxy and hosted repositories as well as other repository groups with one URL for 
+tool configuration. This is possible for Bower repositories by creating a new repository with the 'bower (group)' 
+recipe recipe as documented in <<admin-repositories>>.
 
-////
+Minimal configuration steps are:
+
+- Define 'Name'
+- Select 'Blob store' for 'Storage'
+- Add Bower repositories to the 'Members' list in the desired order
+
+Bower groups can contain hosted or proxy bower repositories. Groups operate by order of selection. In other words, the first asset in a given group is accessed; and each subsequent group is accessed, accordingly. If there are two components of the same name, the first asset will be grabbed, and so on.
+
 
 [[bower-installation]]
 === Installing the Bower Resolver
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -45,7 +45,7 @@ Minimal configuration steps are:
 
 Creating a Bower hosted repository allows you to register packages in the repository manager. The hosted 
 repository acts as an authoritative location for these components. This effectively creates an asset that becomes 
-git a pointer to an external URL (such as a git repository).
+a pointer to an external URL (such as a git repository).
 
 To add a hosted Bower repository, create a new repository with the recipe 'bower (hosted)' as 
 documented in <<admin-repositories>>.
--- a/chapter-bower.asciidoc
+++ b/chapter-bower.asciidoc
@@ -17,19 +17,6 @@ set up a custom URL resolver that plugs into the package manager and interacts w
 resolver is an API introduced in Bower version 1.5. It fetches artifacts and version information through the 
 repository manager, then automatically searches and saves the artifacts in the repository.
 
-[[bower-installation]]
-=== Installing the Bower Resolver
-
-The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to 
-facilitate using packages in your tools and workflows.
-
-Before installing resolver, make sure you have npm installed. Also, be sure to upgrade your Bower 
-package to version 1.5 or higher.
-
-You can install the Bower resolver package using npm with this command:
-----
-npm install -g bower-nexus3-resolver
-----
 
 [[bower-proxy]]
 === Proxying Bower Repositories
@@ -56,25 +43,12 @@ Minimal configuration steps are:
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-When you create a Bower hosted repository in Nexus it allows you to install components (such as libraries and 
-frameworks) from private registries, and public registries that have authentication.
-
-////
-lets you register your own bower package
-////
- 
-Floating - Nexus will create an asset that points to the Git URL. A user will be able to browse Bower assets that are 
-registered to a hosted repository.
+Creating a Bower hosted repository allows you to register packages and store components in the repository 
+manager. The hosted repository acts as an authoritative location for these components. This effectively creates 
+an asset that becomes a pointer to an external URL (such as a git repository).
 
 To add another hosted Bower repository, start a new repository with the recipe 'bower (hosted)' as 
 documented in <<admin-repositories>>.
-
-////
-note bower hosted is its own registry
-
-from joe: I recognize the Hosted section esp with my most recent comments might read a little light. I think it IS a little light because I don't think this is the major focus of Bower implementation. For more flesh, you might consider relaying why someone would want to register to a hosted Bower repo. That should add a sentence or two. But I also think light is good if it's a straightforward area where there's not a lot to be done.
-
-////
 
 Minimal configuration steps are:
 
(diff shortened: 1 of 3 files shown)


mosabua:
The client configuration section should be updated to point to the group. Will review more on Friday.

chapter-bower.asciidoc
```diff
@@ -38,28 +38,43 @@ Minimal configuration steps are:
...
 
-////
+
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+A hosted Bower repository can be used to deploy your own as well as third-party components. You can register a 
+Bower package to install components (such as libraries or frameworks) against a hosted repository in Nexus. This 
+needs to be flexible to allow users to register against git repos
```
fjmilens3: Needs a closing `.`.

chapter-bower.asciidoc
```diff
@@ -38,28 +38,43 @@ Minimal configuration steps are:
...
+
 [[bower-group]]
 === Grouping Bower Repositories
 
-tbd
+A repository group is the recommended way to expose all your Bower repositories from the repository manager to 
+your users, without needing any further client side configuration. A repository group allows you to expose the 
+aggregated content of multiple proxy and hosted repositories as well as other repository groups with one URL for 
+tool configuration. This is possible for Bower repositories by creating a new repository with the 'bower (group)' 
+recipe recipe as documented in <<admin-repositories>>.
```
fjmilens3: "recipe recipe" might be phrased better?

mosabua:
Will review when its ready

chapter-bower.asciidoc
```diff
@@ -38,28 +38,43 @@ Minimal configuration steps are:
 - Define URL for 'Remote storage' e.g. http://bower.herokuapp.com/[http://bower.herokuapp.com]
 - Select a 'Blob store' for 'Storage'
 
-////
+
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+A hosted Bower repository can be used to deploy your own as well as third-party components. You can register a
```
joedragons: From my work with it and reading, the hosted Bower repo doesn't actually hold components, it holds mapping files which point to private or public registries.  You'll note that if you go to components after registering, the hosted repository is actually empty.  You could change this to assets and it would technically be true, tho still a little deceptive IMO.
I would recommend you change to something like "can be used to map to private Bower registries" (you can also add why here if you like, I kept it simple).

chapter-bower.asciidoc
joedragons: Sentence also reads a bit strangely to me but I'm not sure how to improve it.  Needs to be flexible...as opposed to what?  Maven does not (fake example)?  Nothing in the prior sentences seem to lead into why we'd say it was flexible here even if true.  Maybe I'm missing the intent?

chapter-bower.asciidoc
```diff
@@ -38,28 +38,43 @@ Minimal configuration steps are:
...
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+A hosted Bower repository can be used to deploy your own as well as third-party components. You can register a 
+Bower package to install components (such as libraries or frameworks) against a hosted repository in Nexus. This 
+needs to be flexible to allow users to register against git repos
 
-To create another hosted Bower repository, add a new repository with the recipe 'bower (hosted)' as 
+To create the Bower package run the command +bower register <packagename> <giturl>+ against a hosted repository.
```
joedragons: As I believe @simpligility noted, you either need config before this or if not I think should reference the config section such as "Once configured in 11.6, ...".  Or move config section before this (which I believe we discussed when reviewing proxy and it didn't move=)).

chapter-bower.asciidoc
```diff
@@ -38,28 +38,43 @@ Minimal configuration steps are:
...
-////
 
-////
+
 [[bower-group]]
 === Grouping Bower Repositories
 
-tbd
+A repository group is the recommended way to expose all your Bower repositories from the repository manager to 
+your users, without needing any further client side configuration. A repository group allows you to expose the
```
joedragons: Since you need to change the .bowerrc no matter what, you might change "without needing any further client side configuration." to "with minimal additional client side configuration." or some such.  I understand the sentiment here and if you disagree don't feel strongly.

chapter-bower.asciidoc
```diff
@@ -38,28 +38,43 @@ Minimal configuration steps are:
...
 
-////
+
 [[bower-group]]
 === Grouping Bower Repositories
 
-tbd
+A repository group is the recommended way to expose all your Bower repositories from the repository manager to 
+your users, without needing any further client side configuration. A repository group allows you to expose the 
+aggregated content of multiple proxy and hosted repositories as well as other repository groups with one URL for
```
joedragons: "in" instead of "for"?

chapter-bower.asciidoc
joedragons: I think this is just a word duplication?  Reads good to me with one 'recipe' struck.

chapter-bower.asciidoc
```diff
@@ -38,28 +38,43 @@ Minimal configuration steps are:
...
+tool configuration. This is possible for Bower repositories by creating a new repository with the 'bower (group)' 
+recipe recipe as documented in <<admin-repositories>>.
+
+Minimal configuration steps are:
+
+- Define 'Name'
+- Select 'Blob store' for 'Storage'
+- Add Bower repositories to the 'Members' list in the desired order
+
+Bower groups can contain hosted or proxy bower repositories. Groups operate by order of selection. In other words, the first asset in a given group is accessed; and each subsequent group is accessed, accordingly. If there are two components of the same name, the first asset will be grabbed, and so on.
```
joedragons: Minor: This line is exceedingly long and would shock me if it fits the 114 (125?) character per line styleguide I think we've been trying to impose.  Unless that's changed and I missed the memo.

chapter-bower.asciidoc
dulanism: Removed the sentence as the earlier sentence makes the point @joedragons

chapter-bower.asciidoc
dulanism: Two 'recipes' don't make a right. One down.

chapter-bower.asciidoc
dulanism: Yes, you're right about the order. The deets are in an earlier comment to keep is further down but I don't recall. But if we're talking chicken-or-egg, it'll be moved up by the time you read this.

chapter-bower.asciidoc
dulanism: A minor fix; i see your point

chapter-bower.asciidoc
dulanism: in

chapter-bower.asciidoc
dulanism: Style guide lines still apply. I fixed the wrapping.
Also, made edits to use appropriate word, ie member

chapter-bower.asciidoc
dulanism: Hope my edits, per your recommendation make more sense

dulanism:
@mosabua I moved the client configuration bit to the top. There's a was a lot of tweaking in the docs to get an earlier version of this resolver to a customer. IMO, this current state seems like the right order (i.e. config first, installation, etc). Please review. The URL in the example now points to the group, with supporting text. Is there anything else missing from client config section @joedragons @fjmilens3? (What's the deal with the .tgz for the resolver, btw?)

mosabua:
@dulanism .... it does not make sense to have the client configuration first since you wont know what you have configured on the server by then and you wont be able to test it. All other chapters document the config on the Nexus Repository Manager first and then the client tools. This is how you would do it in practice and I suggest you change it in this chapter. Otherwise we should change the order in all other chapters as well imho.

mosabua:
Also the chapter still should indicate that installing bower is done via npm so it makes sense to do a complete npm configuration first... and link to the npm chapter for details.

chapter-bower.asciidoc
```diff
@@ -38,56 +52,73 @@ Minimal configuration steps are:
 - Define URL for 'Remote storage' e.g. http://bower.herokuapp.com/[http://bower.herokuapp.com]
 - Select a 'Blob store' for 'Storage'
 
-////
+
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+A hosted Bower repository can be used to map to private Bower registries. You can register a Bower package to
```
fjmilens3: I'm not sure what this first sentence means. To the extent we have something that could map to another Bower *registry*, that's proxy, not hosted.

chapter-bower.asciidoc
```diff
@@ -38,56 +52,73 @@ Minimal configuration steps are:
...
 - Select a 'Blob store' for 'Storage'
 
-////
+
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+A hosted Bower repository can be used to map to private Bower registries. You can register a Bower package to 
+install components (such as libraries or frameworks) against a hosted repository in Nexus.
```
fjmilens3: I think I know what you mean, but I think we need a cleaner way of saying it (for the second sentence as well).

chapter-bower.asciidoc
```diff
@@ -38,56 +52,73 @@ Minimal configuration steps are:
...
+npm configuration for Nexus, review <<_introduction>> to set up and configure the package manager.
+
+Create a +.bowerrc+ file that points to the URL of a Bower group that will merge your proxy and hosted 
+repositories. You can create a new or modify an existing global file. The reference to the resolver will also be 
+added, after installed. See the example, below.
 ----
 {
-    "registry"    : "http://localhost:8081/repository/your-bower-repo",
-    "resolvers"   : ["bower-nexus3-resolver"]
+    "registry"    : "http://localhost:8081/repository/bower-group",
```
fjmilens3: This configuration is no longer strictly accurate now that we're talking about group and hosted repos. Yes, you'll want the `search` registry to point to the group repo, but you'll also need a `register` entry: 

https://github.com/bower/spec/blob/master/config.md#registry

@joedragons might be able to help you figure out how to write this up as he's had to do this himself for testing purposes, I believe.

chapter-bower.asciidoc
```diff
@@ -38,56 +52,73 @@ Minimal configuration steps are:
...
 
-For example, here's a +.bowerrc+ file that references the URL for the Bower registry in the repository 
-manager:
+[[bower-config]]
+=== Configuring Bower
+
+Installing Bower is done via npm. So, if you install npm first, installing Bower will work. But to do a complete 
+npm configuration for Nexus, review <<_introduction>> to set up and configure the package manager.
+
+Create a +.bowerrc+ file that points to the URL of a Bower group that will merge your proxy and hosted
```
fjmilens3: Not *quite* true, as you'll also need another entry if you're going to want to register stuff to hosted, I think (I don't think you can register through the group).

joedragons:
@dulanism @mosabua It makes 0 sense to me to specify commands that won't work before you've done config such as line 62 "+To create the Bower package run the command +bower register <packagename> <giturl>+ against a hosted repository."  I scanned Maven and npm and neither of them do this that I saw.  NPM has commands to call within it's configuration chapter: http://books.sonatype.com/nexus-book/3.0/reference/npm.html#npm-configuring.  That is why I mentioned moving config up.  Because if you do things in this order they will fail.  If you move that around instead, that's also fine with me.

While you can use npm through Nexus to setup bower, it's not required to my knowledge.  If you add a link I recommend caution in wording.  Our NPM section has minimal information about npm and most information on configuring through Nexus which again I don't believe is required for this effort.  Aka if people want to use Bower using Nexus and NPM not they can.

chapter-bower.asciidoc
```diff
@@ -17,6 +17,20 @@ set up a custom URL resolver that plugs into the package manager and interacts w
 resolver is an API introduced in Bower version 1.5. It fetches artifacts and version information through the 
 repository manager, then automatically searches and saves the artifacts in the repository.
 
+[[bower-installation]]
+=== Installing the Bower Resolver
+
+The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to
```
joedragons: Minor: Word "these" is semi-weird here potentially to anyone starting reading here (rather than from the beginning like I just did).  Consider striking, as I think it reads fine without it.  Otherwise consider more detail.

chapter-bower.asciidoc
joedragons: Should probably say "Bower packages" not "Bower registries".  They are individual, tho I guess if you wanted to do a whole set you could.  I'm not sure if that solves @fjmilens3 's concern totally but I think it helps.

chapter-bower.asciidoc
```diff
@@ -38,56 +52,73 @@ Minimal configuration steps are:
...
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+A hosted Bower repository can be used to map to private Bower registries. You can register a Bower package to 
+install components (such as libraries or frameworks) against a hosted repository in Nexus.
 
-To create another hosted Bower repository, add a new repository with the recipe 'bower (hosted)' as 
+To create the Bower package run the command +bower register <packagename> <giturl>+ against a hosted repository. 
+Nexus will create an asset that point to the Git URL. A user will be able to browse Bower assets that are registered to a hosted repository.
```
joedragons: While they can browse, 11.6 covers browsing overall and this is also true of proxy and group (can browse) so I may be missing the distinction here.  Or what you're trying to say.

chapter-bower.asciidoc
```diff
@@ -38,56 +52,73 @@ Minimal configuration steps are:
...
 
-[[bower-installation]]
-=== Installing the Bower Resolver
+- Define 'Name'
+- Select 'Blob store' for 'Storage'
+- Add Bower repositories to the 'Members' list in the desired order
 
-The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
-facilitate using packages in your tools and workflows.
+Bower groups can contain hosted or proxy repositories. Groups operate by order of selection. In other words, the
```
joedragons: Minor: There is repitition here as you say above groups aggregate proxy and hosted repos.  No concerns if intentional, but pointing out in case not.

chapter-bower.asciidoc
```diff
@@ -38,56 +52,73 @@ Minimal configuration steps are:
...
+=== Browsing Bower Repositories and Searching Packages
+
+You can browse Bower repositories in the user interface inspecting the components and assets and their details as 
+documented in <<browse-browse>>. Further searching can be done in the user interface as described in 
+<<search-components>>. 
 
-[[bower-resolver-config]]
-=== Configuring the Bower Resolver
+Searching for Bower installed packages can be performed in the user interface as described in 
+<<search-components>>.This search will find all Bower modules that are currently stored in the repository
```
joedragons: Search seems to be mentioned twice with a paragraph break in between.  I'm wondering if the second portion of the section paragraph was supposed to be a link to browse assets.
If this stays, a space appears to be missing after <<search-components>>.  AKA "<<search-components>>.This" should be "<<search-components>>. This".

chapter-bower.asciidoc
joedragons: @dulanism Here's the configuration I ended up with for my tests:
"registry": {
       "search": [
          "http://localhost:8081/repository/bowerg"
       ],
       "register": "http://admin:admin123@localhost:8081/repository/bowerh"
    }

It may be noteworthy here (or above, in hosted), that you can have multiple search places but only one register.  So basically you can only have one hosted repo at once.  If you wanted to push to Hosted1 then later to Hosted2, you'd need to manually adjust the registry line here.  OR do what you follow up with below and have the individual .bowerrc point to the correct hosted (makes a nice segue).

joedragons:
I recognize the Hosted section esp with my most recent comments might read a little light. I think it IS a little light because I don't think this is the major focus of Bower implementation. For more flesh, you might consider relaying why someone would want to register to a hosted Bower repo.  That should add a sentence or two.  But I also think light is good if it's a straightforward area where there's not a lot to be done.

chapter-bower.asciidoc
dulanism: Revising/consolidating both sentences "When you create a Bower hosted repository in Nexus it allows you to install components (such as libraries and frameworks) from private and public registries."

chapter-bower.asciidoc
dulanism: note: come back to this

chapter-bower.asciidoc
dulanism: That was left over from some previous edits. "These" gone.

chapter-bower.asciidoc
dulanism: Redundancy. Leveled it out.

chapter-bower.asciidoc
joedragons: BTW, in case not clear, that snip doesn't include the resolver line=)

chapter-bower.asciidoc
dulanism: Fixed... i think new commit makes more sense

chapter-bower.asciidoc
fjmilens3: If you're going to use a per-project `.bowerrc` as @mosabua has suggested, then *in addition to* the normal `.bowerrc` you have above, you'll also want to have a per-project `.bowerrc` that only contains:

   "registry": {
       "register": "http://admin:admin123@localhost:8081/repository/bowerh"
    }

If they're using a special case where they only have the one hosted repo, then they can use the example setup @joedragons has since they won't be editing it all the time. (In fairness, people may just end up preferring to edit their .bowerrc anyway, I don't know how people would use this with Nexus in the wild to be honest.)

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
 - Define URL for 'Remote storage' e.g. http://bower.herokuapp.com/[http://bower.herokuapp.com]
 - Select a 'Blob store' for 'Storage'
 
-////
+
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+Creating a Bower hosted repository allows you to register packages and store components in the repository
```
fjmilens3: Right now we're not storing components, not in the sense that Bower packages are actually being stored there. I think that's the long-term intention but we're not there yet. Do we want this like this, @DarthHater?

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
-[[bower-resolver-config]]
-=== Configuring the Bower Resolver
+The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
+
+You need a +.bowerrc+ JSON file to access the registry URLs. You can create a global file or modify the existing 
+one. Include the registry URL in the file. If, for example, you ever deploy a package to a single Bower 
+hosted repo, set up your +.bowerrc+ like the example, below. 
+----
+"registry": {
```
fjmilens3: This indenting is strange, probably because of Github.

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
+       "search": ["http://localhost:8081/repository/bower-group"],
+       "register": ["http://admin:admin123@localhost:8081/repository/bower-hosted"]
+	   "resolvers": ["bower-nexus3-resolver"]
```
fjmilens3: This isn't correct.  You want something like:

    {
        "registry" : {
            "search" : [ 
                "http://localhost:8081/repository/bower-group"
            ],
            "register" : "http://admin:admin123@localhost:8081/repository/bower-hosted"
       },
       "resolvers" : [ "bower-nexus3-resolver" ]
    }

@joedragons, can you read over this as well to check for typos in what I wrote?

dulanism: Thanks @fjmilens3 ... wasn't sure the structure

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
+Once configured it becomes executable in the +.bowerrc+ located in your home directory.
+
+You have the option to configure multiple search places, but you will only be able to use one register URL. In 
+other words you can only use one hosted repository at a time. 
 
-You need a +.bowerrc+ JSON file to use the Bower resolver. You can create a global file or modify your existing global file. Include the registry URL in the file.
+For example, you might have two hosted repositories (e.g. Hosted1 and Hosted2) and want to push to Hosted1, then 
+later to Hosted2. You would need to manually adjust the register line in the +bowerrc+ example, above.
+
+If you wish to maintain a private registry, change the value in the file as shown:
```
fjmilens3: Hosted is always in some sense a private registry, so I'm not sure this makes a lot of sense. The difference is that if you have more than one that you're using, you may want to go the per-project .bowerrc instead of editing your main one.

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
 - Select a 'Blob store' for 'Storage'
 
-////
+
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+Creating a Bower hosted repository allows you to register packages and store components in the repository 
+manager. The hosted repository acts as an authoritative location for these components. This effectively creates
```
joedragons: I think I may ahve mentioned this before but I'd say "assets" not "components" because nothing shows up in Browse Components for Hosted.  We're basically just storing a mapping (json).  If you knew this and intentionally made this choice, then sorry=)

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
 [[bower-hosted]]
 === Hosting Bower Repositories
 
-A hosted Bower repository can be used to deploy your own as well as third-party components.
+Creating a Bower hosted repository allows you to register packages and store components in the repository 
+manager. The hosted repository acts as an authoritative location for these components. This effectively creates 
+an asset that becomes a pointer to an external URL (such as a git repository).
 
-To create another hosted Bower repository, add a new repository with the recipe 'bower (hosted)' as 
+To add another hosted Bower repository, start a new repository with the recipe 'bower (hosted)' as
```
joedragons: Minor: None are created by default so I think the word "another" may be unneeded.  (I am guessing this may have been copied from Maven or Nuget where they are created by default and this makes sense.)

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
+
 [[bower-group]]
 === Grouping Bower Repositories
 
-tbd
+A repository group is the recommended way to expose all your Bower repositories from the repository manager to 
+your users, with minimal additional client side configuration. A repository group allows you to expose the 
+aggregated content of multiple proxy and hosted repositories as well as other repository groups with one URL in 
+tool configuration. This is possible for Bower repositories by creating a new repository with the 'bower (group)' 
+recipe as documented in <<admin-repositories>>.
```
joedragons: Minor: It is weird to me you link how to document groups but did not for proxy and hosted though those occur first.  It seems likely to me anyone who gets here would already have suffered through how to create (either finding out how or finding link on their own) and would not need the link now, as creating a group without first creating hosted/proxies doesn't make any sense (and actually is impossible).

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
 npm install -g bower-nexus3-resolver
 ----
 
-[[bower-resolver-config]]
-=== Configuring the Bower Resolver
+The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
+
+You need a +.bowerrc+ JSON file to access the registry URLs. You can create a global file or modify the existing 
+one. Include the registry URL in the file. If, for example, you ever deploy a package to a single Bower
```
fjmilens3: I thought we wanted to cover the basic read-only setup first, then go into configuring it for writing to Nexus, maybe even in a separate section?

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
+
+- Define 'Name'
+- Select 'Blob store' for 'Storage'
+- Add Bower repositories to the 'Members' list in the desired order
 
-[[bower-installation]]
-=== Installing the Bower Resolver
+Bower groups operate by order of selection. In other words, the first member in a given group is accessed; and 
+each subsequent group is accessed, accordingly. If there are two components of the same name, the first will be 
+grabbed, and so on.
```
joedragons: Minor: ", and so on" still reads weird to me=)

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
+
+Searching for Bower installed packages can be performed in the user interface, too. This search will find all 
+Bower modules that are currently stored in the repository manager, either because they have been pushed to a 
+hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
+
+
+[[bower-config]]
+=== Configuring Bower
+
+Installing Bower is done via npm. So, if you install npm first, your can run Bower. But to do a complete
```
joedragons: "your" typo; should be "you" I believe.

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
+each subsequent group is accessed, accordingly. If there are two components of the same name, the first will be 
+grabbed, and so on.
 
-The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
-facilitate using packages in your tools and workflows.
 
-Before installing resolver, make sure you have npm installed. Also, be sure to upgrade your Bower 
-package to version 1.5 or higher before you proceed to the resolver installation.
+[[bower-browse-search]]
+=== Browsing Bower Repositories and Searching Packages
```
joedragons: Minor: I know this is a sore spot (sorry to keep harping) but you can't search until you push/pull which you can't do til you configure, so I'd move this below configure at least.

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
 ----
 
-[[bower-resolver-config]]
-=== Configuring the Bower Resolver
+The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
+
+You need a +.bowerrc+ JSON file to access the registry URLs. You can create a global file or modify the existing 
+one. Include the registry URL in the file. If, for example, you ever deploy a package to a single Bower 
+hosted repo, set up your +.bowerrc+ like the example, below.
```
joedragons: Minor: Suggest "like the example below:" or "like this example:".  You might also say/add "replacing your group and hosted repository names with the ones below".

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
-    "resolvers"   : ["bower-nexus3-resolver"]
-}
+"registry": "https://bower.herokuapp.com"
+----
+
+If you desire to use a per-project +.bowerrc+, then in addition to the single hosted repository +.bowerrc+ you 
+must have a project-specific +.bowerrc+ that only contains the following:
+
+----
+"registry": {
```
fjmilens3: There's that indentation thing again.

chapter-bower.asciidoc
dulanism: Weird. I edited this section. Don't know why you're seeing this @fjmilens3

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
 ----
-{
-    "registry"    : "http://localhost:8081/repository/your-bower-repo",
-    "resolvers"   : ["bower-nexus3-resolver"]
-}
+"registry": "https://bower.herokuapp.com"
+----
+
+If you desire to use a per-project +.bowerrc+, then in addition to the single hosted repository +.bowerrc+ you 
+must have a project-specific +.bowerrc+ that only contains the following:
```
fjmilens3: It doesn't have to *only* contain this, they may have other stuff for their own setup too.  But they do need this, yes.

chapter-bower.asciidoc
fjmilens3: Not a problem. Depending on what convention you follow for representing .bowerrc entries, I don't know if you'll want to keep the initial { and } or remove those from the text.  I'm thinking remove, now that I look ahead.

joedragons: Looks good to me.  Mine has more (unnecessary) whitespace, but otherwise this is what I have=)

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
+The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
+
+You need a +.bowerrc+ JSON file to access the registry URLs. You can create a global file or modify the existing 
+one. Include the registry URL in the file. If, for example, you ever deploy a package to a single Bower 
+hosted repo, set up your +.bowerrc+ like the example, below. 
+----
+"registry": {
+       "search": ["http://localhost:8081/repository/bower-group"],
+       "register": ["http://admin:admin123@localhost:8081/repository/bower-hosted"]
```
joedragons: Minor: PDF has a weird arrow here, I'm guessing to signify a line wrap?  Not sure anything can be done but noting as looks weird to me.
screen shot 2016-02-17 at 10 43 26 am

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
+	   "resolvers": ["bower-nexus3-resolver"]
+    }
+---- 
+
+The URL you configure for the resolver should be in the group. Since you cannot register to a group nor search a 
+hosted repository this configuration is necessary. Also, duplicating the resolver value is not necessary. 
+Once configured it becomes executable in the +.bowerrc+ located in your home directory.
```
joedragons: (comments on this entire paragraph)
1) I'm not sure what the first sentence is saying.
2) Regardless, you **can** search a hosted repository.  However you cannot register to a group (or a proxy).  Because the hosted is contained in the group there's no need to put the hosted repository in the search area.  But you could.  Same with proxy.  Basically, by putting the group there, you're saving yourself from configuring a slew of different repos to search (saves typing time and potential typos).
3) I don't understand the word "executable".  If @fjmilens3 is OK here, I assume this is just me unclear what the resolver is actually doing.  Otherwise recommend perhaps "accessible via the configuration".
4) Minor: I would put "global file" rather than "home directory" because per http://bower.io/docs/config/ (Placement & Order) home directory is just one of the locations for a file not on the individual project level.

chapter-bower.asciidoc
joedragons: Doesn't make sense to me=)  Not sure why we'd tell them how to use herokuapp...that's basically telling them not to use Nexus?

chapter-bower.asciidoc
```diff
@@ -38,57 +39,106 @@ Minimal configuration steps are:
...
-manager:
 ----
-{
-    "registry"    : "http://localhost:8081/repository/your-bower-repo",
-    "resolvers"   : ["bower-nexus3-resolver"]
-}
+"registry": "https://bower.herokuapp.com"
+----
+
+If you desire to use a per-project +.bowerrc+, then in addition to the single hosted repository +.bowerrc+ you
```
joedragons: Assuming the above paragraph stays despite my confusion, I recommend moving it below this one and starting this with the word "Alternatively".  So "Alternatively, if you desire to use..."  This flows super nicely from the section that tells you can only configure one hosted repo at once IMO.

chapter-bower.asciidoc
joedragons: Agree, only can be struck IMO.

chapter-bower.asciidoc
dulanism: Fixed

chapter-bower.asciidoc
dulanism: That's not the intent. Striking it out.

chapter-bower.asciidoc
dulanism: Leftover boilerplate text. Updating per your comments

chapter-bower.asciidoc
DarthHater: Yeah that's phase 2. Might leave it as simple as register bower packages.

chapter-bower.asciidoc
dulanism: Overlooked. Removed.

chapter-bower.asciidoc
dulanism: Okay. I'll agree to agree :)

chapter-bower.asciidoc
dulanism: Fixed.

chapter-bower.asciidoc
dulanism: Point taken. Moved to below config.

chapter-bower.asciidoc
dulanism: Makes sense. And I'll agree to agree. The explanation is good.

chapter-bower.asciidoc
dulanism: One "only" down.

chapter-bower.asciidoc
dulanism: Let's see if the reformatted code example kills the arrow.

chapter-bower.asciidoc
dulanism: 1) Cut out the first couple sentences in the first paragraph. 
2) The group part makes sense. I dunno: brainfart?
3) Uncle. Accessible is better term.
4) Updated to global file; and now we're normalizing Bower to Nexus the config page makes more sense.

chapter-bower.asciidoc
dulanism: Done and done.

chapter-bower.asciidoc
```diff
@@ -12,20 +12,21 @@ the Bower registry and other public registries without incurring repeated downlo
 be proxied in the repository manager. The proxy allows {pro} and {oss} to grab artifacts from Bower.
 
 The Bower proxy is optimized for the front-end. It downloads a component (e.g. jQuery) just once when multiple
```
joedragons: Minor: This is outside the scope of the changed but the proxy will download the component just once even when a single package depends on it.  I think either add the word "even" before when multiple or you can remove "when multiple packages depend on the proxy".  I made minor because I hadn't noticed this before so I feel somewhat it's a minor wording thing.

chapter-bower.asciidoc
```diff
@@ -38,77 +39,102 @@ Minimal configuration steps are:
...
 
-Before installing resolver, make sure you have npm installed. Also, be sure to upgrade your Bower 
-package to version 1.5 or higher before you proceed to the resolver installation.
+[[bower-config]]
+=== Configuring Bower
 
-You can install the downloaded Bower resolver package using npm globally with the command:
+Installing Bower is done via npm. So, if you install npm first, you can run Bower. But to do a complete 
+npm configuration for {oss} and {pro}, review <<_introduction>> to set up and configure the package manager. 
+Also, be sure your Bower package to version 1.5 or higher.
```
fjmilens3: "be sure [to update] your Bower" maybe?

chapter-bower.asciidoc
```diff
@@ -38,77 +39,102 @@ Minimal configuration steps are:
...
+In order to proxy downloads through the repository manager, install the Bower resolver with this command:
 ----
 npm install -g bower-nexus3-resolver
 ----
 
-[[bower-resolver-config]]
-=== Configuring the Bower Resolver
-
-You need a +.bowerrc+ JSON file to use the Bower resolver. You can create a global file or modify your existing global file. Include the registry URL in the file.
+The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to
```
fjmilens3: The Bower resolver doesn't keep track of packages, Bower itself does. The resolver is just part of Bower.

chapter-bower.asciidoc
```diff
@@ -38,77 +39,102 @@ Minimal configuration steps are:
...
+   "resolvers" : [ "bower-nexus3-resolver" ]
 }
-----
+---- 
+
+Duplicating the resolver value is not necessary. Once configured it becomes accessible in the +.bowerrc+ located 
+in your global file.
+
+You have the option to configure multiple search places, but you will only be able to use one register URL. In 
+other words you can only use one hosted repository at a time.
```
fjmilens3: You can only use one hosted repository at a time *if you're registering* to it.  You could use multiple hosted registries to pull stuff down via a group, for example.

chapter-bower.asciidoc
```diff
@@ -12,20 +12,21 @@ the Bower registry and other public registries without incurring repeated downlo
...
+repository manager. The resolver is an API introduced in Bower version 1.5. It fetches artifacts and version 
+information through the repository manager, then automatically searches and saves the artifacts in the repository.
+
 
 [[bower-proxy]]
 === Proxying Bower Repositories
 
-You can set up Bower proxy repository within Nexus to access a remote repository location. For example, you can 
-configure your Bower proxy to access the public Bower registry at 
+You can set up Bower proxy repository within {pro} and {oss} to access a remote repository location. For example,
```
joedragons: Minor: "set up **a** Bower proxy"? (added a)

chapter-bower.asciidoc
```diff
@@ -38,77 +39,102 @@ Minimal configuration steps are:
...
-You need a +.bowerrc+ JSON file to use the Bower resolver. You can create a global file or modify your existing global file. Include the registry URL in the file.
+The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
 
-For example, here's a +.bowerrc+ file that references the URL for the Bower registry in the repository 
-manager:
+You need a +.bowerrc+ JSON file to access the registry URLs. You can create a global file or modify the existing 
+one. Include the registry URL in the file. If, for example, you ever deploy a package to a single Bower 
+hosted repo, set up your +.bowerrc+. In this example you replace your group and hosted repository names with the 
+ones below.
```
fjmilens3: @mosabua, did we decide not to break this out into general read-only setup vs. read-and-write setup?

chapter-bower.asciidoc
```diff
@@ -38,77 +39,102 @@ Minimal configuration steps are:
...
-manager:
+You need a +.bowerrc+ JSON file to access the registry URLs. You can create a global file or modify the existing 
+one. Include the registry URL in the file. If, for example, you ever deploy a package to a single Bower 
+hosted repo, set up your +.bowerrc+. In this example you replace your group and hosted repository names with the 
+ones below. 
 ----
 {
-    "registry"    : "http://localhost:8081/repository/your-bower-repo",
-    "resolvers"   : ["bower-nexus3-resolver"]
+    "registry" : {
```
fjmilens3: Do we have a minimal example of a read-only setup that's only hitting the Bower group repo?

chapter-bower.asciidoc
```diff
@@ -38,77 +39,102 @@ Minimal configuration steps are:
...
-documented in <<browse-browse>>. Further searching can be done in the user interface as described in <<search-components>>. 
+You can browse Bower repositories in the user interface inspecting the components and assets and their details, 
+as described in <<search-components>>.
 
-////
-A search finds all 
-components and assets that are currently stored in the repository manager, either because they have been deployed 
-to a hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
-////
+Searching for Bower installed packages can be performed in the user interface, too. This search will find all
```
fjmilens3: I don't know if I'd say "installed" as such, but that's just me.  Saying "Bower installed package" kind of means a certain thing, and while I think people reading this will understand, I'm not sure if there's a better way to phrase it or not.

chapter-bower.asciidoc
```diff
@@ -38,77 +39,102 @@ Minimal configuration steps are:
...
-
-You need a +.bowerrc+ JSON file to use the Bower resolver. You can create a global file or modify your existing global file. Include the registry URL in the file.
+The Bower resolver keeps track of packages in a manifest file, +bower.json+. Bower provides hooks to 
+facilitate using packages in your tools and workflows.
 
-For example, here's a +.bowerrc+ file that references the URL for the Bower registry in the repository 
-manager:
+You need a +.bowerrc+ JSON file to access the registry URLs. You can create a global file or modify the existing 
+one. Include the registry URL in the file. If, for example, you ever deploy a package to a single Bower 
+hosted repo, set up your +.bowerrc+. In this example you replace your group and hosted repository names with the
```
joedragons: Minor: Suggest wording change from "If, for example, you ever deploy a package to a single Bower hosted repo, set up..." to either:
"Further, if you ever deploy a..."
OR
"If you ever deploy a package to a single Bower hosted repo, further set up..."

You might also be more specific here on how to set up rather than relying on the fact they'd copy paste and/or understand the below example.  If you need more details on what I mean here, feel free to ping me.
I marked minor because I don't believe any of this text is untrue, just could potentially be more clear.

chapter-bower.asciidoc
DarthHater: Probably just Bower packages.

chapter-bower.asciidoc
```diff
@@ -38,77 +39,102 @@ Minimal configuration steps are:
...
+You can browse Bower repositories in the user interface inspecting the components and assets and their details, 
+as described in <<search-components>>.
 
-////
-A search finds all 
-components and assets that are currently stored in the repository manager, either because they have been deployed 
-to a hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
-////
+Searching for Bower installed packages can be performed in the user interface, too. This search will find all 
+Bower modules that are currently stored in the repository manager, either because they have been pushed to a
```
joedragons: Minor: I can only recall seeing Bower stuff being referred to as packages (above you say modules).  These may be interchangable or both right, but I thought I'd note in case.

joedragons:
+1 with a few minor comments.  Thanks!

chapter-bower.asciidoc
joedragons: FYI this is still there.  I highly suspect it's a wrapping issue and not
related to this specifically.

chapter-bower.asciidoc
dulanism: Per a convo with @fjmilens3 I had an earlier version that had this is as the simplest config{
"registry" : "http://localhost:8081/repository/bower-group",
"resolvers" : ["bower­-nexus3-­resolver"]
} 

This could be the read-only set up / and the samples below can be rolled up into a write-only config chapter. Thoughts?

joedragons: This is kinda what I was trying to say https://github.com/sonatype/nexus-book/pull/133/files#r53230008.  This section might be slightly revamped to say "if you want to search" configure X, "if you want to search and host" configure Y, here are examples=)
That explains both what they need to do and gives them examples, can't get better=)

chapter-bower.asciidoc
dulanism: fixed

chapter-bower.asciidoc
dulanism: Fixed

chapter-bower.asciidoc
dulanism: Fixed

chapter-bower.asciidoc
dulanism: Updated

chapter-bower.asciidoc
dulanism: fleshed out the example... rewrote this section

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
 [[bower-introduction]]
 === Introduction
 
-http://bower.io[Bower] is a package manager for front-end web development. Consumers of JavaScript packages who
-routinely build projects need to keep track of and access remote packages. {pro} and {oss} support the Bower 
-registry format for proxy repositories. This allows the repository manager to take advantage of the packages in 
-the Bower registry and other public registries without incurring repeated downloads of packages, since they will 
-be proxied in the repository manager. The proxy allows {pro} and {oss} to grab artifacts from Bower.
-
-The Bower proxy is optimized for the front-end. It downloads a component (e.g. jQuery) just once when multiple
```
mosabua: This was off topic for the repo manager usage.

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
-You can set up Bower proxy repository within Nexus to access a remote repository location. For example, you can 
-configure your Bower proxy to access the public Bower registry at 
-http://bower.herokuapp.com/[http://bower.herokuapp.com] or your own private registry.
-
-Enabling rewrite allows the Bower resolver to force itself through Nexus to retrieve an artifact. This is enabled 
-by default.
+You can set up a Bower proxy repository to access a remote repository location, for example the official Bower
+registry at http://bower.herokuapp.com/[http://bower.herokuapp.com] that is configured as the default on Bower.
 
-If deactivated, no rewrite of the URL occurs. As a result, the original component URL is exposed. Turning off
```
mosabua: Usage of artifact and other problems...

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+You can set up a Bower proxy repository to access a remote repository location, for example the official Bower
+registry at http://bower.herokuapp.com/[http://bower.herokuapp.com] that is configured as the default on Bower.
 
-If deactivated, no rewrite of the URL occurs. As a result, the original component URL is exposed. Turning off 
-rewrite capabilities proxies the information directly from the remote registry without redirecting to the 
-repository manager to retrieve artifacts. The content will be as if navigating directly to the Bower registry in 
-your browser.
- 
-Minimal configuration steps are:
+To proxy a Bower registry, you simply create a new 'bower (proxy)' as documented in <<admin-repositories>> in
```
mosabua: Need to add to generic section about repo creation

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+details. Minimal configuration steps are:
 
 - Define 'Name'
 - Define URL for 'Remote storage' e.g. http://bower.herokuapp.com/[http://bower.herokuapp.com]
 - Select a 'Blob store' for 'Storage'
 
-////
+The 'Bower' specific configuration section include the setting to 'Enable rewrite of package URLs'. This causes
+Bower to retrieve components and their dependencies through the repository manager even if original meta data has
```
mosabua: The specific docs is example after the simple steps in all other chapters. It gets the users faster to success.

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
-- Define 'Name'
+- Define 'Name' - e.g. `bower-internal`
 - Select 'Blob store' for 'Storage'
-////
 
-////
+
 [[bower-group]]
-=== Grouping Bower Repositories
+=== Bower Repository Groups
```
mosabua: We use Repository Group across all the docs..

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+- Select 'Blob store' for 'Storage'
+- Add Bower repositories to the 'Members' list in the desired order
 
 [[bower-installation]]
-=== Installing the Bower Resolver
+=== Installing Bower
 
-The Bower resolver keeps track of these packages in a manifest file, +bower.json+. Bower provides hooks to 
-facilitate using packages in your tools and workflows.
+Bower is typically installed with `npm`. Since the repository manager supports NPM repositories for proxying, we
```
mosabua: The reference to NPM config I requested but was never added

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+Bower. Once this is completed you can install Bower with the usual command.
 
-Before installing resolver, make sure you have npm installed. Also, be sure to upgrade your Bower 
-package to version 1.5 or higher before you proceed to the resolver installation.
+----
+ npm install -g bower
+----
+
+Bower version 1.5 or higher is required and can be verified with
+
```
mosabua: Always good to show users what to do

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+necessary, so that Bower uses the required custom resolver.
 
-For example, here's a +.bowerrc+ file that references the URL for the Bower registry in the repository 
-manager:
+.Global .bowerrc file in your home directory for package download via the group bower-all
 ----
 {
-    "registry"    : "http://localhost:8081/repository/your-bower-repo",
-    "resolvers"   : ["bower-nexus3-resolver"]
+  "registry" : {
```
mosabua: This is more suited for alter changes and addition of register..

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+    "search" : [ "http://localhost:8081/repository/bower-all" ]
+   },
+ "resolvers" : [ "bower-nexus3-resolver" ]
 }
 ----
 
-You may also install the resolver on a per-project basis by adding it as a +devDependency+ in your 
-+package.json+:
+IMPORTANT: The .bowerrc file can be located in various locations. For global configuration for a specific
```
mosabua: Note about location and recommendation to put this in ~

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+they are merged. Details can be found in http://bower.io/docs/config/[the documentation].
+
+With this configuration in place, any further Bower command invocations trigger package downloads via the
+repository manager.
+
+Running an `install` command logs the download via the repository manager:
+
 ----
-"devDependencies" : {
-  "bower-nexus3-resolver" : "*"
```
mosabua: Users want to be able to tell how to verify it works..

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+$ bower install jquery
+bower jquery#*              not-cached nexus+http://localhost:8081/repository/bower-all/jquery#*
+bower jquery#*                 resolve nexus+http://localhost:8081/repository/bower-all/jquery#*
+bower jquery#*                resolved nexus+http://localhost:8081/repository/bower-all/jquery#2.2.0
+bower jquery#^2.2.0            install jquery#2.2.0
+
+jquery#2.2.0 bower_components/jquery
+----
+
+If anonymous access to the repository manager is disabled, you have to specify the credentials for the accessing
```
mosabua: THis stuff was entirely missing

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+
+You can browse Bower repositories in the user interface inspecting the components and assets and their details, as
+described in <<search-components>>.
+
+Searching for Bower packages can be performed in the user interface, too. It finds all packages that are currently
+stored in the repository manager, either because they have been pushed to a hosted repository or they have been
+proxied from an upstream repository and cached in the repository manager.
+
+
+[[bower-register]]
```
mosabua: Separate new section about registering since it is an entirely different use case from simple downloading.. I suggested this multiple times but it was never implemented. All other chapters do the same.

chapter-using-nexus-ui.asciidoc
```diff
@@ -281,6 +281,12 @@ A configurable search using the criteria you select is available via the 'Custom
 section of the 'Browse' main menu. Initially it has no criteria and it allows you to create a search with criteria
 you add with the 'More Criteria' button.
 
+Bower Search::
```
mosabua: This was simply missing.

chapter-bower.asciidoc
```diff
@@ -5,113 +5,209 @@
...
+Alternatively, if you desire to use a per-project +.bowerrc+ file that you potentially version in your source code
+management system with the rest of the package code, you can use a simplified file:
+
+----
+"registry": {
+   "register": "http://localhost:8081/repository/bower-hosted"
+   }
+----
+
+Authentication is managed in the same manner as for proxying with anonymous access disabled as documented in
```
mosabua: I still think we should add a command to register a package and example output. I will add that but currently I could not get this working.

mosabua:
Imho we still need an example for registering and document the strictssl stuff as well.

mosabua:
I added the register example now and as agreed with @dulanism will now merge this.

PR #135, opened 2016-02-17
Minor fixes

joedragons:
* I corrected a - count which seemed ok in the published book but was causing the preview in github to have block text and normal text inverted
* I made an adjustment to the resultant .npmrc to clarify confusion noticed in https://issues.sonatype.org/browse/NEXUS-9829


mosabua:
+1

mosabua:
I can take care of publishing this.

joedragons:
Thanks @mosabua =)

PR #136, opened 2016-02-19
Typo fixes

joedragons:
On scan of new section (RE: NEXUS-9757).
I am confident the above are typos but not 100% confident that my fixes are correct so making a PR rather than just merging.  Would appreciate a +1.

### DIFF
--- a/chapter-bundle-dev.asciidoc
+++ b/chapter-bundle-dev.asciidoc
@@ -150,7 +150,7 @@ and necessary steps for implementation;
 * How can the repository manager authenticate against a proxied remote server?
 * How does the concepts of components and assets used in {pro} map to the format?
 * What is the best way to map the component identifier of name, version and group to the format?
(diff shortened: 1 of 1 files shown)


dulanism:
+1 for grammatical fixes... though exempli gratia should be abbreviated as "e.g." not "E.g."

mrprescott:
+1

joedragons:
@dulanism This is done a lot through our documentation, so I think that's a bigger effort.  I rarely use those abbrevs myself, but I know E.e. was wrong=)  Thanks!

mosabua:
Latin has uppercase at the beginning of the sentence just like English, German and many other languages. 

http://english.stackexchange.com/questions/28172/when-a-sentence-starts-with-e-g-should-the-e-be-capitalized

Also using e.g. is okay as part of our style guide. But you can use "for example" as well.

mosabua:
Thanks for those fixes btw.

PR #139, opened 2016-02-22
[NEXUS-9273] Delete Component doc

dulanism:
CI http://bamboo.s/browse/BOOK-NEXUS99
JIRA https://issues.sonatype.org/browse/NEXUS-9273


mosabua:
There must be some text missing. Also we typically dont add screenshots like that.

mosabua:
And I assume this is work in progress ;-)

dulanism:
@mosabua literally just opened this up... quick on the draw :)

chapter-using-nexus-ui.asciidoc
```diff
@@ -380,6 +380,19 @@ with the 'Name' of the asset you want to inspect to view the asset information d
 .Example for Component Information and List of Associated Assets
 image::figs/orig/component-information.png[scale=35]
 
+To delete a component select one on the list, then press the 'Delete component' button above. A modal will pop up
```
joedragons: Minor: Above this already discussing details about having selected one, so I think "select one on the list, then" is unnecessary.  It is true however.
Minor: You could add a ref to the image (eg. as shown in Figure X).

chapter-using-nexus-ui.asciidoc
```diff
@@ -380,6 +380,19 @@ with the 'Name' of the asset you want to inspect to view the asset information d
 .Example for Component Information and List of Associated Assets
 image::figs/orig/component-information.png[scale=35]
 
+To delete a component select one on the list, then press the 'Delete component' button above. A modal will pop up 
+to confirm the deletion.
+
+[[delete-component-example.png]]
+.Delete Component from a Proxy or Hosted Repository
+image::figs/web/delete-component-example.png[scale=35]
```
joedragons: Minor: This image seems redundant with the one above it, except for the fact that it shows the button (which the other should, it's just now out of date).  Not really sure a solution but thought I'd mention.

chapter-using-nexus-ui.asciidoc
```diff
@@ -380,6 +380,19 @@ with the 'Name' of the asset you want to inspect to view the asset information d
...
+
+When you delete the component it will remove all associated assets. You can only delete 
+hosted and proxy repositories.
```
joedragons: Pretty sure this should be "You can only delete components from hosted and proxy repositories." (added 'components from')

chapter-using-nexus-ui.asciidoc
```diff
@@ -380,6 +380,19 @@ with the 'Name' of the asset you want to inspect to view the asset information d
...
+
+NOTE: Certain components will remain after deletion. While a Docker image can be deleted its layers component 
+cannot.
```
joedragons: This reads weird to me.  I think you're trying to say "image can be deleted, its asset layers cannot".  Admittedly that sounds weird since this is talking about deleting a component.  So another way to say it is "image can be deleted, the layers that make it up remain after its deletion."

joedragons:
+1 with comments (some minor).  Or happy to review again if you prefer that after making changes.

joedragons:
Minor: I also just noticed I don't see "delete-component.png" referenced, so while the addition doesn't hurt anything not sure the value at the moment.  Mentioning in case an oversight (to upload or to not use)

dulanism:
That's an older image I intend to remove

chapter-using-nexus-ui.asciidoc
dulanism: Are you talking about text starting at line 390? @joedragons

joedragons: @dulanism That is the text I'm talking about yes.  Currently "You can only delete hosted and proxy repositories."

chapter-using-nexus-ui.asciidoc
dulanism: Rewrite: While a Docker image can be deleted, the layers that make it up remain after its deletion.

chapter-using-nexus-ui.asciidoc
dulanism: I'm going to remove the older image & tweak the text that references it

chapter-using-nexus-ui.asciidoc
```diff
@@ -373,12 +373,18 @@ shows the age of the component.
...
-[[fig-component-information]]
+[[fig-component-details]]
 .Example for Component Information and List of Associated Assets
-image::figs/orig/component-information.png[scale=35]
+image::figs/web/component-details.png[scale=35]
+
+To delete a component press the 'Delete component' button as shown in <<fig-component-details>>. A modal will pop 
+up to confirm the deletion. You can only delete components from hosted and proxy repositories.
+
+NOTE: Certain component assets will remain after deletion. While a Docker image can be deleted, the layers that make it up remain after its deletion
```
joedragons: Minor: "While a Docker image can be deleted, the layers that make it up remain after its deletion" - think a period is missing here.  Also, suggest lead with "For example, " unless this is a point you especially want to make.

joedragons:
Rechecked at @dulanism 's request.  +1 again=)

mosabua:
The screenshot is NOT appropriate in this context since it does not show any metadata like most popular version, age and so on. Please fix this by using a Maven related component that has all that info.

dulanism:
@mosabua Oh. Okay. I understand what you mean. My apologies for the misunderstanding. Will fix.

PR #143, opened 2016-03-07
SSL cert re-write for Docker

dulanism:
JIRA https://issues.sonatype.org/browse/NEXUS-9870
(also includes) JIRA https://issues.sonatype.org/browse/NEXUS-9777
KB article: https://support.sonatype.com/hc/en-us/articles/217542177

### DIFF
--- a/chapter-docker.asciidoc
+++ b/chapter-docker.asciidoc
@@ -66,6 +66,9 @@ fact that Docker does not support context paths in the registry API.
 
 ==== Tips for SSL Certificate Usage
 
+////
+section needs re-write
+////
 The requirement of Docker to use HTTPS forces the usage of SSL certificates. By default, Docker looks up the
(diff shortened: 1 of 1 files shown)


peterlynch:
-1 to this PR so far...

The book needs to mention these main points 

- https is required to publish, this is a docker requirement, see docker docs
- if a self signed certificate if used inside Nexus, docker needs to trust this explicitly, in a manner described in the official docker docs,
- NEVER use --insecure-registry to attempt to get docker to trust your certificate THIS IS NOT SUPPORTED
- and finally, if you want to use keytool to generate the certificate instead of openssl, then see our help center article

Also this assumes somewhere in the doc we talk about
- how to configure Nexus Jetty to use an SSL connector
- how to configure the SSL port of each docker repository 
   - required for hosted, deployment
   - optional for proxy repositories and group access 

That said, if this PR merely deleted the entire SSL Docker in the book, I could also get behind that for now.

PR #149, opened 2016-03-14
Update privilege intro and subsection

dulanism:
JIRA: https://issues.sonatype.org/browse/NEXUS-9858
CI: http://bamboo.s/browse/BOOK-NEXUS111


chapter-security.asciidoc
```diff
@@ -92,22 +92,14 @@ including any administrative access and has to be avoided.
...
-
-The repository manager uses a number of types of privileges:
-
-application:: privileges related to the user interface and related functionality.
-repository-admin:: privileges related to the administration and configuration of a specific repository separated
-into access to '*' (all), 'add', 'browse', 'delete', 'edit' and 'read'.
-repository-view:: privileges controlling access to the content of a specific repository separate into access to
-'*' (all), 'add', 'browse', 'delete', 'edit' and 'read'.
+You can assign default or custom privileges to a role within the repository manager. Click 'Privileges' in the 
+left sidebar, nested under 'Security' to access the screen. In <<fig-privileges-list>> the 'Privileges' panel shows a list of built-in permissions your can assign to a role and, ultimately, apply to a user.
```
joedragons: While the above two first sentences are true, this reads a little strange to me, because ultimately clicking on the Privilege section is not the way you get privileges assigned to a role.
I'm not 100% sure what the sentences are trying to accomplish so it's hard for me to make any suggestions definitively.  One possibility is changing the opening sentence to something like "One or more priviliges comprise what access to the Nexus Repository Manager a role has."
If that misses what you're trying to say, feel free to ping me if I can help with further guidance.

chapter-security.asciidoc
```diff
@@ -92,22 +92,14 @@ including any administrative access and has to be avoided.
...
 
-Both 'repository-admin' as well as 'repository-view' privileges are automatically dcreated and deleted by the
-repository manager, when repositories are created or deleted.
+[[fig-privileges-list]]
+.List of Security Privileges
```
joedragons: Minor: You may want to say "partial list"/"example list" or show the jagged cut of the screen to show there are more.

chapter-security.asciidoc
```diff
@@ -92,22 +92,14 @@ including any administrative access and has to be avoided.
...
+image::figs/web/privileges-list.png[scale=80]
 
-The list of privileges displayed in <<fig-privileges-list>> displays an icon for the privilege 'Type' as the first
-column and then adds:
+The list of privileges displays an icon for the privilege Type as the first column and then adds:
```
joedragons: Minor: I see this was already there but I'm not keen on the word "adds".  It infers to me you're doing math or creating something.  Could strike it or substitute it with something like "follows" or "after", etc.

chapter-security.asciidoc
```diff
@@ -92,22 +92,14 @@ including any administrative access and has to be avoided.
...
-repository-view:: privileges controlling access to the content of a specific repository separate into access to
-'*' (all), 'add', 'browse', 'delete', 'edit' and 'read'.
+You can assign default or custom privileges to a role within the repository manager. Click 'Privileges' in the 
+left sidebar, nested under 'Security' to access the screen. In <<fig-privileges-list>> the 'Privileges' panel shows a list of built-in permissions your can assign to a role and, ultimately, apply to a user.
 
-Both 'repository-admin' as well as 'repository-view' privileges are automatically dcreated and deleted by the
-repository manager, when repositories are created or deleted.
+[[fig-privileges-list]]
+.List of Security Privileges
+image::figs/web/privileges-list.png[scale=80]
```
joedragons: In the PDF this is cutoff and probably needs scaled down more.  See attached:
screen shot 2016-03-15 at 10 53 00 am

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
 
 Further details are available after pressing on a specific row in the detail view.
 
-[[fig-privileges-list]]
-.List of Security Privileges
-image::figs/web/privileges-list.png[scale=60]
+Click the 'Create privilege' button to view a list of privilege types you can customize for your roles. This will 
+take you to a screen that lists a number of types of privileges, seen in <<fig-privileges-types>>.
```
joedragons: Minor: While these two sentences are true, I think there's some redundancy here that could be cleaned up.  Maybe just merge into one sentence and add on "as seen in" to the first.

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+
+[[fig-privileges-types]]
+.Choosing Privilege Types
+image::figs/web/privileges-types.png[scale=45]
+
+Select the privilege you want to apply to the role by clicking on the type. The privilege types are as follows:
```
joedragons: "Select the privilege you want to apply to the role by clicking on the type." sentence is not clear to me.  Perhaps "Select the privilege type corresponding to the area of the repository manager you wish to grant permissions."?

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+
+Application:: privileges related to the user interface and related functionality.
```
joedragons: While true, unless left intentionally generic, this could apply to any of the types.  For example, repo admin is also related to UI and related functionality.  Maybe change "related functionality" to "functionality related to the defined aspect of the application".  Or "to be defined".

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+Wildcard:: privileges related to search that will match any sequence of permissions.
+
+Below are some sample privileges you may assign to roles. The examples will show 'Application' and 
+'Repository View' privilege types. 
+
+In this example, create an Application privilege type.
+
+[[fig-privileges-application]]
+.Creating an Application Privilege
+image::figs/web/privileges-application.png[scale=80]
```
joedragons: Minor: Similar to the above, this is off screen in the PDF and may need to be scaled down.  Unlike the above, nothing important is cut, so marking minor.

screen shot 2016-03-15 at 11 13 57 am

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+
+The form provides 'Name', 'Description', 'Domain', and 'Actions'. In <<fig-privileges-application>>, above, the 
+form is completed for a role only allowed read access to an LDAP domain.
```
joedragons: "role" should be "privilege" unless you're trying to say privilege that can be assigned to a role, then I think you should be explicit.

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+Repository View:: privileges controlling access to the content of a specific repository; actions 
+can be defined with '*' (all), 'add', 'browse', 'delete', 'edit' and 'read' functions.
+Script:: privileges related to scripted commands capable of '*' (all), 'add', 'browse', 'delete', 'edit' and 
+'read' functions.
+Wildcard:: privileges related to search that will match any sequence of permissions.
+
+Below are some sample privileges you may assign to roles. The examples will show 'Application' and 
+'Repository View' privilege types. 
+
+In this example, create an Application privilege type.
```
joedragons: Minor: Saying "create" sounds instructional, like they should be following along.  Which can be the case but may not be.  Suggest "we create" or something non-instructional like "an Application privilege type is shown" (striking 'create').

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+
+[[fig-privileges-application]]
+.Creating an Application Privilege
+image::figs/web/privileges-application.png[scale=80]
+
+The form provides 'Name', 'Description', 'Domain', and 'Actions'. In <<fig-privileges-application>>, above, the 
+form is completed for a role only allowed read access to an LDAP domain.
+
+In another example, create a Repository View privilege type.
```
joedragons: Minor: Saying "create" sounds instructional, like they should be following along. Which can be the case but may not be. Suggest "we create" or something non-instructional like "an Application privilege type is shown" (striking 'create').

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+
+[[fig-privileges-repository-view]]
+.Creating a Repository View Privilege
+image::figs/web/privileges-repository-view.png[scale=80]
```
joedragons: Minor: Similar to above, this is off screen in the PDF and may need to be scaled down. Unlike above, nothing important is cut, so marking minor.

screen shot 2016-03-15 at 11 21 22 am

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+
+The form provides 'Name', 'Description', 'Format', 'Repository', and 'Actions'. In 
+<<fig-privileges-repository-view>>, above, the form is completed for a role allowed edit privileges for a Docker
```
joedragons: "role" should be "privilege".  Also recommend adding "only" around 'edit' to drive the point home.  Akin to this: "the form is completed for a privileged allowing only read and edit privileges for a...".  Note, I also added "read" because it does allow that and only sounded weird with out.  If you disagree with only, I still suggest adding read, otherwise what is said does not match what is shown IMO.

chapter-security.asciidoc
```diff
@@ -119,9 +111,49 @@ Permission:: the internal permission definition as used by the embedded security
...
+[[fig-privileges-repository-view]]
+.Creating a Repository View Privilege
+image::figs/web/privileges-repository-view.png[scale=80]
+
+The form provides 'Name', 'Description', 'Format', 'Repository', and 'Actions'. In 
+<<fig-privileges-repository-view>>, above, the form is completed for a role allowed edit privileges for a Docker 
+hosted repository.
+
+Click the 'Create privilege' button to save a new custom privilege. The privilege can be found listed among the 
+default privileges on the main 'Privileges' screen.
```
joedragons: While true, this has the potential to be lost after the examples.  You may want to move it before the examples.  If intentionally placed after the examples, such as the user is meant to follow along, you may want to put it after the first example as well, in case they only read one.  I favor telling how things are done then examples however.

chapter-security.asciidoc
dulanism: @joedragons revamp: 
Privileges can be assigned to a role in the {pro} and {oss} users. When you assign a privilege type, six underlying permissions are created: create, read, update, delete, edit, and browse. To start the process click 'Privileges' in the left sidebar, nested under 'Security'. In <<fig-privileges-list>> the 'Privileges' panel shows a list of built-in permissions your can assign to a role and, ultimately, apply to a user.

Also can you corroborate that six permissions are true? And if permissions is an accurate term?

chapter-security.asciidoc
dulanism: Updated to Partial... I can circle back to make any necessary image edits when I open the book review ticket in JIRA

chapter-security.asciidoc
dulanism: "....followed by:"

chapter-security.asciidoc
dulanism: Yeah, saw that convo. Adjusting size

chapter-security.asciidoc
dulanism: Updated

chapter-security.asciidoc
dulanism: Reading again, not clear to me, too. I like your suggestion. Revised

chapter-security.asciidoc
joedragons: "When you assign a privilege type, six underlying permissions are created:" - reads weird to me.  When you assign a privilege all the associated permissions are granted.  When you create a privilege, the six underlying actions are available (as well as * encompassing all).  However (and to answer your first Q), I caution generalizing this because some CRUD will work and some BREAD will work and as far as I know it's not interchangable.  It's unfortunately confusing in that regard.
Honestly, I don't really like the way it says "actions can be defined with * (all), add, browse, delete, edit and read." deeper in the doc, however, I believe this is done because it really shouldn't be generalized.  I didn't comment on review because I couldn't think of a better way to do it.

As far as second Q, permissions is an accurate definition of what is done but the term matching the variables is "action".

chapter-security.asciidoc
dulanism: resizing

chapter-security.asciidoc
dulanism: Fixed

chapter-security.asciidoc
dulanism: I'm gonna try "we create" as it keeps with the conversation tone "you-we"

chapter-security.asciidoc
dulanism: I'm gonna try "we create" as it keeps with the conversation tone "you-we"

chapter-security.asciidoc
dulanism: resized.

chapter-security.asciidoc
joedragons: Sorry, I also lied, there's create in crud and add in bread, so there's 7...I think.  If I can count;)

chapter-security.asciidoc
dulanism: updated

chapter-security.asciidoc
dulanism: point made. updated

dulanism:
I've marked 2-3 comments not yet revisited. Moving back to WIP till fixed.

chapter-security.asciidoc
```diff
@@ -92,22 +92,17 @@ including any administrative access and has to be avoided.
...
-
-The repository manager uses a number of types of privileges:
+Privileges can be assigned to a role for both {pro} and {oss} users.
 
-application:: privileges related to the user interface and related functionality.
-repository-admin:: privileges related to the administration and configuration of a specific repository separated
-into access to '*' (all), 'add', 'browse', 'delete', 'edit' and 'read'.
-repository-view:: privileges controlling access to the content of a specific repository separate into access to
-'*' (all), 'add', 'browse', 'delete', 'edit' and 'read'.
+To start the process click 'Privileges' in the left sidebar, nested under 'Security'. In <<fig-privileges-list>>
```
joedragons: I comment on this last time (and has been reworded) but these first two sentences are true, however, to assign a privilege to a role you do not need to go into privileges first (you can go right to roles).  I think currently the wording that makes it seem like you do is "To start the process".  If you switch this with something like "In order to see existing privileges" I think then it reads less this way.

chapter-security.asciidoc
```diff
@@ -92,22 +92,17 @@ including any administrative access and has to be avoided.
...
-The repository manager uses a number of types of privileges:
+Privileges can be assigned to a role for both {pro} and {oss} users.
 
-application:: privileges related to the user interface and related functionality.
-repository-admin:: privileges related to the administration and configuration of a specific repository separated
-into access to '*' (all), 'add', 'browse', 'delete', 'edit' and 'read'.
-repository-view:: privileges controlling access to the content of a specific repository separate into access to
-'*' (all), 'add', 'browse', 'delete', 'edit' and 'read'.
+To start the process click 'Privileges' in the left sidebar, nested under 'Security'. In <<fig-privileges-list>> 
+the 'Privileges' panel shows a list of built-in permissions you can assign to a role and, ultimately, apply to a
```
joedragons: Minor: Figure label was changed to "partial list" I think this should too.

chapter-security.asciidoc
```diff
@@ -92,22 +92,17 @@ including any administrative access and has to be avoided.
...
 
-Both 'repository-admin' as well as 'repository-view' privileges are automatically dcreated and deleted by the
-repository manager, when repositories are created or deleted.
+[[fig-privileges-list]]
+.Partial List of Security Privileges
+image::figs/web/privileges-list.png[scale=35]
 
-The list of privileges displayed in <<fig-privileges-list>> displays an icon for the privilege 'Type' as the first
-column and then adds:
+The list of privileges displays an icon for the privilege Type as the first column, followed by:
```
joedragons: Type in single quote markup?

chapter-security.asciidoc
```diff
@@ -119,9 +114,51 @@ Permission:: the internal permission definition as used by the embedded security
 
 Further details are available after pressing on a specific row in the detail view.
 
-[[fig-privileges-list]]
-.List of Security Privileges
-image::figs/web/privileges-list.png[scale=60]
+Click the 'Create privilege' button to view a list of privilege types you can customize for your roles, as seen
```
joedragons: "you can customize for your roles" - you don't customize the privilege types, you customize the privileges that the types are used to create.

chapter-security.asciidoc
```diff
@@ -119,9 +114,51 @@ Permission:: the internal permission definition as used by the embedded security
...
+image::figs/web/privileges-types.png[scale=35]
+
+Select the privilege type corresponding to the area of the repository manager you wish to grant permissions. The 
+privilege types are as follows:
+
+Application:: privileges related to a defined aspect of an application.
+Repository Admin:: privileges related to the administration and configuration of a specific repository.
+Repository View:: privileges controlling access to the content of a specific repository.
+Script:: privileges related to scripted commands capable of executing permissions granted.
+Wildcard:: privileges related to search that can match any sequence of permissions.
```
joedragons: Minor: This list consistently (good) ends with periods however I feel like this is atypical of our documentation, where list explainations that are sentence fragaments don't end in periods.

chapter-security.asciidoc
```diff
@@ -119,9 +114,51 @@ Permission:: the internal permission definition as used by the embedded security
...
+[[fig-privileges-types]]
+.Choosing Privilege Types
+image::figs/web/privileges-types.png[scale=35]
+
+Select the privilege type corresponding to the area of the repository manager you wish to grant permissions. The 
+privilege types are as follows:
+
+Application:: privileges related to a defined aspect of an application.
+Repository Admin:: privileges related to the administration and configuration of a specific repository.
+Repository View:: privileges controlling access to the content of a specific repository.
```
joedragons: Consider changing "access" to "visibility".  Access could mean getting at the components and this only affects viewing them (as far as I know).

chapter-security.asciidoc
```diff
@@ -119,9 +114,51 @@ Permission:: the internal permission definition as used by the embedded security
...
+Script:: privileges related to scripted commands capable of executing permissions granted.
+Wildcard:: privileges related to search that can match any sequence of permissions.
+
+In all 'Privilege Types', above, the variables assigned to a role are defined as 'Actions'. 'Actions' can 
+either be CRUD (create, read, update, and delele) or BREAD (browse, read, edit, add, and delete) storage
```
joedragons: * delete (under CRUD) has typo - shows as "delele"
* NOTE: I am +0 to putting CRUD and BREAD phrases in here, but I understand this as written and it is true.  It somewhat ephasizes our non-decision/non-consistency which is also (unfortunately) true.

chapter-security.asciidoc
```diff
@@ -119,9 +114,51 @@ Permission:: the internal permission definition as used by the embedded security
...
+Wildcard:: privileges related to search that can match any sequence of permissions.
+
+In all 'Privilege Types', above, the variables assigned to a role are defined as 'Actions'. 'Actions' can 
+either be CRUD (create, read, update, and delele) or BREAD (browse, read, edit, add, and delete) storage 
+functions.
+
+To save a new custom privilege click the 'Create privilege' button. The privilege can be found listed 
+among the default privileges on the main 'Privileges' screen.
+
+Below are some sample privileges you may assign to roles. The examples will show 'Application' and
```
joedragons: Minor: Unless specifically desiring to emphasize use, consider changing "assign to roles" to "create".  Reasoning, all privileges can be assigned to roles which I don't think is what you're trying to emphasize here.

chapter-security.asciidoc
```diff
@@ -119,9 +114,51 @@ Permission:: the internal permission definition as used by the embedded security
...
+'Repository View' privilege types. 
+
+In this example, we create an Application privilege type.
+
+[[fig-privileges-application]]
+.Creating an Application Privilege
+image::figs/web/privileges-application.png[scale=35]
+
+The form provides 'Name', 'Description', 'Domain', and 'Actions'. In <<fig-privileges-application>>, above, the 
+form is completed for a privilege only allowed read access to an LDAP domain.
```
joedragons: Consider adding another sentence with what this results as.  Like "If assigned this privilege, a user would be able to view LDAP administration configuration but not edit it, create a new LDAP configuration nor delete any existing LDAP configurations." (I may say configuration too much but you get the gist, I hope=)).

chapter-security.asciidoc
joedragons: Actually I want to retest this...will do now.

chapter-security.asciidoc
```diff
@@ -119,9 +114,51 @@ Permission:: the internal permission definition as used by the embedded security
...
+
+In another example, we create a Repository View privilege type.
+
+[[fig-privileges-repository-view]]
+.Creating a Repository View Privilege
+image::figs/web/privileges-repository-view.png[scale=35]
+
+The form provides 'Name', 'Description', 'Format', 'Repository', and 'Actions'. In 
+<<fig-privileges-repository-view>>, above, the form is completed for a privilege allowing only read and edit 
+privileges for a Docker hosted repository.
```
joedragons: Consider adding another sentence with what this results as.  Like "If assigned this privilege, a user would be able to view and edit docker images in the docker-internal repository however they could not delete them add them or even browse them via the UI."

chapter-security.asciidoc
joedragons: Nevermind on the above comments, deleting them all.

joedragons:
+1 = I'm comfortable enough to give this PR a +1 despite the fact some of my comments I don't think are minor and hopefully will be touched up.  Am happy to review again if you like.  Thanks!

chapter-security.asciidoc
dulanism: Knowing the privilege > role > user flow as you write here, I want to keep this introduction very simple 

> Privileges can be assigned to a role for both {pro} and {oss} users.

>To access 'Privileges' go to the 'Security' section. An extensive list of privileges for roles are >already built in the repository manager, partially depicted in <<fig-privileges-list>>. This feature >also allows your to create custom privileges for roles within the repository manager.

joedragons: Looks good to me.

PR #150, opened 2016-03-16
Docker SSL and repo config updates

dulanism:
JIRA https://issues.sonatype.org/browse/NEXUS-9870
CI http://bamboo.s/browse/BOOK-NEXUS112

### DIFF
--- a/chapter-docker.asciidoc
+++ b/chapter-docker.asciidoc
@@ -73,18 +73,17 @@ The requirement of Docker to use HTTPS forces the usage of SSL certificates. By 
 validity of the certificate by checking with certificate authorities. If you purchased a certificate
 that is registered with these authorities, all functionality works as desired. 
 
-If you create a certificate yourself with tools such as `openssl`, it is not registered. This is 
-often called a self-signed certificate. Using a self-signed certificate requires further configuration steps to 
-ensure that Docker can explicitly trust it.
+If you create a certificate yourself with tools such as `openssl`, it is self-signed and not registered. Using a 
+self-signed certificate requires further configuration steps to ensure that Docker can explicitly trust it.
+ 
+WARNING: Docker Daemon can stand up instances with the `--insecure-registry` flag to skip validation of 
+a self-signed certificate. But the repository manager does not support the use of the flag, as it generates known 
+bugs and other implementation issues.
 
-WARNING: Adminstrators may attempt a work-around with the `--insecure-registry` flag. Docker Daemon does not 
-validate it as trustworthy. We recommend you avoid using the flag as it leads to errors impacting certificate 
-validation.
-
-To generate a self-signed server certificate for the repository manager use Java keytool, a utility that lets you 
-manage your own private key pairs and certificates. See our 
-https://support.sonatype.com/hc/en-us/articles/217542177[knowledge base article] to set up this more trustworthy 
-solution.
+To generate a trustworthy self-signed certificate for the repository manager use keytool, a utility that 
+lets you manage your own private key pairs and certificates. See our 
+https://support.sonatype.com/hc/en-us/articles/217542177[knowledge base article] to learn how to configure the 
+utlilty.
 
 
 [[docker-registry-api]]
--- a/chapter-docker.asciidoc
+++ b/chapter-docker.asciidoc
@@ -66,6 +66,9 @@ fact that Docker does not support context paths in the registry API.
 
 ==== Tips for SSL Certificate Usage
 
+{pro} and {oss} are not configured with HTTPS connectors by default as it requires an SSL certificate 
+to be generated and configured manually.
+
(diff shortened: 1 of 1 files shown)


chapter-docker.asciidoc
```diff
@@ -66,66 +66,27 @@ fact that Docker does not support context paths in the registry API.
...
-----
-
-Red Hat Enterprise Linux/CentOS 7 uses the configuration file `etc/sysconfig/docker`. It contains a configuration
-line for the insecure registry setup that you can uncomment and edit as needed:
-
-----
-INSECURE-REGISTRY='--insecure-registry localhost:18843 --insecure-registry localhost:18444'
-----
+If you create a certificate yourself with tools such as `openssl`, it is not registered. This is 
+often called a self-signed certificate. Using a self-signed certificate requires further configuration steps to
```
joedragons: Minor: While true, word "often" reads weird to me.  Consider mashing up the first two sentences into something like 'If you create a certificate yourself with tools such as `openssl`, it is self-signed and not registered.'

chapter-docker.asciidoc
```diff
@@ -66,66 +66,27 @@ fact that Docker does not support context paths in the registry API.
...
-----
-INSECURE-REGISTRY='--insecure-registry localhost:18843 --insecure-registry localhost:18444'
-----
+If you create a certificate yourself with tools such as `openssl`, it is not registered. This is 
+often called a self-signed certificate. Using a self-signed certificate requires further configuration steps to 
+ensure that Docker can explicitly trust it.
 
-Reloading the configuration and restarting the docker daemon activates the new settings and can be performed with
+WARNING: Adminstrators may attempt a work-around with the `--insecure-registry` flag. Docker Daemon does not 
+validate it as trustworthy. We recommend you avoid using the flag as it leads to errors impacting certificate
```
joedragons: "Docker Daemon does not validate it as trustworthy" - I realize this is a touchy phrasing, but I do not believe this is true as written, since we CAN get instances stood up with --insecure_registry so it does validate as trustworthy and work.  To me it just seems like there are problems with insecure registry in some cases (known bugs or incomplete implementation or just certain usages) where we decided not to support it.  I'm not sure how you can say that nice and succinctly tho.  Maybe @peterlynch has thoughts.

chapter-docker.asciidoc
```diff
@@ -66,66 +66,27 @@ fact that Docker does not support context paths in the registry API.
...
+validation.
 
-----
-systemctl daemon-reload
-systemcttl restart docker
-----
+To generate a self-signed server certificate for the repository manager use Java keytool, a utility that lets you 
+manage your own private key pairs and certificates. See our 
+https://support.sonatype.com/hc/en-us/articles/217542177[knowledge base article] to set up this more trustworthy 
+solution.
```
joedragons: Some thoughts, all minor:
* I'm +0 on the word server in "..a self-signed server certificate..". It's true, but not sure it adds anything
* Suggest possessive Java (Java's keytool) or just removing Java.
* I'm +0 on the word more in "..this more trustworthy solution."  My guess is it is in comparison to insecure registry however since we're not recommending that, then all the word does is add emphasis to that fact.  I guess the same arguement could be made of the word trustworthy, less strong about that.

joedragons:
+1 caveat I think it's worth getting @peterlynch and maybe @DarthHater 's thoughts on the warn as I think it worth getting a double check on what we want to say.  Definitively don't want to accuse anyone of anything even inadvertantly.  Rest of my notes were minor.

chapter-docker.asciidoc
dulanism: I'd call that a colloquial gaffe. Tweaking per your suggestion

chapter-docker.asciidoc
dulanism: Wrote: 

>WARNING: Docker Daemon can stand up instances with the `--insecure-registry` flag to skip >validation of a self-signed certificate. But the repository manager does not support the use of the >flag, as it generates known bugs and other implementation issues.

joedragons: The only thing I am unsure of is you might put "Sonatype" and not "the repository manager".  But you might clear that with @DarthHater or @peterlynch .  I like the rewrite otherwise.

joedragons:
+1 still - rechecked on request.

mosabua:
I thought the agreement was to still have the documentation but add a warning about it not being supported. Now its all gone so the support team might get more questions again. Was that cleared with @rseddon and @peterlynch  .. we got  the content from them initially to be added.

dulanism:
Hi @mosabua. Yes it was discussed with @peterlynch and the team already. If we don't support --insecure-registry due to known bugs we thought, "why keep it"? I do have a backup of the earlier section, of which we can discuss later this week.

mosabua:
That's fine by me then.

PR #152, opened 2016-03-18
Replaced incorrect npm registry url in NPM chapter intro.

andham:
Signed-off-by: Anders Hammar <anders@hammar.net>

### DIFF
--- a/chapter-npm.asciidoc
+++ b/chapter-npm.asciidoc
@@ -9,8 +9,8 @@ The command line tool +npm+ is a package management solution for Javascript-base
 create and use 'node packaged modules' and is built into the popular Javascript platform
 http://www.nodejs.org[Node.js], which is mostly used for server-side application development.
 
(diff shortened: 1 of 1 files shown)


mosabua:
When talking about configuring the registry .. the URL you used (registry... is correct). However in this context here it is more about the user interface and search. Which is available at www ... so I would leave it as is and not merge this.

peterlynch:
I agree using "is the default package registry" when referring to www.npmjs.org is more confusing than referring to registry.npmjs.org.

I don't believe even mentioning www.npmjs.org adds much value in our book and I could get behind removing the entire paragraph completely.

In the proxy repository section further down, we tell people the correct url to use anyway.

mosabua:
I will find a new wording next week and sort this out.

mosabua:
Fixed now with 

https://github.com/sonatype/nexus-book/commit/8872df687e27166dd603c425646a137005bc86f2

https://github.com/sonatype/nexus-book/commit/77b99a32b087cab6654d16cd93e8e3ec85017183

Will publish live shortly

PR #154, opened 2016-03-20
Service config for 3.0

dulanism:
CI http://bamboo.s/browse/BOOK-NEXUS114
JIRA https://issues.sonatype.org/browse/NEXUS-9851
and JIRA https://issues.sonatype.org/browse/NEXUS-9759 -  Installer

*note branch conflict to resolve before merging


mosabua:
As a big fat first note you should mention that the installer includes a feature to install and configure the service (and test that).

And this section then only applies to manual installation from the archives... and link it up respectively from those sections.

dulanism:
@mosabua Not clear on second part: "And this section then only applies to manual installation from the archives... and link it up respectively from those sections."

Yes, this section is manual installation. But are you asking write/edit running installer sections for each os (e.g. os x, linux, window), then link each section to the respective manual installation sections? Sorry if I'm not clear.

chapter-installation.asciidoc
```diff
@@ -267,16 +277,41 @@ After changes to the files above, symlink +$NEXUS_HOME/bin/nexus+ to +/etc/init.
...
+----
+cd /etc/init.d
+update-rc.d nexus defaults
+sudo service nexus start
+----
+
+After starting the service for any Linux-based operating systems, verify that the service started successfully.
+
+----
+tail -f /usr/opt/nexus/logs/wrapper.log
```
mosabua: Thats the wrong path/file for 3.x ... not sure but probably should be 

/opt/nexus/data/log/karaf.log

but you will have to test that..

chapter-installation.asciidoc
```diff
@@ -537,7 +567,7 @@ repository manager and its components. If yes, click 'Next'.
 The next step displays the directory of the repository manager. All configurations, components, assets, and 
 {oss}-related data will be removed from the directory if you check 'Yes, remove Nexus application data'.
 
-Click 'Next' to complete the uninstall process.
+Click 'Next', then 'Finish' to complete the uninstall process.
```
mosabua: Add uninstalling when installed from the archive (= remove service config, delete directories)

PR #155, opened 2016-03-21
Docker chapter improvements

joedragons:
Based on entire chapter review  after NEXUS-9777 work was completed to make sure no side affects as advertised.
Being NEXUS-9777 is already closed and this is my own initiative, there is no correctsponding JIRA ticket for this.

Mostly typo fixes however three more serious rewrites of note (none of which were affects from NEXUS-9777 as far as I can tell):
* The previous write suggests you do not enable v1 except in certain circumstances.  I have never agreed with this nor do I see harm in enabling it (it acts as fallback from v2, so no additional failures would be caused; if v2 and v1 fails, you're no worse off than if v2 fails & if v2 works v1 is never hit) so I rewrote the section to clarify if people are unsure they should enable it.  I hope this helps get less support and community contacts about forgetting to enable v1.
* The previous write says that if you're not authenticated, the system prompts you.  This is no longer true at least as of docker 1.10 so I adjusted it such that it says it may prompt you but otherwise you need to docker login.
* The previous write says when tagging docker image ID was required which is not true, you can use image name.

In all cases I tried to leave as much text existing as possible while maintaining the corrected statements.

### DIFF
--- a/chapter-docker.asciidoc
+++ b/chapter-docker.asciidoc
@@ -17,8 +17,8 @@ more and more other publicly available registries such as the https://cloud.goog
 Container Registry] and others.
 
 {pro} and {oss} support Docker registries as the Docker repository format for hosted and proxy repositories. You
-can then expose these repositories to the client-side tools directly or as a repository group, which is a
-repository that merges and exposes the contents of multiple repositories in one convenient URL.
+can expose these repositories to the client-side tools directly or as a repository group, which is a repository 
+that merges and exposes the contents of multiple repositories in one convenient URL.
 
 This allows you to reduce time and bandwidth usage for accessing Docker images in a registry as well as share your
 images within your organization in a hosted repository. Users can then launch containers based on those images,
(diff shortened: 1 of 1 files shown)


dulanism:
+1

chapter-docker.asciidoc
```diff
@@ -216,8 +217,8 @@ repository.  Individual login operations must be performed for each repository a
 access in an authenticated manner.
 
 TIP: Specifically when planning to push to a repository a preemptive login operation is advisable as it removes
-the need for use interaction and is therefore suitable for continuous integration server setups and the
-automations scenarios.
+the need for use interaction and is therefore suitable for continuous integration server setups and autonomous
```
mosabua: Should be "automation"

joedragons: I've never heard this word as an adjective so I went with another
(incorrectly assuming it was wrong, apparently).
Can we agree on "automated"?
Am I missing the sentence point altogether?
Thanks,
Joe

mosabua: The aim is to mention the usage in scenarios where everything is automated and performed without user interaction - an "automation scenario".  Feel free change but autonomous scenario is definitely not what is intended in this context..

joedragons: As I've never heard that expression before, I'll change to "automated"
(which I believe is the same, as you even used the word in your
explaination) and if you feel strongly, then I'll let you adjust further.
Apologies for my confusion.
Thanks!

PR #157, opened 2016-03-23
Document scheduled task to mvn snapshot

dulanism:
CI: http://bamboo.s/browse/BOOK-NEXUS116
JIRA: https://issues.sonatype.org/browse/NEXUS-9943


mosabua:
I would NOT add a screenshot ... the list is just going to get longer and a picture provides no real value imho. Or at least cut it down in size.

kellyrob99:
Agreed on the screenshot

dulanism:
Okay @kellyrob99 & @mosabua

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,37 @@ for supporting search and browse actions.
 Remove Maven indexes:: This task is the counter-part to the task 'Publish Maven indexes' and
 can remove the index.
 
+Remove Snapshots from Maven Repository:: This task can be scheduled to remove Maven Snapshots. Often, you will 
+want to remove snapshots from a snapshot repository to preserve storage space. This task supports this deletion 
+for time stamped snapshots as created by Maven 3.x in a deployment repository. Note that configuring and
+running this job is not enough to reclaim disk space. Files are not deleted by the 'Remove Snapshots' job. They 
+are only moved into the trash folder. Therefore you will also need to configure a scheduled job to empty 
+the trash folder. When you create a scheduled task to remove snapshots, you can specify the 'Repository/Group' to
```
kellyrob99: there is no trash folder in NX3, the equivalent would be Compact Blobstore to reclaim space

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,37 @@ for supporting search and browse actions.
...
+preserve per component.  For example, if you configured this option with a value of 2, the repository manager will
+always preserve at least two snapshot components. A value of -1 indicates that all snapshots should be preserved.
+
+'Snapshot retention (days)';; This configuration option allows you to
+specify the number of days to retain snapshot components.  For example,
+if you want to make sure that you are always keeping the last three
+day's worth of snapshot components, configure this option with a value
+of 3. The minimum count overrides this setting.
+
+'Remove if released';; The checked 'Remove if released' will dump all corresponding snapshot releases.
```
kellyrob99: delete/remove instead of dump makes more sense here

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,37 @@ for supporting search and browse actions.
...
+
+'Grace period after release (days)';; This can lead to builds that still reference the snapshot
+dependency to fail. This grace period parameter allows you to specify a number of days to delay the deletion, 
+giving the respective projects referencing the snapshot dependency time to upgrade to the release component or 
+the next snapshot version.
+
+CAUTION: The deletion of Maven snapshots when a release occurs takes precedence over the number you select 
+in the 'Minimum snapshot count' dropdown. If configured improperly, this task configuration may supplant the
```
kellyrob99: the entire 'supplant' sentence can probably be deleted.

chapter-configuration.asciidoc
```diff
@@ -334,24 +334,23 @@ The following task types are available to perform specific maintenance:
 
 Compact blob store:: Content deleted from a <<admin-repository-blobstores,blob store>> is not physically deleted
 from the storage device. Instead it is only internally marked for deletion. This task performs the actual
-deletion of the relevant files, and therefore frees up the storage space.
-////
+deletion of the relevant files, and therefore frees up the storage space. On this form you are required to 
+assign a 'Task name', 'Blob store', and 'Task frequency'. Optionally, you can assign a notification email.
```
joedragons: +0 to this addition as written.  Specifically, Task name, Task frequency and optionally email are standard Task configurations.  If you are going to add this here, nothing prevents you from adding it everywhere.  I believe the general task variable definition above describes these fields.  If you think there needs to be more emphasis (besides blob store), I'd consider changing that section.  If you believe emphasis on this specific task is needed, I am not sure why.
If you had just said "you are required to assign 'Blob store'" I wouldn't have made any comment.

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,35 @@ for supporting search and browse actions.
 Remove Maven indexes:: This task is the counter-part to the task 'Publish Maven indexes' and
 can remove the index.
 
+Remove Snapshots from Maven Repository:: This task can be scheduled to remove Maven Snapshots. Often, you will
```
joedragons: Task case changed in most recent build to "Remove snapshots from Maven repository".  This case now matches the case wording of the other tasks.

chapter-configuration.asciidoc
dulanism: I meant to take that out per last commit.

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,35 @@ for supporting search and browse actions.
...
+
+'Minimum snapshot count';; This configuration option allows you to specify a minimum number of snapshots to
+preserve per component.  For example, if you configured this option with a value of 2, the repository manager will
+always preserve at least two snapshot components. A value of -1 indicates that all snapshots should be preserved.
+
+'Snapshot retention (days)';; This configuration option allows you to
+specify the number of days to retain snapshot components.  For example,
+if you want to make sure that you are always keeping the last three
+day's worth of snapshot components, configure this option with a value
+of 3. The minimum count overrides this setting.
```
joedragons: Minor: While this is true and matches the NX2 doc, you may consider changing "minimum count" to "'Minimum snapshot count'" (note added single quote markup and matched the variable name) to make sure there's no confusion over what we're referencing.

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,35 @@ for supporting search and browse actions.
...
+
+'Remove if released';; The checked 'Remove if released' will delete/remove all corresponding snapshot releases.
```
joedragons: I'm worried that this is less clear (than NX2) as written.  Before I checked against NX2, I thought a mention to GAV might help clarity.  Apparently NX2 has this mention.
For ref: "If enabled and a released component with the same GAV coordinates is detected all snapshots will be removed."
I am wondering if this was already discussed to change earlier and that's why this is different.  If that's so, I defer to that choice.
Otherwise some mashup of these two I think could be better.
Like "If checked, all snapshots will be removed matching any released component found with the same GAV coordinates."

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,35 @@ for supporting search and browse actions.
...
+
+'Grace period after release (days)';; This can lead to builds that still reference the snapshot
```
joedragons: It appears there's a leading sentence missing from NX2 copyover.  "The configuration Remove if released causes snapshots to be deleted as soon as the scheduled task is executed."  Without this sentence, there's no clue on WHAT can lead builds to fail.  If this is not an oversight and this was intentionally struck, I think the (now first) sentence needs to be tweaked or also removed.
My guess is this is an oversight tho, I believe the NX2 sentence missing is still true.

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,35 @@ for supporting search and browse actions.
...
+dependency to fail. This grace period parameter allows you to specify a number of days to delay the deletion,
```
joedragons: While there in NX2, the word "delay" here surprises me.  To me this infers after the grace period the items will be deleted which would be news to me (haven't tested for this).  I suspect that it just ignores snapshots that are newer than the grace period but I could be wrong.
@kellyrob99 would probably know but I suspect he'd not want us to delay this forward motion for him.
Might be worth pinging @fjmilens3 if he has bandwidth to see what he sees in the code and/or I can try and setup a test over the weekend.  Or we can go with delay and then correct later if we find it to be wrong (both versions perhaps).

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,35 @@ for supporting search and browse actions.
...
+of 3. The minimum count overrides this setting.
+
+'Remove if released';; The checked 'Remove if released' will delete/remove all corresponding snapshot releases.
+
+'Grace period after release (days)';; This can lead to builds that still reference the snapshot
+dependency to fail. This grace period parameter allows you to specify a number of days to delay the deletion, 
+giving the respective projects referencing the snapshot dependency time to upgrade to the release component or 
+the next snapshot version.
+
+CAUTION: The deletion of Maven snapshots when a release occurs takes precedence over the number you select
```
joedragons: Suggest reworking this to specifically include 'Remove if released' as that takes some of the guesswork out.  Like "Deletion of Maven shapshots when 'Remove if release' is checked takes..."

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,35 @@ for supporting search and browse actions.
...
+
+'Remove if released';; The checked 'Remove if released' will delete/remove all corresponding snapshot releases.
+
+'Grace period after release (days)';; This can lead to builds that still reference the snapshot
+dependency to fail. This grace period parameter allows you to specify a number of days to delay the deletion, 
+giving the respective projects referencing the snapshot dependency time to upgrade to the release component or 
+the next snapshot version.
+
+CAUTION: The deletion of Maven snapshots when a release occurs takes precedence over the number you select 
+in the 'Minimum snapshot count' dropdown. Also, it is possible to configure the task in such a
```
joedragons: Minor: I would change "dropdown" to "field".  If you prefer to put the specific field type I believe it's called a "spinbox" or "spin-box".  @slhenty could tell you for sure or I could recheck.  I think genericizing in this case is fine.

chapter-configuration.asciidoc
```diff
@@ -376,6 +375,35 @@ for supporting search and browse actions.
...
+'Remove if released';; The checked 'Remove if released' will delete/remove all corresponding snapshot releases.
+
+'Grace period after release (days)';; This can lead to builds that still reference the snapshot
+dependency to fail. This grace period parameter allows you to specify a number of days to delay the deletion, 
+giving the respective projects referencing the snapshot dependency time to upgrade to the release component or 
+the next snapshot version.
+
+CAUTION: The deletion of Maven snapshots when a release occurs takes precedence over the number you select 
+in the 'Minimum snapshot count' dropdown. Also, it is possible to configure the task in such a 
+way that the results are unexpected. For example, if configured to keep 0 minimum snapshots older than 0 days,
```
joedragons: Minor: Suggest changing "are unexpected" to "may be unexpected" or "may seem unexpected".

joedragons:
NOTE: I've reviewed this before I've tested the latest fixes but I am hoping that doesn't matter.  If I note it does, I'll come back and comment.

I did notice that in "Files Changed" it shows two screenshots added while above it says they're not needed.  My guess is the PR as written will still add them and they just arn't referenced which seems unnecessary but also not a big deal.  Mentioning since I noticed.

chapter-configuration.asciidoc
dulanism: Re-write option: If checked, all snapshots that match any released component found with the same GAV coordinates will be removed.

joedragons: +1

chapter-configuration.asciidoc
dulanism: I think this serve as better tie in (starting from 'Remove if released')

> Remove if released: If checked, all snapshots that match any released component found with the same GAV coordinates will be removed.

> Grace period after release (days): This grace period parameter allows you to specify a number of days to delay the deletion, giving the respective projects referencing the snapshot dependency time to upgrade to the release component or the next snapshot version.

chapter-configuration.asciidoc
dulanism: Good point

chapter-configuration.asciidoc
dulanism: Conditionally. Yes, makes sense.

chapter-configuration.asciidoc
dulanism: I like learning new words but I'll just leave it as field for now. But open to refutation :)

chapter-configuration.asciidoc
dulanism: Minor tweak fixed

chapter-configuration.asciidoc
fjmilens3: Okay, some things here (for @dulanism and @joedragons):

The grace period does not actually delay the removal of a snapshot (it's not like it's enqueued and gets removed from the system later). The grace period is used when purging released snapshots, so that snapshots that 1) have an associated release but 2) have an updated timestamp that falls within the grace period aren't purged from the system.

joedragons: Cool thanks @fjmilens3.  That's how I suspected it worked.  Appreciate the confirm.

joedragons: Actually, along the vein of correcting the NX2 documentation as well as making sure that's how it SHOULD work (rather than understanding how it now works), @fjmilens3 did you perchance compare against how it works in NX2?

dulanism:
I didn't reference the images because after committing, @mosabua & @kellyrob99 suggested they have little value to the user. I'll remove both.

chapter-configuration.asciidoc
fjmilens3: Checked just now. The actual implementation of course is markedly different in the NX2 codebase, but the overall intent seems to be the same. Specifically, no, it's not like it's going to magically delete something later without a separate run of the task; it's just there as a criterion that has to be satisfied to proceed with the removal.

(It's in `DefaultSnapshotRemover#releaseExistsForSnapshot()`, and usages thereof for context, if anyone's interested.)

dulanism: Thanks for the clarification, @fjmilens3

chapter-configuration.asciidoc
```diff
@@ -376,6 +374,38 @@ for supporting search and browse actions.
...
+
+'Snapshot retention (days)';; This configuration option allows you to specify the number of days to retain 
+snapshot components.  For example, if you want to make sure that you are always keeping the last three
+day's worth of snapshot components, configure this option with a value of 3. The 'Minimum snapshot count' 
+configuration overrides this setting.
+
+'Remove if released';; If checked, all snapshots that match any released component found with the same GAV 
+coordinates will be removed.
+
+'Grace period after release (days)';; This grace period parameter allows you to specify a number of days before
```
joedragons: Minor: "This grace period parameter...".  The value of stating the parameter as an adjective right after you give the name of the parameter seems minimal.  "This parameter..." seems equally fine to me.

chapter-configuration.asciidoc
```diff
@@ -376,6 +374,38 @@ for supporting search and browse actions.
...
+day's worth of snapshot components, configure this option with a value of 3. The 'Minimum snapshot count' 
+configuration overrides this setting.
+
+'Remove if released';; If checked, all snapshots that match any released component found with the same GAV 
+coordinates will be removed.
+
+'Grace period after release (days)';; This grace period parameter allows you to specify a number of days before 
+released snapshots are purged. If a release associated to a snapshot has an updated timestamp and falls 
+within the set grace period, it will not be purged. However, the next time the task runs it will be removed
+since the timestamp will be expired.
```
joedragons: Do not believe the "However" sentence is true.  If you switch to "may" then it becomes true however not sure what value that adds.  Example, if my grace period is 100 days and the snapshot is 1 day old, if I run it tomorrow, the now 2 day old snapshot would still not be removed.

chapter-configuration.asciidoc
```diff
@@ -376,6 +374,38 @@ for supporting search and browse actions.
...
+
+This setting will give the respective project that references the snapshot dependency time to 
+upgrade to the release component or the next snapshot version.
```
joedragons: This does not display as intended (included with Grace period text in previous paragraph).  I think the intent is it should be but not 100% sure.  Noticed reviewing the pdf, might appear different in html.

chapter-configuration.asciidoc
dulanism: I miss understood an earlier comment. It make sense to me to strike the last sentence just keeping the first two. The sentence "If a release associated to a snapshot has an updated timestamp and falls within the set grace period, it will not be purged." aptly defines that parameter does.

joedragons:
+1 tho the final outstanding Q in NEXUS-9421 might require a tweak.

mosabua:
I corrected the wording around Maven SNAPSHOT versions, added example and other minor things. 

At this stage I am +1 for merge.

One thing I would like to know though... does this task delete immediately or move to trash?

dulanism:
From what I learned there is no trash folder in NX3. The equivalent would be Compact Blobstore to reclaim space

mosabua:
Great.. ready to go then

PR #159, opened 2016-03-28
Misc privilege section

mosabua:
Privilege section updates in terms of images, corrected text and more.


mosabua:
Just fyi @dulanism

PR #160, opened 2016-03-30
Adjust necessary images to fit screen width of 1280 for overall documentation 

dulanism:
For images currently published in the v3 book
JIRA: https://issues.sonatype.org/browse/NEXUS-9906
CI: http://bamboo.s/browse/BOOK-NEXUS119


mosabua:
These images are WAY to big still... the overall width of the documentation has to work for 1280 width .. that means the images need to be quite a bit smaller.. Please run a local build and verify that the image does not go beyond the right margin in the HTML output.

mosabua:
You can also show a little less information e.g. from the header or shrink the image in an image editor if absolutely necessary. Ping me if you need help with this

dulanism:
Will come back to assets-info & assets-attribute. The conversion from /orig/ to /web/ still doesn't render the border.

mosabua:
If you need help with the image border stuff I can debug the convert script..

dulanism:
Will do another round-robin later this week, early next.

joedragons:
I reviewed the PDF and looks good.  Going to put some minor comments in JIRA soas not to spam this PR with images.
While reviewing the PDF, I definately think the site needs reviewed because I noticed two images oversized in the current website, so no +1 yet, but ongoing=)

joedragons:
Also reviewed the epub and left comments in JIRA.  I misread the status, so am holding off on site until @dulanism is done=)

mosabua:
I would merge and publish asap. We can always refine on the go in the next days but we need to get the right images live NOW.

PR #161, opened 2016-03-31
Json format fix

dulanism:
added quotes around keys, changed to curly brackets

### DIFF
--- a/chapter-scripting.asciidoc
+++ b/chapter-scripting.asciidoc
@@ -222,13 +222,13 @@ log.info('Script security completed successfully')
 Passing parameters to the script has to use JSON encoded arguments like
 
 ----
-[
-  id          : 'foo',
-  name        : 'bar',
-  description : 'baz',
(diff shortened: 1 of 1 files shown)


dulanism:
PR created to make sure this format (JSON) can be properly parsed with Groovy JsonSlurper class, as noted by @bradbeck.

mosabua:
Well... the original one was tested to be working. If you change it.. you have to verify that the new markup works. It should.. but its up to you to ensure that @dulanism

kellyrob99:
+1, tested and works

mosabua:
Thanks for verifying @kellyrob99 ... then I am +1 as well .

Btw @dulanism  ... when you change docs in the scripting chapter you potentially also need to fix/adapt/test the actual scripts. They can be found here https://github.com/sonatype/nexus-book-examples/tree/nexus-3.0.x/scripting and are linked from the docs.

PR #166, opened 2016-04-12
Porting & updating proxy server configuration, base url setup

dulanism:
JIRA: https://issues.sonatype.org/browse/NEXUS-9226
https://issues.sonatype.org/browse/NEXUS-9292


chapter-configuration.asciidoc
```diff
@@ -279,6 +271,24 @@ the 'Hosts to exclude from HTTP/HTTPS proxy' setting. You can add a hostname in
...
+==== Base URL Creation
+
+You can change the 'Base URL' for your repository manager installation, which is used when generating links in
+emails and RSS feeds. For example, the instance for Sonatype development is available at
+http://respository.sonatype.org, and it makes use of this 'Base URL' field to ensure that links in emails and RSS
+feeds point to the correct URL. Internally the repository manager is running on a different port and context than
+the public port 80 and root context.
+
+[[fig-capability-base-url]]
+.Creating Base URL Capability
```
mosabua: Not sure we need an image here. I dont see any value it provides..

chapter-configuration.asciidoc
```diff
@@ -279,6 +271,24 @@ the 'Hosts to exclude from HTTP/HTTPS proxy' setting. You can add a hostname in
 <<fig-admin-system-http-proxy>> shows the 'HTTP Proxy' administration interface. The HTTPS configuration interface
 looks the same and is found below the HTTP configuration.
 
+[[admin-base-url]]
+==== Base URL Creation
+
+You can change the 'Base URL' for your repository manager installation, which is used when generating links in
+emails and RSS feeds. For example, the instance for Sonatype development is available at
```
mosabua: There are no RSS feeds.

chapter-configuration.asciidoc
```diff
@@ -279,6 +271,24 @@ the 'Hosts to exclude from HTTP/HTTPS proxy' setting. You can add a hostname in
...
+http://respository.sonatype.org, and it makes use of this 'Base URL' field to ensure that links in emails and RSS
+feeds point to the correct URL. Internally the repository manager is running on a different port and context than
+the public port 80 and root context.
```
mosabua: This is done via a proxy server setup.. which is why you need that documented and linked to. Having a proxy around NXRM is the common use case for having baseurl.. without it you dont need to set it because the repo manager knows on its own what the URL is. You can also configure baseurl if you want a specific DNS server name to be used in emails and more instead of e.g. a IP number.

chapter-configuration.asciidoc
```diff
@@ -64,6 +64,7 @@ typically need to configure, after successful installation. The following sectio
 * Advanced configuration with <<admin-system-capabilities,Capabilities>>
 * <<admin-system-emailserver,Email/SMTP server configuration>>
 * <<admin-system-http,HTTP/HTTPS proxy server configuration>>
+* Setting a <<admin-base-url,Base URL>> for an application
```
mosabua: the base URL is not for an application.. it is for the Nexus Repository Manager..

chapter-configuration.asciidoc
```diff
@@ -228,7 +220,7 @@ image::figs/web/admin-system-general.png[scale=50]
 TIP: This configuration is especially important if the repository manager is proxied by an external proxy server
```
peterlynch: This wording is confusing. proxied, external proxy server, protocol..

Nexus has HTTP proxy server config, proxy repositories, and reverse proxies in front of it. This tip is trying to talk about reverse proxies.

At a minimum remove this entire tip it adds noise and potential confusion, no value

dulanism: Thanks @peterlynch. It was hidden comment I inherited. I'll remove it.

chapter-configuration.asciidoc
```diff
@@ -279,6 +271,19 @@ the 'Hosts to exclude from HTTP/HTTPS proxy' setting. You can add a hostname in
 <<fig-admin-system-http-proxy>> shows the 'HTTP Proxy' administration interface. The HTTPS configuration interface
 looks the same and is found below the HTTP configuration.
 
+[[admin-base-url]]
+==== Base URL Creation
+
+The `Base URL` is the URL via which users can access the repository manager. This setting must be activated if
```
peterlynch: "...is the URL endpoint at which users should access the user interface."

that is all that it is. we can't even say it is the base URL for repositories anymore given docker repo semantics.

it also does NOT **need** to be activated - it should be set explicitly to match the base url including context path / or /nexus or /nexus3 that end users use to access the UI.

chapter-configuration.asciidoc
```diff
@@ -279,6 +271,19 @@ the 'Hosts to exclude from HTTP/HTTPS proxy' setting. You can add a hostname in
...
 looks the same and is found below the HTTP configuration.
 
+[[admin-base-url]]
+==== Base URL Creation
+
+The `Base URL` is the URL via which users can access the repository manager. This setting must be activated if 
+you want to host the repository manager behind a reverse proxy server. You do not need to set the base URL if 
+you are not proxying the repository manager, because it knows its own default URL. 
+
+To configure and run a proxy server, see <<install-proxy>>.
```
peterlynch: install-proxy is a bad anchor name due to previous 'proxy' naming confusion. same with 'proxy server` term. We must make sure we use the full term name to avoid confusion.

- Reverse proxy
- proxy repository
- HTTP proxy server

suggest just removing this sentence entirely and if 'reverse proxy' term is being used anywhere, link to a definition of it.

chapter-installation.asciidoc
```diff
@@ -373,6 +373,225 @@ manually start it after the configuration you can use
 sudo launchctl load /Library/LaunchDaemons/com.sonatype.nexus.plist
 ----
 
+[[install-proxy]]
+=== Running Behind a Reverse Proxy
+
+{inall}
+
+Nexus Repository Manager is a sophisticated server application with a web-application user interface, answering
```
peterlynch: does the product name Nexus Repository Manager have a macro or something?

chapter-installation.asciidoc
```diff
@@ -373,6 +373,225 @@ manually start it after the configuration you can use
...
+
+* security and auditing concerns
+* network administrator familiarity
+* organizational policy
+* disparate application consolidation
+* virtual hosting
+* exposing applications on restricted ports
+* SSL termination
+
+This book provides some general guidance on how to configure common reverse proxy servers to work with {pro} and
```
peterlynch: "This book" doesn't seem correct. We are talking about what is below in this section of the book.

chapter-installation.asciidoc
```diff
@@ -373,6 +373,225 @@ manually start it after the configuration you can use
...
+where the repository manager needs to be proxied at a different base path you must change the default path by 
+editing a property value. In <<admin-base-url>> follow the steps to change or update the base URL if you want an 
+alternate server name.
+
+==== Example: Reverse Proxy on Restricted Ports
+
+*Scenario*: You need to expose the repository manager on restricted port 80. The repository manager should not be 
+run with the root user. Instead run your reverse proxy on the restricted port 80 and the repository manager on 
+the default port 8081. End users will access the repository manager using the virtual host URL 
+http://www.example.com/nexus instead of http://localhost:8081/nexus.
```
peterlynch: This scenario in Nexus 2 was intended to only cover changing the port and host name via a reverse proxy. The context path /nexus is going to need to also be explicitly changed in nexus 3 in order to get this example to work.

Suggest the example be changed to use the default Nexus context path of just / .

```
ServerName www.example.com
ServerAdmin admin@example.com
ProxyPass /nexus http://localhost:8081/
ProxyPassReverse / http://localhost:8081/
ErrorLog logs/www.example.com/nexus/error.log
CustomLog logs/www.example.com/nexus/access.log common
```

chapter-installation.asciidoc
```diff
@@ -373,6 +373,225 @@ manually start it after the configuration you can use
...
+			proxy_pass http://localhost:8081/nexus;
+			proxy_set_header Host $host;
+			proxy_set_header X-Real-IP $remote_addr;
+			proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+		}
+	}
+}
+----
+
+==== Example: Reverse Proxy Virtual Host at Base Path
```
peterlynch: Instead make this about serving at custom context path.

Example: Reverse Proxy Virtual Host at Custom Context Path

chapter-installation.asciidoc
```diff
@@ -373,6 +373,225 @@ manually start it after the configuration you can use
...
+
+*Scenario:* You need to expose the repository manager using a custom host name +repo.example.com+ on a restricted 
+port at a base path of slash (+/+).
```
peterlynch: base path of /nexus

chapter-installation.asciidoc
```diff
@@ -373,6 +373,225 @@ manually start it after the configuration you can use
...
+----
+ProxyRequests Off
+ProxyPreserveHost On
+
+<VirtualHost *:80>
+  ServerName repo.example.com
+  ServerAdmin admin@example.com
+  ProxyPass / http://localhost:8081/
+  ProxyPassReverse / http://localhost:8081
+  ErrorLog logs/repo.example.com/nexus/error.og
```
peterlynch:
```
ProxyPass /nexus http://localhost:8081/nexus
ProxyPassReverse /nexus http://localhost:8081/nexus
```

similar changes for nginx

chapter-installation.asciidoc
```diff
@@ -373,6 +373,225 @@ manually start it after the configuration you can use
...
+)
+----
+
+==== Example: Reverse Proxy SSL Termination at Base Path
+
+*Scenario:* Your organization has standardized a reverse proxy to handle SSL certificates and termination. The 
+reverse proxy virtual host will accept HTTPS requests on the standard port +443+ and serve content from the 
+repository manager running on the default non-restricted HTTP port +8081+ transparently to end users.
+
+Ensure your external host name (+repo.exmaple.com+) routes to your reverse proxy server and edit the webapp path
```
peterlynch: exmaple typo

chapter-installation.asciidoc
```diff
@@ -373,6 +373,225 @@ manually start it after the configuration you can use
...
+----
+
+==== Example: Reverse Proxy SSL Termination at Base Path
+
+*Scenario:* Your organization has standardized a reverse proxy to handle SSL certificates and termination. The 
+reverse proxy virtual host will accept HTTPS requests on the standard port +443+ and serve content from the 
+repository manager running on the default non-restricted HTTP port +8081+ transparently to end users.
+
+Ensure your external host name (+repo.exmaple.com+) routes to your reverse proxy server and edit the webapp path 
+to be slash (+/+).
```
peterlynch: the "edit the webapp path to be slash" should link out to the Nexus book section on how to change the context path of Nexus.

chapter-installation.asciidoc
```diff
@@ -510,6 +729,7 @@ context path and host.
 
 system.properties:: Configuration parameters used for the JVM and application start up.
 
+[[config-memory-jvm]]
 ==== Updating Memory Allocation and other JVM Paramaters
```
peterlynch: config-jvm-memory

memory is more specific than jvm therefor should be last

chapter-installation.asciidoc
```diff
@@ -524,7 +744,7 @@ file `$install-dir/bin/nexus.vmoptions`. Increased memory configuration can be s
 
 Other JVM parameters such as GC algorithm can be configured in the same location.
 
-
+[[config-http-port]]
```
peterlynch: would prefer to have changing the context in a separate anchored section. there are lots of things configurable in that file, but changing port and changing context path are the two most common and I would like to have anchors to each

chapter-installation.asciidoc
```diff
@@ -537,6 +757,7 @@ application-port=9081
 nexus-context-path=/components/
 ----
 
+[[config-data-storage]]
 ==== Configuring Data Storage Location
```
peterlynch: suggest config-data-dir as that is name I think we are going with.

how to configure the nexus tmp dir is worth isolating by itself. people often want to isolate temp data from application data.

mosabua:
Totally agree with the various feedback from @peterlynch . This needs to be better adapted to 3.x as compared a simple port of the 2.x docs section. I can review when its closer if desired.

dulanism:
@peterlynch & @mosabua Thanks for the feedback. I moved that status back to WIP as I'm currently revising.

chapter-installation.asciidoc
dulanism: Noted. Link to: https://books.sonatype.com/nexus-book/3.0/reference/install.html#_changing_http_port_and_context

dulanism:
@peterlynch Re: the line where you ask me to distinguish reverse proxy from the other forms of the term proxy -  That makes sense. But since there is a configuration dependency btw the "create base URL" & "running repoman behind a reverse proxy settings" section, why don't I move the "To configure and run...[reverse proxy] server" sentence to end, and link the sections. I moved this comment below as it won't disappear with the changes.

chapter-configuration.asciidoc
dulanism: updated

chapter-configuration.asciidoc
```diff
@@ -279,6 +266,20 @@ the 'Hosts to exclude from HTTP/HTTPS proxy' setting. You can add a hostname in
 <<fig-admin-system-http-proxy>> shows the 'HTTP Proxy' administration interface. The HTTPS configuration interface
 looks the same and is found below the HTTP configuration.
 
+[[admin-base-url]]
+==== Base URL Creation
+
+The `Base URL` is the endpoint at which users should access the user interface. This setting is used to host 
+the repository manager behind a reverse proxy server. The basic requirement is to make sure the context path 
+explicitly matches the base URL down to the root folder (e.g +/+, +/nexus+, or +/nexus3+). You do not need to set 
+the base URL if you are not proxying the repository manager, because it knows its own default URL.
```
peterlynch: This whole paragraph is still cloudy to me. I offer this suggestion at wording:

----

The Base URL is the address you want end users to use to navigate to the user interface. 

Nexus only uses this value to construct absolute URLs to your user interface inside of email notifications.

The most common reason why the address would be different for end users, is if you have a <<reverse-proxy>> in front of Nexus terminating http requests at a different address than where Nexus is running.

To set a Base URL:

1. Go to the 'System' section of the 'Administration' menu and select 'Capabilities'.
2. Search for an existing 'Base URL' capability and select it if it exists _or_ click the 'Create capability' button and choose 'Base URL' from the 'Select Capability Type' list.
3. Enter a new URL value. Press the 'Create capability' or 'Save' button to persist your changes.

PR #167, opened 2016-04-14
NEXUS-9826 PyPi Proxy Repo

DarthHater:
JIRA: https://issues.sonatype.org/browse/NEXUS-9826

Not ready for review... yet!


mosabua:
As discussed with @DarthHater this PR can NOT be merged into the nexus-3.0.x branch. After completion it will have to merged manually into a nexus-3.1.x branch that is yet to be created. cc @dulanism

mosabua:
But the PR can still be used as is for review and so on .. hence I am reopening after accidental closure..

book-nexus.asciidoc
```diff
@@ -29,6 +29,10 @@ include::chapter-npm.asciidoc[]
 
 include::chapter-bower.asciidoc[]
 
+////
```
mosabua: You can remove the comment so you can test locally ;-)

chapter-pypi.asciidoc
```diff
@@ -0,0 +1,46 @@
...
+=== Introduction
+
+{pro} and {oss} support the PyPi repository format for hosted and proxy repositories. This allows the repository
+manager to take advantage of the packages in the official PyPi repository and other public registries without
+incurring repeated downloads of packages.
+
+[[pypi-proxy]]
+=== Proxying PyPi Repositories
+
+You can set up a PyPi proxy repository to access a remote repository location, for example the official PyPi repository at https://pypi.python.org/pypi[https://pypi.python.org/pypi] that is configured as the default on PyPi.
```
joedragons: "that is configured as the default" reads weird...I think @DarthHater said he used Bower as a guideline, I wonder (didn't check) if that used Maven or Nuget as a guideline which have default repos.  Otherwise, unclear to me what you're trying to say here.

joedragons:
Saw "WIP" after I reviewed, sorry=)  But also wonder if we want to put "PyPI" everywhere (instead of PyPi) since per https://pypi.python.org/pypi it's Python Package Index and that's what they use.  That's what I put in the testguide.  I have no doubt there's all kinds of varients here=D

DarthHater:
Closing in favor of another that merges into 3.1.x

PR #170, opened 2016-04-22
Added updated license section with managing active lic users

dulanism:
JIRA: https://issues.sonatype.org/browse/NEXUS-9996
https://issues.sonatype.org/browse/NEXUS-9999

License upload and active user management

### DIFF
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -879,21 +879,20 @@ side resolution done by Maven each time.
 === License Management
 {inrmonly}
 
-IMPORTANT: You must be logged in as a user with sufficient privileges to manage licenses.
-
 A paid license is necessary to upgrade {oss} to {pro}, and to keep {pro} paid features operational. You must be 
 logged in as a user with sufficient privileges to manage licenses.
 
 [[admin-license-renew]]
 ==== Uploading a License
 
-The paid license is a special +.lic+ file that can be uploaded from the 'Licensing' section located in the 
-'Administration' menu. To install it, locate the file with the 'Select license...' button in the user interface. 
-Follow the steps below:
+The paid license is a special +.lic+ file that you upload to the 'Licensing' feature view, found in the 
+'Administration' menu. To install the license:
 
+* Upload the file from the 'Select license...' button.
 * Select the correct +.lic+ file in the file selection dialog, and press 'Open'.
 * Click the 'Install license' button.
 * Enter your password when the re-authentication dialog appears.
+* Click 'I agree' to the terms stated in the End User License Agreement.
 * Click 'Authenticate' to complete the upload.
 
 After the file is selected and uploaded, the 'License type' field will display the corresponding bundle 
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -900,11 +900,12 @@ you purchased.
 
 ==== Managing Active Users
 
-Users with sufficient privileges can generate a record of user access during a session in the repository manager. 
-Go to the 'License Users' feature view, nested below 'Licensing', to access the table. 
+Users with sufficient privileges can generate a record of users who had sessions within the last 7 days in the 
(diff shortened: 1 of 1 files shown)


chapter-configuration.asciidoc
```diff
@@ -879,31 +879,32 @@ side resolution done by Maven each time.
...
 * Click 'Authenticate' to complete the upload.
 
 After the file is selected and uploaded, the 'License type' field will display the corresponding bundle 
-purchased. Options include {iq} with {pro}, or {pro} alone.
+you purchased.
+
+==== Managing Active Users
 
-==== License Expiration
+Users with sufficient privileges can generate a record of user access during a session in the repository manager.
```
joedragons: "during a session" reads weird to me.  Maybe "who had sessions"?  But this may also be a good place to put the date limit (7 days).  So "...who had sessions within the last 7 days in the...".
This is assuming I'm reading this correctly, and you're not trying to say it's the users with the privileges can access this during their session, which doesn't seem necessary to state.

chapter-configuration.asciidoc
```diff
@@ -879,31 +879,32 @@ side resolution done by Maven each time.
...
 
 After the file is selected and uploaded, the 'License type' field will display the corresponding bundle 
-purchased. Options include {iq} with {pro}, or {pro} alone.
+you purchased.
+
+==== Managing Active Users
 
-==== License Expiration
+Users with sufficient privileges can generate a record of user access during a session in the repository manager. 
+Go to the 'License Users' feature view, nested below 'Licensing', to access the table.
```
joedragons: Minor: I feel like this usually says Administration menu but maybe we don't here becuase we stated it above.  I know there is a mixed camp of whether people will be reading these straight through vs going right to the section they need.  This would suffer (minorly) if people do the latter.

chapter-configuration.asciidoc
```diff
@@ -879,31 +879,32 @@ side resolution done by Maven each time.
...
+==== Managing Active Users
 
-==== License Expiration
+Users with sufficient privileges can generate a record of user access during a session in the repository manager. 
+Go to the 'License Users' feature view, nested below 'Licensing', to access the table. 
 
-When a {pro} license expires all functionality will be disabled in the user interface, except for the ability to 
-install a new license file. To avoid interruption of service be sure to upload a renewed license before the 
-existing license expires.
+The table displays the IP address ('IP'), last entry date ('Date'), user privilege ('User'), e.g. 'admin', and
```
joedragons: * Minor: Suggest "accessed" instead of "entry"
* Believe we should put "name" instead of "privilege" since privilege represents a portion of a role which is a portion of a user (name).  I'm also not convinced the "e.g." is useful to this end but feel less strongly about that.

chapter-configuration.asciidoc
```diff
@@ -879,31 +879,32 @@ side resolution done by Maven each time.
...
 
-==== License Expiration
+Users with sufficient privileges can generate a record of user access during a session in the repository manager. 
+Go to the 'License Users' feature view, nested below 'Licensing', to access the table. 
 
-When a {pro} license expires all functionality will be disabled in the user interface, except for the ability to 
-install a new license file. To avoid interruption of service be sure to upload a renewed license before the 
-existing license expires.
+The table displays the IP address ('IP'), last entry date ('Date'), user privilege ('User'), e.g. 'admin', and 
+client ('User agent'). To access the report click 'Download' to produce a CSV file of all users.
```
joedragons: Minor:
* "access" reads weird, since you access this report on the screen (before downloading).  Not sure another elegant word without using "download" (duh) tho.  Maybe "view the full report"?
* Suggest changing "all" to "listed".

dulanism: :) Only because I used the word "generate" at the beginning. Let's see what I can do.

dulanism:
@joedragons Lemme know if this is good to go. Cheers!

joedragons:
+1.  Thanks!

PR #172, opened 2016-04-25
PyPI configuration for proxy - extended

dulanism:
CI: http://bamboo.s/browse/BOOK-NEXUS135
JIRA https://issues.sonatype.org/browse/NEXUS-9826

### DIFF
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -7,11 +7,11 @@
 
 The Python Package Index, or PyPI, is a vast repository of open-source Python packages supplied by the worldwide 
 community of Python developers. The official PyPI is available for searches at 
-https://pypi.python.org/pypi[https://pypi.python.org/pypi]. Bu
+https://pypi.python.org/pypi[https://pypi.python.org/pypi].
 
 {pro} and {oss} support the Python Package Index (PyPI) repository format for proxy repositories. This 
 allows the repository manager to take advantage of packages in the Python repository and other public registries 
-without incurring repeated downloads of packages. The PyPI website, is maintained by 
+without incurring repeated downloads of packages. The PyPI website, is maintained by the
 https://www.python.org/psf/[Python Software Foundation].
 
 ////
@@ -21,9 +21,9 @@ manager supports PyPI packages as a repository group (as in... the repository gr
 contents of multiple repositories in one convenient URL)  
 ////
 
-Administrators who use the repository manager will download and run libraries to retrieve, facilitate, and manage 
-Python projects. By design, Python installs to the root directory with the version number embedded so you can 
-maintain multiple versions of Python without conflicts.
+Repository manager user will download and run libraries to retrieve, facilitate, and manage Python projects. By 
+design, Python installs to the root directory with the version number embedded so you can maintain multiple 
+versions of Python without conflicts.
 
 [[pypi-proxy]]
 === Proxying PyPI Repositories
@@ -40,11 +40,12 @@ detail. Minimal configuration steps are:
 * Select a 'Blob store' for 'Storage'
 
 The repository manager can access Python packages and tools from the index. The proxy repository for PyPI 
-packages provides a nearby cache of files available on the index. This allows the local network client to access 
+packages provides a cache of files available on the index. This allows the local network client to access 
 components from the Python Package Index more reliably.
 
 The proxy configuration for a PyPI proxy repository includes a configuration URL to access the index. 
-Adminstrators will be able to run a search against a remote repository, as mentioned in <<pypi-browse-search>>.
+Users will be able to run a browse and search assets against a remote repository, as mentioned in 
+<<pypi-browse-search>>.
 
 ////
 [[pypi-hosted]]
@@ -63,14 +64,15 @@ TBD
 [[pypi-installation]]
 === Installing PyPI Client Tools
 
-The latest versions of such Linux distributions as CentOS and Ubuntu are already packaged with Python 2.7 and 
-pip[https://pip.pypa.io/en/stable/], a tool for installing and managing Python packages from the index. For Mac 
-OS X and Microsoft Windows, download and install the up-to-date version of Python from 
-https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the Pip installer from 
+The latest versions of such Linux distributions as CentOS and Ubuntu come packaged with Python 2.7 and 
+https://pip.pypa.io/en/stable/[pip], a tool for installing and managing Python packages from the index. For Mac 
+OS X and Microsoft Windows, download and install a Python version compatible with the repository manager from 
+https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the pip installer from 
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
 
-NOTE: Supported Python versions for the repository manager include 2.6, 2.7, 3.4, 3.5. For pip, the repository 
-manager supports version 7 and 8.
+NOTE: {pro} and {oss} support specific versions of Python, pip, and setuptools. For Python the repository manager 
+supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Versions 19 and 20 of setuptools, used to 
+build and distribute Python dependencies, are compatible with the repository manager. 
 
 ////
 Section for the hosted docs - re: since both twine and subsequently pip to install twine, would be hosted
@@ -84,12 +86,16 @@ sudo pip install twine
 [[pypi-configuration]]
 === Configuring PyPI Client Tools
 
-Once you have installed all necessary packages from the Python Package Index, you'll need create and configure 
-your client-side files necessary to proxy index pages for PyPI. Depending on your Python configuration you can 
-use +pip.conf+ or +setup.cfg+ to set all command line option defaults for your proxy repository,
+NOTE: Depending on your preference for either https://pypi.python.org/pypi/setuptools[setuptools] or pip your 
+proxy configuration may vary.
 
-In the example below, a +pip.conf+ is created. The +index+ specifies the index for a proxied PyPI repository 
-setup within the repository manager. +index-url+ is the base URL designated for proxy repository's index pages.
+Once you have installed all necessary client tools from the Python Package Index, create and configure your 
+client-side files necessary to proxy index pages for PyPI. Depending on your Python configuration you can use 
++pip.conf+ or +setup.cfg+ to set all command line option defaults for your proxy repository.
+
+In the example below, a +pip.conf+ is shown. The +index+ specifies the index for a proxied PyPI repository 
+setup within the repository manager. +index-url+ is the base URL designated for the cached index pages within the 
+proxy repository.
 
 ----
 [global]
@@ -114,7 +120,7 @@ password = admin123
 ////
 
 Alternately you can create a +setup.cfg+. Like the +pip.conf+ option, the +index-url+ is the tag created to 
-specify the base URL for PyPI proxy index:
+specify the base URL for the PyPI proxy index.
 
 ----
 [easy_install]
@@ -170,11 +176,11 @@ cert = nexus.pem
 === Browsing PyPI Repositories and Searching Packages
 
 You can browse PyPI repositories in the user interface inspecting the components and assets and their details, as
-described in <<search-components>>.
+described in <<browse-browse>>.
 
-Searching for PyPI packages can be performed in the user interface, too. It finds all packages that are currently
-stored in the repository manager that have been proxied from an upstream repository and cached in the repository 
-manager.
+Searching for PyPI packages can be performed in the user interface as described in <<search-components>>. It 
+finds all packages that are currently stored in the repository manager that have been proxied from an upstream 
+repository and cached in the repository manager.
 
 ////
 Second sentence edit for hosted docs
@@ -182,8 +188,8 @@ It finds all packages that are currently stored in the repository manager, eithe
 to a hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
 ////
 
-From the command line you can search available PyPI packages defined in your configuration file (+setup.cfg+ or 
-+pip.conf+). To search, run:
+From the command line you can search available PyPI packages defined in your configuration. This method is 
+limited to pip (+pip.conf+). To search, run:
 
 ----
 pip search example-package
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -185,6 +185,7 @@ python2.7 setup.py bdist_egg
 ////
 [[pypi-upload-dist]]
 === Uploading Distributions
+TBD
 
 ----
 twine upload dist/*
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -163,18 +163,24 @@ python setup.py sdist
 
 *Build a wheel.*
 
+Wheel is a built-package format that automatically recompiles your applications during every install. It uses the 
+extension +bdist_wheel+, which detects the Python installation so that is usable with any Python instance of the 
+same version.
+
 In this example run the following command for a Python 2.7 environment:
 
 ----
 python2.7 setup.py bdist_wheel
 ----
 
-The +bdist_wheel+ will detect the Python installation so that is usable with any Python instance of the same 
-version.
-
 *Build an egg.*
 
-Run the following command:
+Like Wheel, Egg is a cross-platform distribution format for pure Python packages. It uses the extension 
++bdist_egg+ to build individual Eggs. Run the following command for a Python 2.7 environment:
+
+----
+python2.7 setup.py bdist_egg
+----
 
 ////
 [[pypi-upload-dist]]
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -5,9 +5,9 @@
 [[pypi-introduction]]
 === Introduction
 
-The Python Package Index, or PyPI, is a repository of open-source Python packages. It contains a vast catalog of 
-libraries supplied by the worldwide community of Python developers. The official PyPI is available for searches 
-at https://pypi.python.org/pypi[https://pypi.python.org/pypi].
+The Python Package Index, or PyPI, is a vast repository of open-source Python packages supplied by the worldwide 
+community of Python developers. The official PyPI is available for searches at 
+https://pypi.python.org/pypi[https://pypi.python.org/pypi].
 
 {pro} and {oss} support the Python Package Index (PyPI) repository format for proxy repositories. This 
 allows the repository manager to take advantage of packages in the Python repository and other public registries 
@@ -43,10 +43,13 @@ detail. Minimal configuration steps are:
 * Define URL for 'Remote storage' e.g. https://pypi.python.org/pypi[https://pypi.python.org/pypi]
 * Select a 'Blob store' for 'Storage'
 
+/////
+section should distinguish search for ui vs command line
 The Proxy configuration for a PyPI proxy repository includes a configuration URL to access the index. The index 
 is used for requests related to searches. A user with sufficient privileges will be able to run a search, from 
 the user interface or command line, against a remote repository (e.g. https://pypi.python.org/pypi ), as 
 referenced in <<pypi-browse-search>>.
+////
 
 ////
 [[pypi-hosted]]
@@ -63,13 +66,18 @@ TBD
 ////
 
 [[pypi-installation]]
-=== Installing Python and PyPI Client Tools
+=== Installing and PyPI Client Tools
 
 The latest versions of such Linux distributions as CentOS and Ubuntu, as well as El Capitan for Mac OS X, are 
 already packaged with Python 2.7 and Pip, a tool for installing and managing Python packages from the index. 
 However, if you run Microsoft Windows, download and install the up-to-date version of Python from
 https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the Pip installer from 
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
+
+////
+Explain common tools...
+Common tools you may use to create, update, and edit configuration files
+////
 
 Supported Python versions for the repository manager include 2.6, 2.7, 3.4, 3.5. From the command line verify the 
 version on your system.
@@ -78,50 +86,36 @@ version on your system.
 python --version
 ----
 
-The repository manager supports version 7 and 8 of the PIP installer. Run the following to check the PIP version:
+The repository manager supports version 7 and 8 of the pip installer. Run the following to check the pip version:
 
 ----
 pip --version
 ----
 
-If you do not have PIP, run the command to install version 7 or 8. For example, install version 8:
+If you do not have pip, run the command to install version 7 or 8. For example, install version 8:
 
 ----
 pip install pip==8
 ----
 
-Use PIP to install +virtualenv+, a virtual enviroment tool for Python applications. Rather than installing PyPI 
-packages globally, use +virtualenv+ to sandbox a specific running Python environment. +virtualenv+ will access a 
-selected against different versions of the Python interpreter without modifying your main Python 
-installation. Run the command as follows:
-
-----
-sudo pip install virtualenv
-----
-
-Next install Twine. Twine is a utility that provides secure authentication to PyPI over HTTPS. It is an 
-alternative to `python setup.py upload` which is available to older version of Python. Twine's only function is 
-to support uploading distributions. To install with PIP, run the command as follows:
+////
+Section for the hosted docs - re: since both twine and subsequently pip to install twine, would be hosted
+Next install Twine. Twine is a utility that provides secure authentication to PyPI over HTTPS. Twine's only 
+function is to upload distributions.
 
 ----
 sudo pip install twine
 ----
-
-Now activate a +virtualenv+. Change the environment name and version number as required. In the example below, an 
-environment for Python 2.6 is created, then activated:
-
-----
-virtualenv py26 --python=python2.6
-cd py26
-source ./bin/activate
-----
+////
 
 [[pypi-configuration]]
 === Configuring PyPI Client Tools
 
 Once you have installed all necessary packages from the Python Package Index and active your +virtualenv+, create 
 and configure +pip.conf+. +pip.conf+ is a configuration file that allows you to set all command line option
-defaults in a standard ini style. In the example below, reference the remote index package to be proxied:
+defaults in a standard INI style.  In the example file below, the +index+ specifies the index for a proxied 
+PyPI repository setup within the repository manager. +index-url+ is the base URL designated for proxy 
+repository's index pages.
 
 ----
 [global]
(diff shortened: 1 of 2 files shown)


chapter-pypi.asciidoc
```diff
@@ -43,7 +43,10 @@ detail. Minimal configuration steps are:
 * Define URL for 'Remote storage' e.g. https://pypi.python.org/pypi[https://pypi.python.org/pypi]
 * Select a 'Blob store' for 'Storage'
 
-A user with sufficient privileges will be able to run pip search against a proxied repository.
+The Proxy configuration for a PyPI proxy repository includes a configuration URL to access the index. The index 
+is used for requests related to searches. A user with sufficient privileges will be able to run a search, from 
+the user interface or command line, against a remote repository (e.g. https://pypi.python.org/pypi ), as
```
joedragons: I am a little concerned this as written will lead people to believe they will see the same results in the UI as via the CLI which is not (currently) true.  UI shows stuff Nexus has, CLI shows stuff the index has (currently).  I said currently twice, but as far as I know there's no plans to change this.  Other formats we have the UI works this way as well.

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
+----
 
+Now activate a +virtualenv+. Change the environment name and version number as required. In the example below, an 
+environment for Python 2.6 is created, then activated:
+
+----
+virtualenv py26 --python=python2.6
+cd py26
+source ./bin/activate
+----
```
joedragons: Virtualenv should be for testing purposes only.  They should use whatever python (and tools) they have.

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
 sudo pip install virtualenv
 ----
 
-////
-* use pip to install setup tools
-////
+Next install Twine. Twine is a utility that provides secure authentication to PyPI over HTTPS. It is an 
+alternative to `python setup.py upload` which is available to older version of Python. Twine's only function is 
+to support uploading distributions. To install with PIP, run the command as follows:
```
joedragons: Minor: I've never seen PIP capitolized before (ref https://pip.pypa.io/en/stable/) but marked minor as I trust you here.  In case this was for emphasis and not case sensitivity recommend some kind of markup instead.  (++?)

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
+----
 
 [[pypi-configuration]]
 === Configuring PyPI Client Tools
 
-Once you have installed all necessary packages from the Python Package Index, create and configure +pip.conf+. 
-+pip.conf+ is a configuration file that allows you to set all command line option defaults in a standard ini 
-style. In the example below, add the proxy package index url:
+Once you have installed all necessary packages from the Python Package Index and active your +virtualenv+, create 
+and configure +pip.conf+. +pip.conf+ is a configuration file that allows you to set all command line option
```
joedragons: Because of my previous note, I think the other writeup (with no ref to virtualenv) is preferred/more accurate for these two lines.

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
+defaults in a standard ini style. In the example below, reference the remote index package to be proxied:
```
joedragons: Minor: I am not sure the word 'package' is correct as now written.  I might be confused, might be worth checking with @fjmilens3 .

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
-* Create and active your +virtualenv+
-* use pip to install easy_install 
-////
+Next bootstrap your target environment with +setuptools+ using PIP. +easy_install+ is a module bundled with 
++setuptools+. +easy_install+ that gives you a quick way to install packages remotely. In the example, version 19 
+is used.
+
+----
+pip install setuptools==19
+----
```
joedragons: Again, not sure why we're telling them what tools to use.  Even if we're telling them what tools to use, we shouldn't tell them to use this=)  I think this can be struck and we can skip to configuration.

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
+
+Now, create a +setup.cfg+ for for +easy_install+. Add the following lines, then save:
```
joedragons: Similar to the above, suggest rewriting slightly to infer use this if they're using setuptools.  Otherwise this shouldn't be needed.

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
-nano setup.cfg
-for hosted 
-.pypirc
+for hosted add .pypirc configuration
 ////
 
-[[pypi-ssl]]
-==== SSL Usage for PyPI Client Tools
+[[pypi-create-dist]]
+=== Packaging and Distributing Projects
```
joedragons: I feel like Bad Santa but I think this entire section is not needed.  They should know how to create wheels/eggs and I don't think this guide is intended to tell them how (any more than we tell people how to create pom.xmls or dockerfiles).

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
+
+Update your +pip.conf+ as follows:
+
+----
+[global]
+index = https://localhost:8443/repository/pypi-proxy/pypi
+index-url = https://localhost:8443/repository/pypi-proxy/simple
+cert = nexus.pem
+----
+////
```
joedragons: Good stuff for proxy.  However untestable at the moment=(

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
-e.g. From the command line you can search available packages with pip search. Run the command pip search [package name]. For example, if you want to install 
+Second sentence edit for hosted docs
+It finds all packages that are currently stored in the repository manager, either because they have been pushed 
+to a hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
 ////
 
-Searching for PyPI packages can be performed in the user interface, too. It finds all packages that are currently
-stored in the repository manager, either because they have been pushed to a hosted repository or they have been
-proxied from an upstream repository and cached in the repository manager.
+From the command line you can search available PyPI packages. To search for a package, run:
```
joedragons: Minor: I am not sure if it's worth a spelling out that this searches whatever index they have in setup.cfg.  If they don't define that, they'll be searching the pypi site.  If they have a group, it'll search their hosted/proxy.  I'm explaining btw, not suggesting that be the note.
Something as simple as "From the command line, you can search PyPI packages from the index defined in your setup.cfg."  Just a thought, I don't feel strongly so made minor.
But if they have proxy in their setup.cfg and want to search group...that might be explained by that small edition.

chapter-pypi.asciidoc
```diff
@@ -96,58 +99,163 @@ installation. Run the command as follows:
...
+[[pypi-deactivate]]
+==== Deactivating the Environment
+
+To switch back to your default environment, run the following:
+
+----
+deactivate
+----
+
+You should see the command prompt change, and your original versions of Python and PIP should be available.
```
joedragons: The deactivate section I also think can be cut.  It shuts down virtualenv which is for testing purposes only.  Actual users should know how to operate their own envs without our descriptions (if they use virtualenv already they should know to deactivate).

joedragons:
Also made some comments to @dulanism about the stuff already there offthread since it was hard to comment to grey areas in GitHub as usual.  I reviewed commented out areas as well as non, tho in future I prefer distinct PRs because it's easier to build and read along than flipping back and forth to make sure I don't miss anything.

chapter-pypi.asciidoc
fjmilens3: I think we do need to be clear about that distinction, unless we think that they just understand it because of how proxies work in NX3.

chapter-pypi.asciidoc
```diff
@@ -23,7 +23,7 @@ contents of multiple repositories in one convenient URL)
 
 To reduce time and bandwidth the repository manager can access Python packages and tools from the index. The 
 proxy repository for PyPI packages provides a nearby cache of files available on the index. This allows the local 
-network client to access files from the Python Package Index more reliably.
+network client to access components from the Python Package Index more reliably.
```
fjmilens3: I'm not sure what word to use here, honestly. PyPI doesn't have components, we create them based on what we pull down, but PyPI itself basically just has a name and a bunch of assets (among other things, the versioning happens on the assets, not the component). I think it's at least reasonably accurate and consistent with Nexus, and I don't think it would confuse a Python dev *too* much. @DarthHater, what do you think?

chapter-pypi.asciidoc
fjmilens3: @joedragons is correct here. This configuration doesn't proxy a package, it basically just tells the client where to try and pull packages down from. I'm also unsure if the "to be proxied" makes sense, as you're specifying the proxy location when you set up the repo in Nexus. What the user is doing here is setting up the base url to the Nexus proxy's index pages.

(There's a bit more of a wrinkle in that if you go to something like the PyPI simple index, you get a huge page of links to every possible project. We don't proxy that as we usually don't make that kind of stuff browsable unless the client requires it, and in the case of easy_install and pip it never needs to hit that main index. But it does provide what's essentially the base url that the client appends the name onto, to know where to find the particular index for a particular project. I'm worried I'm probably just making this more confusing rather than simpler by this explanation though.)

chapter-pypi.asciidoc
fjmilens3: Agreed.

chapter-pypi.asciidoc
fjmilens3: I think we should explain how to get the common tools to talk to Nexus by changing/updating/creating config files, I wouldn't go much into detail past that. We have to assume that the people using this understand their tools enough to know why they're doing what they're doing in the first place, and what they need to be doing; that's not our core competency in terms of our documentation.

mosabua: Agreed... only document the NXRM specific aspects.. for the rest refer to the actual documentation of the tool itself.

chapter-pypi.asciidoc
mosabua: Agreed.. the only thing needed to be covered is how to config the deployment to NXRM and any credentials.. and if there is anything specific about NXRM vs other repos

chapter-pypi.asciidoc
mosabua: yep... toss it ;-)

chapter-pypi.asciidoc
joedragons: (disclaimer: I'm not @DarthHater ): I think what is written now is consistent with our other formats.  I believe we call them whatever the devs call them when talking about not in Nexus (in this case packages) and call them components when talking about in Nexus.  While not stated, I always figured this was a subconscious way of helping people learn our terminology=)

chapter-pypi.asciidoc
joedragons: I didn't even think about that.  So, perhaps rewrite (suggestion) "In the example below, we add an index to a PyPI proxy repository.  In turn, this means search and installs will run through it."

dulanism: Rev (last sentences): In the example file below, the _index_ line specifies the proxy location when you set up the repository. _index-url_ is the base URL designated for proxy repository's index pages.

joedragons: The first sentence is true with /pypi on the end (see attached in comparison).  I suspect that it might be "the _index_ line specifies the index for a proxied PyPI repository setup within Nexus."  I am not 100% sure tho.

screen shot 2016-04-26 at 9 47 32 am

dulanism: Tentatively revising with your suggestion, JT.

chapter-pypi.asciidoc
```diff
@@ -1,13 +1,13 @@
...
 
 [[pypi-introduction]]
 === Introduction
 
-The Python Package Index, or PyPI, is a repository of open-source Python packages. It contains a vast catalog of 
-libraries supplied by the worldwide community of Python developers. The official PyPI is available for searches 
-at https://pypi.python.org/pypi[https://pypi.python.org/pypi].
+The Python Package Index, or PyPI, is a vast repository of open-source Python packages supplied by the worldwide 
+community of Python developers. The official PyPI is available for searches at 
+https://pypi.python.org/pypi[https://pypi.python.org/pypi]. Bu
```
joedragons: Got an extra "Bu" here it appears.

chapter-pypi.asciidoc
```diff
@@ -43,7 +39,12 @@ detail. Minimal configuration steps are:
 * Define URL for 'Remote storage' e.g. https://pypi.python.org/pypi[https://pypi.python.org/pypi]
 * Select a 'Blob store' for 'Storage'
 
-A user with sufficient privileges will be able to run pip search against a proxied repository.
+The repository manager can access Python packages and tools from the index. The proxy repository for PyPI 
+packages provides a nearby cache of files available on the index. This allows the local network client to access
```
joedragons: +0 on word "nearby".  Distance is nebulous(sp?) on the internet=)  Like my spelling:P

chapter-pypi.asciidoc
dulanism: Whoops. I'll kick that out for the next build.

chapter-pypi.asciidoc
```diff
@@ -43,7 +39,12 @@ detail. Minimal configuration steps are:
 * Define URL for 'Remote storage' e.g. https://pypi.python.org/pypi[https://pypi.python.org/pypi]
 * Select a 'Blob store' for 'Storage'
 
-A user with sufficient privileges will be able to run pip search against a proxied repository.
+The repository manager can access Python packages and tools from the index. The proxy repository for PyPI 
+packages provides a nearby cache of files available on the index. This allows the local network client to access 
+components from the Python Package Index more reliably.
+
+The proxy configuration for a PyPI proxy repository includes a configuration URL to access the index. 
+Adminstrators will be able to run a search against a remote repository, as mentioned in <<pypi-browse-search>>.
```
joedragons: +0 on this last sentence; it seems sort of out of place (talking about setting up proxy then suddenly are like "btw search").  If this remains, I am -1 to saying "Administrators".  Search can be done by default anonymously.

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
 
 [[pypi-installation]]
-=== Installing Python and PyPI Client Tools
+=== Installing PyPI Client Tools
 
-The latest versions of such Linux distributions as CentOS and Ubuntu, as well as El Capitan for Mac OS X, are 
-already packaged with Python 2.7 and Pip, a tool for installing and managing Python packages from the index. 
-However, if you run Microsoft Windows, download and install the up-to-date version of Python from
+The latest versions of such Linux distributions as CentOS and Ubuntu are already packaged with Python 2.7 and 
+pip[https://pip.pypa.io/en/stable/], a tool for installing and managing Python packages from the index. For Mac
```
joedragons: This link looks weird in the pdf (see attached).  I doubt you want it to show the label and the url.  I looked on other links to see if I could see anything glaring but the label matches the url in all other things in this PR.  I didn't look outside th PR at this time.  I also didn't build the site (just reviewing the pdf) where it could be fine.

screen shot 2016-04-26 at 5 49 19 pm

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
-=== Installing Python and PyPI Client Tools
+=== Installing PyPI Client Tools
 
-The latest versions of such Linux distributions as CentOS and Ubuntu, as well as El Capitan for Mac OS X, are 
-already packaged with Python 2.7 and Pip, a tool for installing and managing Python packages from the index. 
-However, if you run Microsoft Windows, download and install the up-to-date version of Python from
+The latest versions of such Linux distributions as CentOS and Ubuntu are already packaged with Python 2.7 and 
+pip[https://pip.pypa.io/en/stable/], a tool for installing and managing Python packages from the index. For Mac 
+OS X and Microsoft Windows, download and install the up-to-date version of Python from 
 https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the Pip installer from
```
joedragons: Is this a leftover from deciding pip is "pip" (ref "Pip")?

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
-However, if you run Microsoft Windows, download and install the up-to-date version of Python from
+The latest versions of such Linux distributions as CentOS and Ubuntu are already packaged with Python 2.7 and 
+pip[https://pip.pypa.io/en/stable/], a tool for installing and managing Python packages from the index. For Mac 
+OS X and Microsoft Windows, download and install the up-to-date version of Python from 
 https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the Pip installer from 
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
 
-Supported Python versions for the repository manager include 2.6, 2.7, 3.4, 3.5. From the command line verify the 
-version on your system.
+NOTE: Supported Python versions for the repository manager include 2.6, 2.7, 3.4, 3.5. For pip, the repository
```
joedragons: BTW, I know we decided this in ticket and in planning, but it sounds weird to me in text that we say we support 2.7 but not 3.0, 3.1, 3.2 and 3.3.  If others are agreeable I'd prefer we say 2.6, 2.7 and Python 3.  cc/ @fjmilens3 @DarthHater

chapter-pypi.asciidoc
dulanism: updated for next build

chapter-pypi.asciidoc
DarthHater: I'm actually fine with being explicit here in versions. Wouldn't want someone with 3.1 or something with an old version of pip or some odd bug getting an odd experience and expecting it to work.

dulanism: I'm okay with "...2.6, 2.7 and Python 3."

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
+The latest versions of such Linux distributions as CentOS and Ubuntu are already packaged with Python 2.7 and 
+pip[https://pip.pypa.io/en/stable/], a tool for installing and managing Python packages from the index. For Mac 
+OS X and Microsoft Windows, download and install the up-to-date version of Python from 
 https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the Pip installer from 
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
 
-Supported Python versions for the repository manager include 2.6, 2.7, 3.4, 3.5. From the command line verify the 
-version on your system.
+NOTE: Supported Python versions for the repository manager include 2.6, 2.7, 3.4, 3.5. For pip, the repository 
+manager supports version 7 and 8.
```
joedragons: Did we intentionally leave out what versions of setuptools we support?

dulanism: Yeah, since last time I looked at the google doc there was a question mark around 19 & 20.

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
-python --version
+sudo pip install twine
 ----
+////
 
-The repository manager supports version 7 and 8 of the PIP installer. Run the following to check the PIP version:
+[[pypi-configuration]]
+=== Configuring PyPI Client Tools
+
+Once you have installed all necessary packages from the Python Package Index, you'll need create and configure
```
joedragons: Minor: Suggest replacing "installed all necessary packages from the Python Package Index" with "the necessary client tools".  I read quickly and I thought I skipped a section and we were already past installing.  Clear on reread but lean to simpler.

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
 
-The repository manager supports version 7 and 8 of the PIP installer. Run the following to check the PIP version:
+[[pypi-configuration]]
+=== Configuring PyPI Client Tools
+
+Once you have installed all necessary packages from the Python Package Index, you'll need create and configure 
+your client-side files necessary to proxy index pages for PyPI. Depending on your Python configuration you can 
+use +pip.conf+ or +setup.cfg+ to set all command line option defaults for your proxy repository,
+
+In the example below, a +pip.conf+ is created. The +index+ specifies the index for a proxied PyPI repository
```
joedragons: Minor: Suggest changing "created" to "shown" or "displayed".  In connection with the word example, I was expecting a walkthrough type when you said "created", tho I know what you mean.

chapter-pypi.asciidoc
dulanism: That should be [url]pip instead. Fixed.

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
-selected against different versions of the Python interpreter without modifying your main Python 
-installation. Run the command as follows:
+Alternately you can create a +setup.cfg+. Like the +pip.conf+ option, the +index-url+ is the tag created to 
+specify the base URL for PyPI proxy index:
 
 ----
-sudo pip install virtualenv
+[easy_install]
+index-url = http://localhost:8081/repository/pypi-proxy/simple
 ----
```
joedragons: So (as you probably know) these two are dependant upon what tools you use to proxy.  I would say that, rather than assume they know, tho they may.

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
+Second sentence edit for hosted docs
+It finds all packages that are currently stored in the repository manager, either because they have been pushed 
+to a hosted repository or they have been proxied from an upstream repository and cached in the repository manager.
 ////
 
-Searching for PyPI packages can be performed in the user interface, too. It finds all packages that are currently
-stored in the repository manager, either because they have been pushed to a hosted repository or they have been
-proxied from an upstream repository and cached in the repository manager.
+From the command line you can search available PyPI packages defined in your configuration file (+setup.cfg+ or 
++pip.conf+). To search, run:
```
joedragons: pip (cli) search is limited to pip (so limited to pip.conf).  I don't think there's an setuptools search however @fjmilens3 may know of one.  I could research if you think worth as well.

chapter-pypi.asciidoc
```diff
@@ -60,94 +61,130 @@ TBD
...
 
 [[pypi-browse-search]]
 === Browsing PyPI Repositories and Searching Packages
 
 You can browse PyPI repositories in the user interface inspecting the components and assets and their details, as
 described in <<search-components>>.
 
+Searching for PyPI packages can be performed in the user interface, too. It finds all packages that are currently
+stored in the repository manager that have been proxied from an upstream repository and cached in the repository 
+manager.
```
joedragons: Minor: Not sure if intentional or not but the link under browse links to search and there's no link under search.  I compared to docker which links browse to browse (no mention of search) and npm which links browse to browse and search to search distinctly.  Ultimately you get there, but I feel this is (minorly) inconsistent.

chapter-pypi.asciidoc
joedragons: I'm pretty sure this was agreed upon (in planning).  I can't edit the google doc, so that's a Q for @fjmilens3 =)  I tested both tho, if it helps.

chapter-pypi.asciidoc
dulanism: [x] updated

chapter-pypi.asciidoc
dulanism: - [x] updated

chapter-pypi.asciidoc
dulanism: - [x] updated (added a note at the beginning of the section)

chapter-pypi.asciidoc
```diff
@@ -1,17 +1,17 @@
...
-at https://pypi.python.org/pypi[https://pypi.python.org/pypi].
+The Python Package Index, or PyPI, is a vast repository of open-source Python packages supplied by the worldwide 
+community of Python developers. The official PyPI is available for searches at 
+https://pypi.python.org/pypi[https://pypi.python.org/pypi].
 
 {pro} and {oss} support the Python Package Index (PyPI) repository format for proxy repositories. This 
 allows the repository manager to take advantage of packages in the Python repository and other public registries 
-without incurring repeated downloads of packages. The PyPI website, is maintained by 
+without incurring repeated downloads of packages. The PyPI website, is maintained by the
 https://www.python.org/psf/[Python Software Foundation].
```
joedragons: Minor: While true, the last sentence reads a little out of place.  Wonder if it'd be better inserted into the first paragraph.

chapter-pypi.asciidoc
```diff
@@ -21,13 +21,9 @@ manager supports PyPI packages as a repository group (as in... the repository gr
...
 ////
 
-To reduce time and bandwidth the repository manager can access Python packages and tools from the index. The 
-proxy repository for PyPI packages provides a nearby cache of files available on the index. This allows the local 
-network client to access files from the Python Package Index more reliably.
-
-Administrators who use the repository manager will download and run libraries to retrieve, facilitate, and manage 
-Python projects. By design, Python installs to the root directory with the version number embedded so you can 
-maintain multiple versions of Python without conflicts.
+Repository manager user will download and run libraries to retrieve, facilitate, and manage Python projects. By
```
joedragons: **A/the** repoistory manager user...?

chapter-pypi.asciidoc
```diff
@@ -21,13 +21,9 @@ manager supports PyPI packages as a repository group (as in... the repository gr
...
-To reduce time and bandwidth the repository manager can access Python packages and tools from the index. The 
-proxy repository for PyPI packages provides a nearby cache of files available on the index. This allows the local 
-network client to access files from the Python Package Index more reliably.
-
-Administrators who use the repository manager will download and run libraries to retrieve, facilitate, and manage 
-Python projects. By design, Python installs to the root directory with the version number embedded so you can 
-maintain multiple versions of Python without conflicts.
+Repository manager user will download and run libraries to retrieve, facilitate, and manage Python projects. By 
+design, Python installs to the root directory with the version number embedded so you can maintain multiple 
+versions of Python without conflicts.
```
joedragons: +0 to this last sentence.  While true, it doesn't so much impact Nexus implementation of PyPI.

chapter-pypi.asciidoc
```diff
@@ -60,94 +62,135 @@ TBD
...
+[[pypi-configuration]]
+=== Configuring PyPI Client Tools
+
+NOTE: Depending on your preference for either https://pypi.python.org/pypi/setuptools[setuptools] or pip your 
+proxy configuration may vary.
 
-The repository manager supports version 7 and 8 of the PIP installer. Run the following to check the PIP version:
+Once you have installed all necessary client tools from the Python Package Index, create and configure your 
+client-side files necessary to proxy index pages for PyPI. Depending on your Python configuration you can use 
++pip.conf+ or +setup.cfg+ to set all command line option defaults for your proxy repository.
```
joedragons: Suggest "set all command line option defaults for your proxy repository." be "have all commands run against your proxy repository." or something like that.  Key words that cause me pause "option" and "for".

joedragons:
+1, all comments minor.  Thanks!

PR #173, opened 2016-04-28
NEXUS-9825 Pypi hosted configuration

dulanism:
https://issues.sonatype.org/browse/NEXUS-9825
http://bamboo.s/browse/BOOK-NEXUS136

Will also include group https://issues.sonatype.org/browse/NEXUS-9827

### DIFF
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -49,14 +49,12 @@ be able to browse and search assets against a remote repository, as mentioned in
 
 Creating a PyPI hosted repository allows you to upload packages in the repository manager. The hosted 
 repository acts as an authoritative location for these components. This effectively creates an asset that becomes 
-a pointer to an external URL (such as a Git repository). You will create a custom file called +.pypirc+ to manage 
-credentials from the hosted repository, explained in <<pypi-upload>>.
+a pointer to an external URL (such as a Git repository).
 
 Minimal configuration steps are:
 
-* Define 'Name' - e.g. +pypi-all+
+* Define 'Name' - e.g. +pypi-internal+
 * Select 'Blob store' for 'Storage'
-* Choose 'Deployment policy' for the 'Hosted' repository
 
 ////
 Note: a ticket filed - NXBOOK-947 : Consistent naming of repos was created from a community member that might 
@@ -93,7 +91,7 @@ function is to upload distributions. If the user uses pip then here's a sample c
 === Configuring PyPI Client Tools
 
 NOTE: Depending on your preference for either https://pypi.python.org/pypi/setuptools[setuptools], 
-https://pypi.python.org/pypi/twine[twine], https://docs.python.org/2.7/library/distutils.html[distutils], or pip 
+https://pypi.python.org/pypi/twine[twine], https://docs.python.org/2.7/library/distutils.html[distutils], and pip 
 your proxy and hosted configuration may vary.
 
 Once you have installed all necessary client tools from the Python Package Index, create and configure your 
@@ -112,7 +110,7 @@ repository that links to the source code repository". Leaving it here to confirm
 ////
 
 In the +.pypirc+ specify the URL for the target hosted repository in the +repository+ value, along with the 
-+username+ and +password+ to access the repository manager. The +.pypirc file contains distutils, a default 
++username+ and +password+ to access the repository manager. The +.pypirc+ file contains distutils, a default 
 server used by PyPI that provides upload commands that stores assets and authentication information.
 
 ----
@@ -196,9 +194,10 @@ pip search example-package
 
 NOTE: The steps to upload a PyPI package will vary if your system is configured with setuptools or twine.
 
-After you configure your +.pypirc+ you can upload packages from the index to the repository manager. 
-In the example below, using twine, tells you what repository configuration file to use when uploading; the +-r+ 
-flag is used to find the +nexus+ entry in your +.pypirc+.
+After you configure your +.pypirc+ you can upload packages from the index to the repository manager.
+
+In the example below, twine is invoked to tell your repository what server to use when uploading a package. The 
++-r+ flag is used to find the +nexus+ server in your +.pypirc+.
 
 ----
 twine upload -r nexus <filename>
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -11,18 +11,16 @@ https://pypi.python.org/pypi[https://pypi.python.org/pypi], and the site itself 
 https://www.python.org/psf/[Python Software Foundation].
 
 {pro} and {oss} support the Python Package Index (PyPI) repository format for proxy repositories. The repository 
-manager user will download and run libraries to retrieve, facilitate, and manage Python projects. This 
-allows the repository manager to take advantage of packages in the Python repository and other public registries 
+manager user will download and run libraries to retrieve, facilitate, and manage Python projects. Furthermore, 
+it can upload and publish index-available packages as a hosted repository. This allows the 
+repository manager to take advantage of packages in the Python repository and other public registries 
 without incurring repeated downloads.
 
 ////
-In the first sentence above, include hosted (as in... {pro}, {oss} allows you to upload/publish index-available & 
-your own packages & tools as hosted repository) Also, somewhere in the paragraph include how the repository 
-manager supports PyPI packages as a repository group (as in... the repository group merges and exposes the 
-contents of multiple repositories in one convenient URL)  
-////
-
-You can publish your own packages to a private PyPI hosted repository 
+Also, somewhere in the paragraph include how the repository manager supports PyPI packages as a repository group 
+(as in... the repository group merges and exposes the contents of multiple repositories in one convenient URL)
+Another note - can repoman upload and publish your own PyPIpackages and tools? Not sure
+////  
 
 [[pypi-proxy]]
 === Proxying PyPI Repositories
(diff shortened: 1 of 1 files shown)


chapter-pypi.asciidoc
```diff
@@ -43,14 +43,27 @@ components from the Python Package Index more reliably.
...
-////
+
 [[pypi-hosted]]
 === Hosting PyPI Repositories
 
-////
+Creating a PyPI hosted repository allows you to upload packages in the repository manager. The hosted 
+repository acts as an authoritative location for these components. This effectively creates an asset that becomes 
+a pointer to an external URL (such as a Git repository). You will create a custom file called +.pypirc+ to manage 
+credentials from the hosted repository, explained in <<pypi-upload>>.
```
joedragons: I'm +0 on the last sentence for 2 reasons:
* I'm not convinced there needs to be mention of this sentence here as it's talking about configuring hosted.  I could see this either way, thus +0 rather than -1.
* If it stays, tho I have no python dev experience, I think it likely people will already have this file created and will just be editing it.  I doubt folks new to python will start with NXRM.  You might use a generic word such as "maintain" or "utilize" which would be true in the case of creation or editing.
* Finally, the credentials maintained are actually the repository manager credentials not "from the hosted repository".

chapter-pypi.asciidoc
```diff
@@ -43,14 +43,27 @@ components from the Python Package Index more reliably.
...
+
+Minimal configuration steps are:
+
+* Define 'Name' - e.g. +pypi-all+
```
joedragons: Minor: This seems like a name you might utilize for a group.

chapter-pypi.asciidoc
```diff
@@ -43,14 +43,27 @@ components from the Python Package Index more reliably.
...
+* Select 'Blob store' for 'Storage'
+* Choose 'Deployment policy' for the 'Hosted' repository
```
joedragons: While this field is "required" it's not unselectable (it must be selected) [at least via the UI], so it's weird to have this here without further notation/mention.  Compare to docker which does not have this defined.  If there's a specific reason to put this here, it's not clear to me.

chapter-pypi.asciidoc
```diff
@@ -71,34 +84,36 @@ manager supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Ver
...
 ////
 
 [[pypi-configuration]]
 === Configuring PyPI Client Tools
 
-NOTE: Depending on your preference for either https://pypi.python.org/pypi/setuptools[setuptools] or pip, your 
-proxy configuration may vary.
+NOTE: Depending on your preference for either https://pypi.python.org/pypi/setuptools[setuptools], 
+https://pypi.python.org/pypi/twine[twine], https://docs.python.org/2.7/library/distutils.html[distutils], or pip 
+your proxy and hosted configuration may vary.
```
joedragons: Minor: May be wrong, but I think it's "configurations" (adding s) when using and...or switch to "or"...but I think and is right in this context.

chapter-pypi.asciidoc
```diff
@@ -120,6 +144,16 @@ specify the base URL for the PyPI proxy index.
...
 
+If you use twine, a package used to verify HTTPS connections for hosted repositories, your +pip.conf+ file 
+will contain metadata describing distribution properties present on the target system. Here is an example 
+of a +pip.conf+ file with a base URL pointing to a hosted repository:
+
+----
+[global]
+index = http://localhost:8081/repository/pypi-hosted/pypi
+index-url = http://localhost:8081/repository/pypi-hosted/simple
+----
```
joedragons: So this whole subchapter IMO is confusing.  It talks about configuring proxy, then hosted, then proxy, then hosted.  It's really disjoint, esp considering folks MAY (or may not) be doing one or the other.
Now ultimately, we'll have group which will tie everything together, so I'm just making one comment and kind of hoping that group makes this whole section be revamped (and not, necessarily, suggesting you spend anymore time here...but wanted to comment since I do think it reads confusingly).
Note, specifically (and especially if that's not the eventual plan) you tell someone to adjust their pip.conf to proxy then 2 paragraphs later you tell someone to adjust their pip.conf to hosted.  This WILL be group (hopefully) but right now you basically say do one or the other...or if they're not reading closely they'll only be able to do hosted when done if following this verbitim.

chapter-pypi.asciidoc
```diff
@@ -148,18 +181,25 @@ You can browse PyPI repositories in the user interface inspecting the components
...
-----
\ No newline at end of file
+----
+
+[[pypi-upload]]
+=== Uploading PyPI Packages
+
+NOTE: The steps to upload a PyPI package will vary if your system is configured with setuptools or twine.
+
+After you configure your +.pypirc+ you can upload packages from the index to the repository manager.
```
joedragons: Minor: Suggest a paragraph break here.  I think it's consistent with the rest of our documentation (to keep examples disjoint).

chapter-pypi.asciidoc
```diff
@@ -148,18 +181,25 @@ You can browse PyPI repositories in the user interface inspecting the components
...
\ No newline at end of file
+----
+
+[[pypi-upload]]
+=== Uploading PyPI Packages
+
+NOTE: The steps to upload a PyPI package will vary if your system is configured with setuptools or twine.
+
+After you configure your +.pypirc+ you can upload packages from the index to the repository manager. 
+In the example below, using twine, tells you what repository configuration file to use when uploading; the +-r+
```
joedragons: Awkward sentence structure (before ; ).  I think either (not sure which you were going for):
* In the example below, using twine, tell your repository what server to use when uploading;
* In the example below, use twine to tell your repository what server to use when uploading;
NOTE: In both writeups, I changes repository configuration file to server, speaking to "nexus" because that's what I believe you're talking about.  nexus isn't the name of the file it's the name of the section referenced within the file, so if I missed my mark it was either unclear or could use further attention.

joedragons: On reread, there may be a better word too than "server" as server could be Nexus server but I can't think of one right now=(

chapter-pypi.asciidoc
dulanism: ~all~=internal

chapter-pypi.asciidoc
dulanism: Noted for later discussion

chapter-pypi.asciidoc
```diff
@@ -11,16 +11,16 @@ https://pypi.python.org/pypi[https://pypi.python.org/pypi], and the site itself
 https://www.python.org/psf/[Python Software Foundation].
 
 {pro} and {oss} support the Python Package Index (PyPI) repository format for proxy repositories. The repository 
-manager user will download and run libraries to retrieve, facilitate, and manage Python projects. This 
-allows the repository manager to take advantage of packages in the Python repository and other public registries 
+manager user will download and run libraries to retrieve, facilitate, and manage Python projects. Furthermore, 
+it can upload and publish index-available packages as a hosted repository. This allows the 
+repository manager to take advantage of packages in the Python repository and other public registries 
 without incurring repeated downloads.
```
joedragons: Should this last sentence be before the hosted "Furthermore,"?  While this could be true, I think in order for it to be true with hosted, it'd entail downloading a package and pushing it to hosted.  Or maybe generalized a smidge.

chapter-pypi.asciidoc
```diff
@@ -43,14 +43,25 @@ components from the Python Package Index more reliably.
...
 
-////
+
 [[pypi-hosted]]
 === Hosting PyPI Repositories
 
-////
+Creating a PyPI hosted repository allows you to upload packages in the repository manager. The hosted 
+repository acts as an authoritative location for these components. This effectively creates an asset that becomes 
+a pointer to an external URL (such as a Git repository).
```
joedragons: I don't think the last sentence is true, but I may be confused.  @fjmilens3 ?  I suspect this may have been copied from bower, maybe unintentionally left.

chapter-pypi.asciidoc
dulanism: I (temporarily) rewrote his paragraph as: 

Nexus Repository Manager Pro and OSS support the Python Package Index (PyPI) repository format for proxy repositories. Furthermore, the repository manager user can download and run libraries to retrieve, facilitate, and manage Python projects. The repository manager can upload and publish downloaded packages into a hosted repository, once, so it can take advantage of packages in the Python repository and other public registries without incurring repeated downloads. 

(Though this will be updated once pypi[group] is ready; I will adapt accordingly.)

chapter-pypi.asciidoc
```diff
@@ -10,17 +10,20 @@ community of Python developers. The official PyPI is available for searches at
...
-{pro} and {oss} support the Python Package Index (PyPI) repository format for proxy repositories. The repository 
-manager user will download and run libraries to retrieve, facilitate, and manage Python projects. This 
-allows the repository manager to take advantage of packages in the Python repository and other public registries 
-without incurring repeated downloads.
+{pro} and {oss} support the Python Package Index (PyPI) repository format for proxy and hosted repositories. This 
+allows the repository manager to take advantage of the packages in the official Python Package Index and other 
+public registries without incurring repeated downloads of packages.
+
+The repository manager user can download and run libraries to retrieve, facilitate, and manage Python projects. 
+You can publish your own packages to a private index as a hosted repository on the repository manager. This
```
joedragons: Minor: The first sentence saying "The repository manager user" and the second saying "You" could infer that you are not the repository manager user.  This could probably be tightened to use the same, or genericize who is referenced (in one or both cases).

chapter-pypi.asciidoc
```diff
@@ -66,59 +80,95 @@ OS X and Microsoft Windows, download and install a Python version compatible wit
...
+Config will eventually be group, i.e. pypi-all
+
+----
+[global]
+index = http://localhost:8081/repository/pypi-all/pypi
+index-url = http://localhost:8081/repository/pypi-all/simple
+----
 ////
 
 Alternately you can create a +setup.cfg+. Like the +pip.conf+ option, the +index-url+ is the tag created to
```
joedragons: Minor suggest "Alternately you can create a +setup.cfg+ **if using easy_install**."  Puts a differentiation from twine and removes some of the wondering.

joedragons:
+1 w/ minor comments.

chapter-pypi.asciidoc
dulanism: Updated

chapter-pypi.asciidoc
dulanism: updated

PR #177, opened 2016-05-09
NEXUS-9827-pypi group repository

dulanism:
JIRA https://issues.sonatype.org/browse/NEXUS-9827
CI http://bamboo.s/browse/BOOK-NEXUS141

* includes SSL self-signed certificate setup

### DIFF
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -129,6 +129,9 @@ username = admin
 password = admin123
 ----
 
+NOTE: If you have multiple hosted repositories, you can add them to the +.pypirc+ file, each with a different 
+name, pointing to the corresponding respository URL.
+
 After this is configured, you can upload packages to the hosted repository, as explained in <<pypi-upload>>.
 
 *Global pip.conf file with a repository group*
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -10,10 +10,9 @@ community of Python developers. The official PyPI is available for searches at
 https://pypi.python.org/pypi[https://pypi.python.org/pypi], 
 and the site itself is maintained by the https://www.python.org/psf/[Python Software Foundation].
 
-{pro} and {oss} support the Python Package Index (PyPI) for proxy, hosted, and group 
-repository formats, which you can share within your organization. This allows the repository manager to take 
-advantage of the packages in the official Python Package Index and other public registries without incurring 
-repeated downloads. This will reduce time and bandwidth usage for accessing Python packages.
+Both {pro} and {oss} support proxying the Python Package Index. This allows the repository manager to take 
+advantage of the packages in the official Python Package Index without incurring repeated downloads. This will 
+reduce time and bandwidth usage for accessing Python packages.
 
 Also, you can publish your own packages to a private index as a hosted repository on the repository manager, then 
 expose the remote and private repositories as a repository group, which is a repository that merges and exposes 
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -7,8 +7,8 @@
 
 The Python Package Index, or PyPI, is a vast repository of open-source Python packages supplied by the worldwide 
 community of Python developers. The official PyPI is available for searches at
-https://pypi.python.org/pypi[https://pypi.python.org/pypi], and the site itself is maintained by the
-https://www.python.org/psf/[Python Software Foundation].
+https://pypi.python.org/pypi[https://pypi.python.org/pypi], 
+and the site itself is maintained by the https://www.python.org/psf/[Python Software Foundation].
 
 {pro} and {oss} support the Python Package Index (PyPI) repository format for proxy and hosted repositories. This 
 allows the repository manager to take advantage of the packages in the official Python Package Index and other 
@@ -80,7 +80,7 @@ OS X and Microsoft Windows, download and install a Python version compatible wit
 https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the pip installer from 
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
 
-NOTE: {pro} and {oss} support specific versions of Python, +pip+, and +setuptools+. For Python the repository 
+NOTE: {pro} and {oss} support specific versions of Python, pip, and setuptools. For Python the repository 
 manager supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Versions 19 and 20 of +setuptools+, 
 used to build and distribute Python dependencies, are compatible with the repository manager.
 
@@ -90,29 +90,33 @@ used to build and distribute Python dependencies, are compatible with the reposi
 
 NOTE: Depending on your preference for either https://pypi.python.org/pypi/setuptools[setuptools], 
 https://pypi.python.org/pypi/twine[twine], https://docs.python.org/2.7/library/distutils.html[distutils], and 
-+pip+ your proxy and hosted configuration may vary.
+pip your proxy and hosted configuration may vary.
 
 Once you have installed all necessary client tools from the Python Package Index, you can create and configure 
 a +.pypirc+ file to reference packages stored in the repository manager. Depending on your Python configuration 
 you can manage your repository groups with +pip.conf+ or +setup.cfg+ to have all commands, such as search and 
 install, run against your project.
 
-If you are authoring your own packages and want to distribute them to other users in your organization, you have 
-to upload them to a hosted repository on the repository manager, as explained in <<pypi-upload>>. The +.pypirc+ 
-holds your credentials for authentication when hosting a PyPI repository.
+*Configuring a proxy repository to use easy_install*
 
-Alternately you can create a +setup.cfg+, if using +easy_install+. Like +pip.conf+, the +index-url+ is the tag 
-created to specify the base URL for the PyPI package. In this example +index-url+ is set for a proxy repository:
+You can create a +setup.cfg+, if using +easy_install+. The +index-url+ is the tag created to specify 
+the base URL for the PyPI package. In this example +index-url+ is set for a proxy repository:
 
 ----
 [easy_install]
 index-url = http://localhost:8081/repository/pypi-proxy/simple
 ----
 
-In the +.pypirc+, specify the URL you want to deploy to the the target hosted repository in the +repository+ 
-value. Add +username+ and +password+ values to access the repository manager. The +.pypirc+ file contains 
-distutils, a default server used by PyPI that provides upload commands that stores assets and authentication 
-information.
+If you are authoring your own packages and want to distribute them to other users in your organization, you have 
+to upload them to a hosted repository on the repository manager. The +.pypirc+ holds your credentials for 
+authentication when hosting a PyPI repository.
+
+*Configuring your hosted repository with .pypirc*
+
+In the example +.pypirc+ file below, specify the URL you want to deploy to the target hosted repository in 
+the +repository+ value. Add +username+ and +password+ values to access the repository manager. The +.pypirc+ file 
+contains distutils, a default server used by PyPI that provides upload commands that stores assets and 
+authentication information.
 
 ----
 [distutils]
@@ -125,15 +129,9 @@ username = admin
 password = admin123
 ----
 
-If you use twine, a package used to verify HTTPS connections for hosted repositories, your +pip.conf+ file 
-will contain metadata describing distribution properties present on the target system. Here is an example of a 
-+pip.conf+ file with a base URL pointing to a hosted repository:
+After this is configured, you can upload packages to the hosted repository, as explained in <<pypi-upload>>.
 
-----
-[global]
-index = http://localhost:8081/repository/pypi-internal/pypi
-index-url = http://localhost:8081/repository/pypi-internal/simple
-----
+*Global pip.conf file with a repository group*
 
 In the instance you want your +pip.conf+ file to install or search Python within a group, configure the file 
 include the group repository URL.
@@ -144,25 +142,28 @@ index = http://localhost:8081/repository/pypi-all/pypi
 index-url = http://localhost:8081/repository/pypi-all/simple
 ----
 
+If you prefer to configure your global +pip.conf+ for (+pypi-proxy+) or hosted (+pypi-internal+) adjust the file 
+accordingly.
+
 [[pypi-ssl]]
 === SSL Usage for PyPI Repositories
 
 You can point your PyPI proxy repository over HTTPS to ensure a secure connection with a self-signed certificate. 
 To set up the repository manager to serve HTTPS follow the configuration steps in <<ssl>>.
 
-Also, you can set up +pip+ to use the certificate to enable SSL and fetch packages securely. Additional
+Also, you can set up pip to use the certificate to enable SSL and fetch packages securely. Additional
 configuration is necessary for the HTTPS client implementation to work. This assumes the repository manager has 
 already been set up to use SSL, so verify your certificate works. Run the following command:
 
 ----
-openssl verify nexus.pem
+openssl verify <example-cerfificate>
 ----
 
-When your +nexus.pem+ certificate is proven to work, update your +pip.conf+ as follows:
+When your certificate is proven to work, update your +pip.conf+ as follows:
 ----
 [global]
-index = https://localhost:8443/repository/pypi-proxy/pypi
-index-url = https://localhost:8443/repository/pypi-proxy/simple
+index = https://localhost:8443/repository/pypi-all/pypi
+index-url = https://localhost:8443/repository/pypi-all/simple
 cert = nexus.pem
 ----
 
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -167,26 +167,13 @@ index-url = http://localhost:8081/repository/pypi-internal/simple
 ----
 
 ////
-////
-Config will eventually be group, i.e. pypi-all
 
-----
-[global]
-index = http://localhost:8081/repository/pypi-all/pypi
-index-url = http://localhost:8081/repository/pypi-all/simple
-----
-////
+[[pypi-ssl]]
+=== SSL Usage for PyPI Repositories
 
-////
-add section on uploading distributions - see twine above
-Uploading Distributions
-
-add section on 
-SSL Usage for PyPI Client Tools
-
-You can secure inbound and outbound communication between the repository manager and the PyPI repository using 
-SSL. This assumes the repository manager has already been set up to use SSL. If not review SSL configuration at 
-<<ssl>>.
+If you use pip to maintain your packages from the index, you can manage inbound and outbound communication 
+between the repository manager and the PyPI repository with a self-signed SSL certificate. This assumes the 
+repository manager has already been set up to use SSL.
 
 Update your +pip.conf+ as follows:
 ----
@@ -195,7 +182,6 @@ index = https://localhost:8443/repository/pypi-proxy/pypi
 index-url = https://localhost:8443/repository/pypi-proxy/simple
 cert = nexus.pem
 ----
-////
 
 
 [[pypi-browse-search]]
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -52,12 +52,7 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
 Minimal configuration steps are:
 
 * Define 'Name' - e.g. +pypi-internal+
-* Pick a 'Blob store' for 'Storage'
-
-////
-Note: a ticket filed - NXBOOK-947 : Consistent naming of repos was created from a community member that might 
-impact naming conventions for later examples
-////
+* Pick a 'Blob store' for 'Storage'a
 
 
 [[pypi-group]]
@@ -82,11 +77,11 @@ Minimal configuration steps are:
 The latest versions of such Linux distributions as CentOS and Ubuntu come packaged with Python 2.7 and 
 https://pip.pypa.io/en/stable/[pip], a tool for installing and managing Python packages from the index. For Mac 
 OS X and Microsoft Windows, download and install a Python version compatible with the repository manager from
-https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the pip installer from 
+https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the +pip+ installer from 
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
 
-NOTE: {pro} and {oss} support specific versions of Python, pip, and setuptools. For Python the repository 
-manager supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Versions 19 and 20 of setuptools, 
+NOTE: {pro} and {oss} support specific versions of Python, +pip+, and +setuptools+. For Python the repository 
+manager supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Versions 19 and 20 of +setuptools+, 
 used to build and distribute Python dependencies, are compatible with the repository manager.
 
 
@@ -94,8 +89,8 @@ used to build and distribute Python dependencies, are compatible with the reposi
 === Configuring PyPI Client Tools
 
 NOTE: Depending on your preference for either https://pypi.python.org/pypi/setuptools[setuptools], 
-https://pypi.python.org/pypi/twine[twine], https://docs.python.org/2.7/library/distutils.html[distutils], and pip 
-your proxy and hosted configuration may vary.
+https://pypi.python.org/pypi/twine[twine], https://docs.python.org/2.7/library/distutils.html[distutils], and 
++pip+ your proxy and hosted configuration may vary.
 
 Once you have installed all necessary client tools from the Python Package Index, you can create and configure 
 a +.pypirc+ file to reference stored and proxied packages. Depending on your Python configuration you can manage 
@@ -106,8 +101,8 @@ If you are authoring your own packages and want to distribute them to other user
 to upload them to a hosted repository on the repository manager, as explained in <<pypi-upload>>. The +.pypirc+ 
 holds your credentials for authentication when hosting a PyPI repository.
 
-Alternately you can create a +setup.cfg+. Like +pip.conf+, the +index-url+ is the tag created to specify the base 
-URL for the PyPI proxy index.
+Alternately you can create a +setup.cfg+, if using +easy_install+. Like +pip.conf+, the +index-url+ is the tag 
+created to specify the base URL for the PyPI package. In this exampl +index-url+ is set for a proxy repository:
 
 ----
 [easy_install]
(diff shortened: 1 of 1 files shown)


chapter-pypi.asciidoc
```diff
@@ -170,25 +167,13 @@ index-url = http://localhost:8081/repository/pypi-internal/simple
...
-add section on uploading distributions - see twine above
-Uploading Distributions
-
-add section on 
-SSL Usage for PyPI Client Tools
-
-You can secure inbound and outbound communication between the repository manager and the PyPI repository using 
-SSL. This assumes the repository manager has already been set up to use SSL. If not review SSL configuration at 
-<<ssl>>.
+If you use pip to maintain your packages from the index, you can manage inbound and outbound communication
```
fjmilens3: I'm somewhat lost on this section. You can point the proxy to PyPI over https to make sure that's more secure, and you can set up your local pip to use the Nexus certificate so that can go over SSL as well. But the two are different and touch on different aspects of how packages actually get downloaded.

chapter-pypi.asciidoc
```diff
@@ -99,34 +98,26 @@ https://pypi.python.org/pypi/twine[twine], https://docs.python.org/2.7/library/d
...
+your repository groups with +pip.conf+ or +setup.cfg+ to have all commands, such as search and install, run 
+against your project.
 
 If you are authoring your own packages and want to distribute them to other users in your organization, you have 
 to upload them to a hosted repository on the repository manager, as explained in <<pypi-upload>>. The +.pypirc+ 
 holds your credentials for authentication when hosting a PyPI repository.
 
-Alternately you can create a +setup.cfg+. Like the +pip.conf+ option, the +index-url+ is the tag created to 
-specify the base URL for the PyPI proxy index.
+Alternately you can create a +setup.cfg+. Like +pip.conf+, the +index-url+ is the tag created to specify the base
```
fjmilens3: As an aside, this is something they'll probably only run into if they're using easy_install.

chapter-pypi.asciidoc
dulanism: I still have more to add to this section (SSL), I just uncommented it out in the most recent commit. It was placeholder from the original google doc pypi testing guide.

chapter-pypi.asciidoc
```diff
@@ -15,15 +15,10 @@ allows the repository manager to take advantage of the packages in the official
...
-time and bandwidth usage for accessing Python packages as well as share your packages within 
-your organization in a hosted repository.
-
-////
-Also, somewhere in the paragraph include how the repository manager supports PyPI packages as a repository group 
-(as in... the repository group merges and exposes the contents of multiple repositories in one convenient URL)
-Another note - can repoman upload and publish your own PyPIpackages and tools? Not sure
-////  
+your own packages to a private index as a hosted repository on the repository manager, then expose the remote and 
+private repositories to PyPI as a repository group, which is a repository that merges and exposes the contents of
```
joedragons: Minor: Suggest remove "to PyPI".  The sentence remains true and I am not sure what's gained by the "to".

chapter-pypi.asciidoc
```diff
@@ -15,15 +15,10 @@ allows the repository manager to take advantage of the packages in the official
...
-your organization in a hosted repository.
-
-////
-Also, somewhere in the paragraph include how the repository manager supports PyPI packages as a repository group 
-(as in... the repository group merges and exposes the contents of multiple repositories in one convenient URL)
-Another note - can repoman upload and publish your own PyPIpackages and tools? Not sure
-////  
+your own packages to a private index as a hosted repository on the repository manager, then expose the remote and 
+private repositories to PyPI as a repository group, which is a repository that merges and exposes the contents of 
+multiple repositories in one convenient URL. This allows you to reduce time and bandwidth usage for accessing
```
joedragons: Minor: Suggest "All this allows...".

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
 Minimal configuration steps are:
 
 * Define 'Name' - e.g. +pypi-internal+
-* Select 'Blob store' for 'Storage'
+* Pick a 'Blob store' for 'Storage'a
```
joedragons: Typo "a" at the end here.  Looks like it's also messing up the markup in the pdf, hopefully removing the a resolves it, otherwise might be something else.

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
 
 [[pypi-installation]]
 === Installing PyPI Client Tools
 
 The latest versions of such Linux distributions as CentOS and Ubuntu come packaged with Python 2.7 and 
 https://pip.pypa.io/en/stable/[pip], a tool for installing and managing Python packages from the index. For Mac 
 OS X and Microsoft Windows, download and install a Python version compatible with the repository manager from
-https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the pip installer from 
+https://www.python.org/downloads/[https://www.python.org/downloads/]. Download the +pip+ installer from 
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
```
joedragons: Minor: This may just be currently unfortunate wrapping of the pdf, but I see a - at the end of the line that doesn't appear to be present in this URL.  Link works fine.  I didn't check the html build (or epub) at this time.

screen shot 2016-05-13 at 5 48 58 pm

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
 
-NOTE: {pro} and {oss} support specific versions of Python, pip, and setuptools. For Python the repository 
-manager supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Versions 19 and 20 of setuptools, 
+NOTE: {pro} and {oss} support specific versions of Python, +pip+, and +setuptools+. For Python the repository 
+manager supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Versions 19 and 20 of +setuptools+,
```
joedragons: Minor: Your ++ markup seems inconsistent to me here.  Python is also a tool and the second pip ("for pip versions 7 and 8") is not marked up.

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
-Repository manager users might a re: since both twine and subsequently pip to install twine, would be hosted
-Next install Twine. Twine is a utility that provides secure authentication to PyPI over HTTPS. Twine's only 
-function is to upload distributions. If the user uses pip then here's a sample command:
-////
-
 
 [[pypi-configuration]]
 === Configuring PyPI Client Tools
 
 NOTE: Depending on your preference for either https://pypi.python.org/pypi/setuptools[setuptools],
```
joedragons: Minor: I noticed while reviewing the new ++ markup that above you define/link pip then subsequent references have ++ markup, but above (current line 84) you reference/markup setuptools and link here.  Minor inconsistent IMO.

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
++pip+ your proxy and hosted configuration may vary.
 
 Once you have installed all necessary client tools from the Python Package Index, you can create and configure 
-a +.pypirc+ file to reference stored and proxied packages. Depending on your Python configuration you can use 
-+pip.conf+ or +setup.cfg+ to have all commands run against your project.
-
-////
-This paragraph will be updated to mention pypirc behavior, interaction with PyPI repository group. 
-////
+a +.pypirc+ file to reference stored and proxied packages. Depending on your Python configuration you can manage
```
joedragons: Minor: Word "proxied" causes me pause here.  As far as I can tell the pypirc configuration just affects hosted.  So by saying proxied, it could be inferred it affects more.  Technically, once uploaded all hosted and proxied files are proxied however I think that symantic is more than the word intended=)  Recommend striking "and proxied" or changing proxied to anotehr word that infers proxy without saying "proxy" (or switching to "hosted" depending on the intent).

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
 
 Once you have installed all necessary client tools from the Python Package Index, you can create and configure 
-a +.pypirc+ file to reference stored and proxied packages. Depending on your Python configuration you can use 
-+pip.conf+ or +setup.cfg+ to have all commands run against your project.
-
-////
-This paragraph will be updated to mention pypirc behavior, interaction with PyPI repository group. 
-////
+a +.pypirc+ file to reference stored and proxied packages. Depending on your Python configuration you can manage 
+your repository groups with +pip.conf+ or +setup.cfg+ to have all commands, such as search and install, run
```
joedragons: I believe search is always using pip.  There may be a way to search using easy_install I am unaware of, but if not this can be reworded to not include search "such as to have all install commands...".

chapter-pypi.asciidoc
dulanism: I was looking at other sites on how "setuptools" and "pip" are styled. I'll update without the markup.

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
-////
-This paragraph will be updated to mention pypirc behavior, interaction with PyPI repository group. 
-////
+a +.pypirc+ file to reference stored and proxied packages. Depending on your Python configuration you can manage 
+your repository groups with +pip.conf+ or +setup.cfg+ to have all commands, such as search and install, run 
+against your project.
 
 If you are authoring your own packages and want to distribute them to other users in your organization, you have 
 to upload them to a hosted repository on the repository manager, as explained in <<pypi-upload>>. The +.pypirc+ 
 holds your credentials for authentication when hosting a PyPI repository.
```
joedragons: This whole section still reads weird to me, so since I think this is final cut (with group now integrated), going to make suggestion.  I'd remove this paragraph and add the ref to after the configuration section).  I'll comment below where I think the ref should be added.

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
 
-Alternately you can create a +setup.cfg+. Like the +pip.conf+ option, the +index-url+ is the tag created to 
-specify the base URL for the PyPI proxy index.
+Alternately you can create a +setup.cfg+, if using +easy_install+. Like +pip.conf+, the +index-url+ is the tag
```
joedragons: This says alternatively and like pip.conf, but the pip.conf section is 2 paragraphs down.  Consider moving this paragraph under pip.conf or rewording such as "To configure your proxy to use +easy_install+..." (that's just a possible leadin, more rewording would be needed).

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
+against your project.
 
 If you are authoring your own packages and want to distribute them to other users in your organization, you have 
 to upload them to a hosted repository on the repository manager, as explained in <<pypi-upload>>. The +.pypirc+ 
 holds your credentials for authentication when hosting a PyPI repository.
 
-Alternately you can create a +setup.cfg+. Like the +pip.conf+ option, the +index-url+ is the tag created to 
-specify the base URL for the PyPI proxy index.
+Alternately you can create a +setup.cfg+, if using +easy_install+. Like +pip.conf+, the +index-url+ is the tag 
+created to specify the base URL for the PyPI package. In this exampl +index-url+ is set for a proxy repository:
```
joedragons: Typo here "exampl" missing an e (example).

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
-////
-In paragraph above, add section that consumers can download the registered package via the repository group
-Also, this sentence refers to registering "This establishes a metadata file in the repository that links to the 
-source code repository". Leaving it here to confirm if the pypirc does such.
-////
-
-In the +.pypirc+, specify the URL for the target hosted repository in the +repository+ value, along with the 
-+username+ and +password+ to access the repository manager. The +.pypirc+ file contains distutils, a default 
-server used by PyPI that provides upload commands that stores assets and authentication information.
+In the +.pypirc+, specify the URL you want to deploy to the the target hosted repository in the +repository+
```
joedragons: Minor: Recommend leadin, what the configuration is for like "To configure your hosted repository,".

chapter-pypi.asciidoc
```diff
@@ -57,76 +52,67 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
-source code repository". Leaving it here to confirm if the pypirc does such.
-////
-
-In the +.pypirc+, specify the URL for the target hosted repository in the +repository+ value, along with the 
-+username+ and +password+ to access the repository manager. The +.pypirc+ file contains distutils, a default 
-server used by PyPI that provides upload commands that stores assets and authentication information.
+In the +.pypirc+, specify the URL you want to deploy to the the target hosted repository in the +repository+ 
+value. Add +username+ and +password+ values to access the repository manager. The +.pypirc+ file contains 
+distutils, a default server used by PyPI that provides upload commands that stores assets and authentication 
+information.
```
joedragons: As mentioned above (https://github.com/sonatype/nexus-book/pull/177/files#r63258062), if you strike that paragraph, you could add "After this is configured, you can upload packages to the hosted repository, as explained in Section 13.9." (or whatever varient) to the end.

chapter-pypi.asciidoc
```diff
@@ -140,65 +126,45 @@ password = admin123
...
-////
-Config will eventually be group, i.e. pypi-all
+In the instance you want your +pip.conf+ file to install or search Python within a group, configure the file 
+include the group repository URL.
 
 ----
 [global]
 index = http://localhost:8081/repository/pypi-all/pypi
 index-url = http://localhost:8081/repository/pypi-all/simple
 ----
```
joedragons: To help simplify I recommend just leaving the pip.conf reference group with a comment following that it can be substituted to proxy or hosted repositories if that's all they're using.  Basically you have 3 different pip.confs here that all say the same thing except the fact the location is different.  I think it can be condensed.
I reviewed top down, so sorry this is late.  If you agree earlier comments may be moot.  If you disagree, I'd at least move this up.  I think our continued hope is people use a group (I think many other formats read that way).

chapter-pypi.asciidoc
```diff
@@ -140,65 +126,45 @@ password = admin123
...
```
joedragons: Minor: Outside the scope of the changes (thus minor) but it may be worth adding a note like "If you have multiple hosted repositories, you can add them here, each with a different name and pointing to the corresponding respoitory URL."  I think this is different and some of our other formats only allow a single repo to be configured.  This might be intuitive tho. cc/ @fjmilens3 second opinion?

chapter-pypi.asciidoc
```diff
@@ -140,65 +126,45 @@ password = admin123
...
+Also, you can set up +pip+ to use the certificate to enable SSL and fetch packages securely. Additional
+configuration is necessary for the HTTPS client implementation to work. This assumes the repository manager has 
+already been set up to use SSL, so verify your certificate works. Run the following command:
 
 ----
-[global]
-index = http://localhost:8081/repository/pypi-all/pypi
-index-url = http://localhost:8081/repository/pypi-all/simple
+openssl verify nexus.pem
 ----
```
joedragons: For me, where the nexus.pem comes from was not obvious and I think worth adding here.  That might be my ignorance tho.  cc/ @fjmilens3 second opinion?

chapter-pypi.asciidoc
```diff
@@ -140,65 +126,45 @@ password = admin123
...
-<<ssl>>.
-
-Update your +pip.conf+ as follows:
+When your +nexus.pem+ certificate is proven to work, update your +pip.conf+ as follows:
 ----
 [global]
 index = https://localhost:8443/repository/pypi-proxy/pypi
 index-url = https://localhost:8443/repository/pypi-proxy/simple
 cert = nexus.pem
 ----
```
joedragons: Minor: Again recommend our samples be for group (tested & it does work in case you're wondering).

chapter-pypi.asciidoc
fjmilens3: Correct, pip has a search command implemented, easy_install does not.

chapter-pypi.asciidoc
fjmilens3: I think that makes sense if we're not telling people how to get the cert in general. I might defer to what we do for other repos where you need to get a cert and configure something locally with it, whatever the normal way of explaining that is.

joedragons: That's a good point; I don't recall this being anywhere else but if I'm wrong, could link there or if you did and I missed it, ignore my comment.

chapter-pypi.asciidoc
fjmilens3: I don't think you can add them *like that* for your pip.conf or setup.cfg, can you? I think those involve going into the --extra-index-url stuff, and I think anyone that knowledgeable isn't going to need us to explain that to them. 

The place where you specify the actual names and links is in your .pypirc For how you can specify a specific server with -r when using twine, that's separate and in your .pypirc file. That has more to do with uploading stuff via twine and/or distutils than it does about pulling them down.

chapter-pypi.asciidoc
dulanism: I linked to the SSL config section. Thx @fjmilens3 & @joedragons

fjmilens3: My first thought is to say that you'll need to obtain the appropriate .pem, and then you could have an example of how to get it yourself if you don't have another means (I'm wondering about different internal distribution procedures for certs as well).

I'd think the how-to-get-the-cert part would be something that might eventually end up happier in a separate section since this may not be the only place we have to live with such things.

chapter-pypi.asciidoc
joedragons: No sorry, I meant for the .pypirc file.  I misinjected the comment into the wrong place.  Going to remove and reposition (if possible).

joedragons: Minor (2nd attempt): Outside the scope of the changes (thus minor) but it may be worth adding a note like "If you have multiple hosted repositories, you can add them here, each with a different name and pointing to the corresponding respoitory URL." I think this is different and some of our other formats only allow a single repo to be configured. This might be intuitive tho. cc/ @fjmilens3 second opinion?

chapter-pypi.asciidoc
dulanism: Removed, updated

chapter-pypi.asciidoc
dulanism: Updated.

chapter-pypi.asciidoc
dulanism: Removed.

chapter-pypi.asciidoc
joedragons: Okie, so then it doesn't need to be named nexus.pem so probably can generalize the text slightly (like below "When your certificate is proven to work," etc)

chapter-pypi.asciidoc
dulanism: Adding bold sub-header: Configuring your hosted repository

chapter-pypi.asciidoc
dulanism: Updated and condensed accordingly.

chapter-pypi.asciidoc
dulanism: I updated the example command as follows:

`openssl verify <example-certificate>`

chapter-pypi.asciidoc
dulanism: Noted.

chapter-pypi.asciidoc
```diff
@@ -7,23 +7,18 @@
...
-
-////
-Also, somewhere in the paragraph include how the repository manager supports PyPI packages as a repository group 
-(as in... the repository group merges and exposes the contents of multiple repositories in one convenient URL)
-Another note - can repoman upload and publish your own PyPIpackages and tools? Not sure
-////  
+your own packages to a private index as a hosted repository on the repository manager, then expose the remote and 
+private repositories as a repository group, which is a repository that merges and exposes the contents of 
+multiple repositories in one convenient URL. All this allows you to reduce time and bandwidth usage for accessing 
+Python packages as well as share your packages within your organization in a hosted repository.
```
joedragons: Last sentence reads a bit weird.  I think I've previously read assuming that the first half ("...reduce time and bandwidth...") refers to proxy but that's also the case for hosted really.  The second part ("...share your packages...") also refers to both, so it's weird that it says hosted specifically and the first part doesn't.  And all that is after you wrote about groups.  I am guessing this is somewhat legacy and needs a minor revamp.

chapter-pypi.asciidoc
```diff
@@ -57,19 +52,24 @@ To host a PyPI package, create a new 'pypi(hosted)' recipe as documented in <<ho
...
-////
 [[pypi-group]]
 === PyPI Repository Groups
 
-TBD
-////
+A repository group is the recommended way to expose all your PyPI repositories from the repository manager to 
+your users, with minimal additional client side configuration. A repository group allows you to expose the 
+aggregated content of multiple proxy and hosted repositories as well as other repository groups with one URL in 
+tool configuration. PyPI group repositories can be created a with the 'pypi(group)' recipe as documented in
```
joedragons: "can be created a with the" - I think the 'a' might be extra here.

chapter-pypi.asciidoc
```diff
@@ -81,52 +81,42 @@ https://www.python.org/downloads/[https://www.python.org/downloads/]. Download t
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
 
 NOTE: {pro} and {oss} support specific versions of Python, pip, and setuptools. For Python the repository 
-manager supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Versions 19 and 20 of setuptools, 
+manager supports 2.6, 2.7, 3.4, 3.5. For pip versions 7 and 8 are supported. Versions 19 and 20 of +setuptools+,
```
joedragons: I think on pass 1, I commented that it was weird that pip and setuptools were marked up and Python wasn't.  Now it's weird that setuptools is marked up and pip and Python weren't=)  If there's consistency/documentation ruling here I am missing it.

dulanism: That was overlooked. Will fix.

chapter-pypi.asciidoc
dulanism: Overlooked. Will fix

chapter-pypi.asciidoc
```diff
@@ -81,52 +81,42 @@ https://www.python.org/downloads/[https://www.python.org/downloads/]. Download t
 https://pip.pypa.io/en/stable/installing/[https://pip.pypa.io/en/stable/installing/].
 
 NOTE: {pro} and {oss} support specific versions of Python, pip, and setuptools. For Python the repository
```
joedragons: Minor: Might be worth checking with @DarthHater but this may be one place we want to say Sonatype supports.  I imagine (tho have not tested) that Python 3.3 works with this.  So to say the Repository Manager supports is a bit deceiving.

chapter-pypi.asciidoc
joedragons: More importantly (thus distinct comment), setuptools is now 20 and 21 (latest 2 versions; 21 came out last week, I think).  You may actually just want to revamp saying the latest two versions of the tool, that'll prevent us from making frequent updates such as this.

chapter-pypi.asciidoc
joedragons: What I meant more than markup was that if your intent was to put the link at the earliest reference to the tool, there's an earlier reference to it above.  Comment was minor as we have no surity that they'll be reading any of this in order (and assume/agree putting links every time is overkill).

chapter-pypi.asciidoc
```diff
@@ -81,52 +81,42 @@ https://www.python.org/downloads/[https://www.python.org/downloads/]. Download t
...
-
-Alternately you can create a +setup.cfg+. Like the +pip.conf+ option, the +index-url+ is the tag created to 
-specify the base URL for the PyPI proxy index.
+You can create a +setup.cfg+, if using +easy_install+. The +index-url+ is the tag created to specify 
+the base URL for the PyPI package. In this example +index-url+ is set for a proxy repository:
 
 ----
 [easy_install]
 index-url = http://localhost:8081/repository/pypi-proxy/simple
 ----
```
joedragons: It is possible to have easy_install also work with groups, so you may want to put a note like the one under pip.conf group (which was "If you prefer to configure your global pip.conf for (pypi-proxy) or hosted (pypi-internal) adjust the file accordingly.") here.
PS: I like the new subheaders=)

chapter-pypi.asciidoc
```diff
@@ -81,52 +81,42 @@ https://www.python.org/downloads/[https://www.python.org/downloads/]. Download t
...
 
-////
-In paragraph above, add section that consumers can download the registered package via the repository group
-Also, this sentence refers to registering "This establishes a metadata file in the repository that links to the 
-source code repository". Leaving it here to confirm if the pypirc does such.
-////
+If you are authoring your own packages and want to distribute them to other users in your organization, you have 
+to upload them to a hosted repository on the repository manager. The +.pypirc+ holds your credentials for 
+authentication when hosting a PyPI repository.
```
joedragons: While true, I think this section belongs in the subheader "*Configuring your hosted repository with .pypirc*" aka it needs moved down slightly.

chapter-pypi.asciidoc
joedragons: Beautiful.

chapter-pypi.asciidoc
```diff
@@ -139,66 +129,46 @@ username = admin
...
 ----
 [global]
 index = http://localhost:8081/repository/pypi-all/pypi
 index-url = http://localhost:8081/repository/pypi-all/simple
 ----
-////
 
-Alternately you can create a +setup.cfg+, if using easy_install. Like the +pip.conf+ option, the +index-url+ is 
-the tag created to specify the base URL for hosted components from the index.
+If you prefer to configure your global +pip.conf+ for (+pypi-proxy+) or hosted (+pypi-internal+) adjust the file
```
joedragons: Minor: Is the word "proxy" missing before "(+pypi-proxy+)"?  The meaning gets across but it's inconsistent with the followup "hosted (+pypi-internal+)".  Alternatively, you could probably just remove the examples and put "...adjust the index and index URL to match".

chapter-pypi.asciidoc
```diff
@@ -139,66 +129,46 @@ username = admin
...
+accordingly.
 
-----
-[easy_install]
-index-url = http://localhost:8081/repository/pypi-internal/simple
-----
+[[pypi-ssl]]
+=== SSL Usage for PyPI Repositories
+
+You can point your PyPI proxy repository over HTTPS to ensure a secure connection with a self-signed certificate.
```
joedragons: Suggesting generalizing "You can point your PyPI proxy repository" to "You can proxy packages".  This then shows that no matter what you're proxying (group, proxy, hosted) it can be setup using SSL.  As I read it now (I believe as it was originally written) it was just for proxy but it should work with all (definitely tested and works for groups).

chapter-pypi.asciidoc
```diff
@@ -139,66 +129,46 @@ username = admin
...
 
-add section on 
-SSL Usage for PyPI Client Tools
-
-You can secure inbound and outbound communication between the repository manager and the PyPI repository using 
-SSL. This assumes the repository manager has already been set up to use SSL. If not review SSL configuration at 
-<<ssl>>.
-
-Update your +pip.conf+ as follows:
+When your certificate is proven to work, update your +pip.conf+ as follows:
```
joedragons: Minor: "as the following example:"? (similar to note above, this can be group, proxy or hosted.  Other places say example which I think is fitting here too)

joedragons:
+1 assuming non-minor notes I made are addressed.  Am happy to review again if desired.

PR #181, opened 2016-05-17
OSS-PAID bundle docs

dulanism:
JIRA - https://issues.sonatype.org/browse/NEXUS-10117
CI - TBD

### DIFF
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -888,6 +888,10 @@ side resolution done by Maven each time.
 A paid license is necessary to upgrade {oss} to {pro}, and to keep {pro} paid features operational. You must be 
 logged in as a user with sufficient privileges to manage licenses.
 
+////
+update this section per NEXUS-10117
+////
+
 [[admin-license-renew]]
 ==== Uploading a License
 
--- a/chapter-configuration.asciidoc
+++ b/chapter-configuration.asciidoc
@@ -905,8 +905,8 @@ After the file is successfully uploaded, you will receive a notification to rest
 restart the 'License type', shown in the 'Licensing' panel, will display the features associated with your 
 license.
 
-[[admin-lic-users]]
-==== Managing Active Users
+[[admin-lic-connections]]
+==== Managing Recent Connections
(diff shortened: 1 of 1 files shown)


chapter-configuration.asciidoc
```diff
@@ -901,8 +901,9 @@ The paid license is a special +.lic+ file that you upload to the 'Licensing' fea
 * Click 'I agree' to the terms stated in the End User License Agreement.
 * Click 'Authenticate' to complete the upload.
 
-After the file is selected and uploaded, the 'License type' field will display the corresponding bundle 
-you purchased.
+After the file is successfully uploaded, you will receive a notification to restart the repository manager. After
+restart the 'License type', shown in the 'Licensing' panel, will display the paid bundle corresponding to the
```
joedragons: As hinted in your ticket comment I think "the 'License type'" should either be "the 'License type' field" or just "'License type'" (no the).  I think the word field is fine.  If I misunderstood your Q, let me know.

kellyrob99: I think perhaps a better description might be "will display the features associated with your license", but we might want to defer to sales nomenclature - @dulanism Care to ask how they refer to licensed features?

dulanism: Will do @kellyrob99

joedragons:
@dulanism can you include the update to https://github.com/sonatype/nexus-internal/commit/19842e4dabe4e8fd361b3ce76c5527c8e56595eb in this PR?  Otherwise I can make another edit once this is clear.  Let me know if adjustment is unclear, thanks.

joedragons:
So after building it out and restoring OSS, I found what I had guessed to be true.  "Available in Nexus Repository Manager only" no longer applies.  As written, this section applies to both OSS and PRO.  The only way I think it'd be different is if we discussed the license details which I believe we talked about specifically not doing when we first created this section (before this update).

joedragons:
+1 after comments are addressed.  Happy to rereview if you like, as usual.

kellyrob99:
+1

PR #185, opened 2016-05-23
Updated sentence to affirm RHC only provides security for components that Sonatype supports

dulanism:
To avoid the confusion for those who may take the statement to mean
that we supply security and license data for docker images (real use
case).

### DIFF
--- a/chapter-repository-health-check.asciidoc
+++ b/chapter-repository-health-check.asciidoc
@@ -13,7 +13,7 @@ known data about the components in proxy repositories, including license informa
 and other statistics like relative usage popularity and age. {rhc} allows you to examine a high level overview of
 the available security and license data of components in an analyzed repository.
 
-When enabled, {rhc} analyzes the security of all components for which Sonatype provides support. maven2 proxy
+When enabled, {rhc} analyzes the risk of all components for which Sonatype provides security. maven2 proxy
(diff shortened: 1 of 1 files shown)


joedragons:
+1

chapter-repository-health-check.asciidoc
```diff
@@ -13,7 +13,7 @@ known data about the components in proxy repositories, including license informa
 and other statistics like relative usage popularity and age. {rhc} allows you to examine a high level overview of
 the available security and license data of components in an analyzed repository.
 
-When enabled, {rhc} analyzes all components found in a proxy repository regardless of the format. maven2 proxy
+When enabled, {rhc} analyzes the security of all components for which Sonatype provides support. maven2 proxy
```
DarthHater: Sentence reads a bit wonky. "analyzes the risk of all components for which Sonatype has support for."  Risk might be a better word since we also check licenses.

dulanism:
"When enabled, {rhc} analyzes the risk of all components for which Sonatype provides security..."

mhansen-sonatype:
Here is something for consideration:

"{rhc} analyzes the contents of the components within each repository that is enabled.  Since new versions of software are continually released and new component formats are routinely being added, not every component or component format has corresponding data available.  However, new data is made available continuously, with the priority being influenced by the frequency of usage - i.e. simply using {rhc} helps improve the data you ultimately have access to."

PR #186, opened 2016-05-26
refactor of usage of pro and new nxrm macro

mosabua:
just fyi @dulanism

### DIFF
--- a/nexus-book-declarations.asciidoc
+++ b/nexus-book-declarations.asciidoc
@@ -5,6 +5,7 @@ Declarations of macros to save on typing and increase consistency of terms used
 ////
 The "things" we ship, we do NOT add solutions
 /////
+:nxrm: Nexus Repository Manager
 :oss: Nexus Repository Manager OSS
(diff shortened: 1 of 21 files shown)


nexus-book-declarations.asciidoc
```diff
@@ -5,6 +5,7 @@ Declarations of macros to save on typing and increase consistency of terms used
 ////
 The "things" we ship, we do NOT add solutions
 /////
+:nxrm: Nexus Repository Manager
```
mosabua: Here is the new macro .. fyi @dulanism @annrollo @jwayman

nexus-book-declarations.asciidoc
```diff
@@ -5,6 +5,7 @@ Declarations of macros to save on typing and increase consistency of terms used
...
 :oss: Nexus Repository Manager OSS
 :pro: Nexus Repository Manager Pro
```
mosabua: And here the new name..

PR #191, opened 2016-06-14
fix some typos in the 'Installation' chapter

rlaun:


dulanism:
Thank you, @rlaun. I'll update the chapter with your edits.

chapter-installation.asciidoc
```diff
@@ -838,7 +838,7 @@ Therefore, if the port is set to `9081`, the exposed URL will be `http://localho
 
 <<installation-archive,Distribution archive installation>> of the repository manager configures the location of
 the <<data-directory, data directory>> to be nested inside the application directory. The
-<<installation-installer,installers>> allows the user to configure the location of this directory to be any
```
mosabua: this should be "installers allow" ...

rlaun: Sorry, I was not sure which way is the more correct one.
I decided to phrase it that way because the "Directories" section uses "installer allows", too.

PR #195, opened 2016-06-23
Create chapter for Atlassian crowd support for NXRM pro 3.1

dulanism:
For ticket NEXUS-10335
CI: http://bamboo.s/browse/BOOK-NEXUS157

- Create chapter
- Test/walkthru product
- Adapt/update chapter ported from 2.13 to 3.1 (pro only)

2nd revision


book-nexus.asciidoc
```diff
@@ -17,6 +17,8 @@ include::chapter-configuration.asciidoc[]
 
 include::chapter-security.asciidoc[]
 
+include::chapter-crowd.asciidoc[]
```
joedragons: Minor: Being PRO only, I wonder if this should be further down in our list.  OTOH, maybe higher to drive attention.  NX2 has it around the middle with formats following it (as formats follow here).  Just a thought.  More first impression than feeling strongly.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
+[[crowd]]
+== Atlassian Crowd Support
+
+{inrmonly}
+
+http://www.atlassian.com/software/crowd/[Atlassian Crowd] is a single sign-on and identity management product that
```
joedragons: Minor: URL (currently) redirects to HTTPS.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+applications. {nxrm} contains a security realm that allows you to configure the repository manager to authenticate
+against an Atlassian Crowd instance.
+
+The following steps are necessary to configure Crowd-based authentication:
+
+. <<crowd-nexus-prepare,Prepare Nexus>>
+. <<crowd-application-prepare,Prepare Atlassian Crowd>>
+. <<crowd-configuration,Configure the Nexus Crowd Connection>>
+. <<crowd-sect-mapping,Configure Nexus Crowd Security>>
+. <<crowd-realm,Activate the Nexus Crowd Realm>>
```
joedragons: Minor: While true, this seems very NX2y indexy style.  I think people who want to set this up can use the larger index or just scroll thru the single page (or do a find).  I don't feel we do this style anywhere else in our NX3 docs.  Interestingly, even NX2 LDAP doesn't do this style.  Again don't feel strongly, but 2 cents, impression.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+[[crowd-nexus-prepare]]
+=== Prepare {pro} for Atlassian Crowd
+
+Atlassian Crowd support is an option feature preinstalled and ready to configure in {pro} 3.1+.
```
joedragons: I'm not sure what "an option feature" is so suggesting we strike the words "an option".  Sentence is still true IMO. (maybe minor if I'm ignorant of a term)
Minor: I also think you can remove the version since 3.1 is the first version of PRO.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+The following steps are necessary to configure Crowd-based authentication:
+
+. <<crowd-nexus-prepare,Prepare Nexus>>
+. <<crowd-application-prepare,Prepare Atlassian Crowd>>
+. <<crowd-configuration,Configure the Nexus Crowd Connection>>
+. <<crowd-sect-mapping,Configure Nexus Crowd Security>>
+. <<crowd-realm,Activate the Nexus Crowd Realm>>
+
+[[crowd-nexus-prepare]]
+=== Prepare {pro} for Atlassian Crowd
```
joedragons: Minor: I personally think considering the content of this sub-section (6.1), you could fold this into the introduction and remove this as a section.  As written you basically have a sub-section for the fact it exists and a warning about it.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+you already use LDAP support, we recommend adding your LDAP server as a Crowd directory accessible to the Crowd
+'nexus' application instead of using both LDAP and Crowd realms in the repository manager.
+
+[[crowd-application-prepare]]
+=== Prepare Atlassian Crowd
+
+[[crowd-compatibility]]
+==== Compatibility
+
+Always use the latest version of Crowd available at the time your version of {pro} was released. When upgrading to
```
joedragons: Minor: "released" may be the wrong word depending on what we're trying to convey (not 100% sure=)).  So if we release in July but they install in Sept I *feel* like we'd want them to be on Sept, not July.  If so, then I'd switch "released" to "installed" (or "upgraded", etc).  Minor tho.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+a newer Crowd server, carefully review the Crowd server release notes for REST API backwards compatibility issues.
+
+Crowd support in {pro} 2.1 and greater only works in Crowd versions (3.1+) that support the Crowd
```
joedragons: 2.1 typo (should be 3.1)?  However, see my above note about don't think we need NXRM versions at all in this section.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+NOTE: These instructions are a general guide to adding an application to Crowd. For current detailed 
+instructions, visit the https://confluence.atlassian.com/display/CROWD/Adding+an+Application[official
+Crowd documentation].
+
+To connect {pro} to Crowd, you will need to configure {pro} as an application in Crowd.
+
+. Login to Crowd as a user with administrative rights.
+. Click on the 'Applications' tab.
+. Click 'Add application', shown in <<fig-crowd-add-application>>, to display the form and create a new
```
joedragons: Minor: To me this sounded like 'Add application' was shown in the figure.  Suggest minor rework to "Click Add application to display a form, shown in Figure 6.1,"

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+application with the following values in the 'Details' tab of the Add Application form:
+
+* Application Type: Generic Application
+
+* Name: nexus
+
+* Description: (optional)
+
+. Choose a password for this application. Nexus will use this password
+to authenticate with the Crowd server. Confirm the password, then click the 'Next' button.
```
joedragons: Minor: I'm +0 to having this as bullet #4 rather than the 4th bullet in bullet #3.  I feel like the only reason it's a distinct bullet is because we're explaining why Nexus needs the password which I am also +0 too.

If this stays as is, we use the term "Nexus" here (which as mentioned above, I feel is dated unless this guide is meant to infer it'll share crowd responsibilities with IQ etc).

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+[[fig-crowd-add-application]]
+.Adding a New Application
```
joedragons: Minor: Suggest "Adding a New Crowd Application" just to be super specific.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+image::figs/web/crowd-add-application.png[scale=50]
+
+Click 'Next' to advance to the 'Connection' breadcrumb. On this screen you need to supply the URL of your 
+application instance and the remote IP address for {pro}. It shows the Connection form
```
joedragons: "It shows the Connection form..." - I'm not sure what "It" is in this sentence.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+Click 'Next' to advance to the 'Connection' breadcrumb. On this screen you need to supply the URL of your 
+application instance and the remote IP address for {pro}. It shows the Connection form 
+configured for a local instance of {pro}. If you were configuring Crowd and {pro} in a production environment, 
+you would supply the URL that users would use to load the repository manager user interface in a web browser and 
+you would supply the IP address that the repository manager will be connecting from.  Once you have completed the 
+'Connection' form, click on 'Next' to advance to the 'Directories' form.
+
+The 'Directories' form allows you to select the user directory used for {pro} authentication. In this example, 
+the default 'User Management' directory will be used.
```
joedragons: Minor: Up to now, I'm not convinced this second sentence has any value.  If we're providing guidance we should suggest what they should use.  More likely to me seems to be they know what they want to do and we shouldn't make any suggestions and the second sentence would be more valuable like "Please utilize a configuration that makes sense to your setup." or nothing at all.

It is possible that utilizing the default will be mentioned later (but I am suspect of that based on my Crowd work)...wanted to mention here while I thought of this.  If it makes sense later, I'll come back and delete this comment.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+Clicking on the 'Next' button in the 'Directories' form advances to the 'Authorisation' form. If any of the 
+directories selected in the previous form contain groups, each group is displayed on this form next to a 
+checkbox. You can select 'Allow all users' for a directory or you can select specific groups that are allowed to 
+authenticate to {pro} via Crowd. This option would be used if you wanted to limit repository manager access to 
+specific subgroups within a larger Crowd directory. If your entire organization is stored in a single Crowd 
+directory, you may want to limit repository manager access to a group that contains only developers and 
+administrators.
```
joedragons: Minor: Seems cutoff to me.  I'm assuming clicking next and then finish are intentionally omitted otherwise, suggest they be added for completeness of instruction.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+[[crowd-configuration]]
+=== Configure {pro} Crowd Integration
+
+
+[[crowd-ssl]]
+==== Configure {pro} to Trust Crowd’s Secure URL (Optional)
+
+Although optional, we advise the connection from {pro} to your Crowd server to use the HTTPS protocol.
+
+If the Crowd Server certificate is not signed by a public certificate authority, you may have to explicitly trust
```
joedragons: Minor: If "Crowd Server certificate" is not a thing, we say "Crowd server" just before and after this sentence so the capitalization of "server" is inconsistent.  We also say "Crowd" without server earlier in this document, so you could also say "Crowd certificate" (if true).

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+the server certificate using <<ssl,SSL support>>. A common symptom observed are +peer not authenticated+
```
joedragons: Minor: If there's an anchor, I think "5.9.1 Outbound SSL - Trusting SSL Certificates of Remote Repositories" would be slightly better.  I thought this was a mis-link til I read more closely.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+messages, when trying to connect to the Crowd server.
```
joedragons: Minor: Suggest "...to an **untrusted** Crowd server."

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+. <<crowd-capability,Enable the Crowd Capability>>
+. <<crowd-ssl-trust,Add the Crowd server certificate to the truststore>>
+. <<crowd-config-connection,Configure Crowd Connection URL using
+the HTTPS url>>
+
+[[crowd-capability]]
+===== Enable the Crowd Capability
+
+To enable 'Crowd' go to 'Capabilities' located under 'System' in the 'Administration' menu. Perform the following 
+steps:
```
joedragons: Minor: IMO, this is ordering adds unnecessary complexity because the creation (not configuration) is done when you perform the "Configure Crowd Connection URL using the HTTPS url".  It will work this way however but basically forces you to do both of this and that step at once due to fields being required.

In fact, while I assume we have this entire section ("Configure {pro} to Trust Crowd’s Secure URL (Optional)") before "Configure Crowd Connection URL using the HTTPS url" because we want people to pay attention and do it, it really flows better as a post-cursor IMO.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+. click the 'Create capability' button to get to the 'Select Capability Type' table
+. select 'Capabilities' to open the 'Capabilities' panel located in the 'Administration' menu
+. select 'Crowd' to open the 'Create Crowd Capability' panel
```
joedragons: Minor: These 3 bullets start in lower case letters while the 4th starts capped.  Not sure there's any rule on that but it looks weird to me.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+. Click the 'Add' button in the panel toolbar. Select 'SSL: Crowd' in the 'Type' field. Make sure the 'Enabled' 
+checkbox is checked, and click the 'Save' button.
```
joedragons: This bullet needs rewritten it seems to be written against NX2.  Basically all you need to do here is check "Use the Nexus truststore".  However, as mentioned above re:ordering, you have to fill out all the fields you'd fill out when configuring Crowd as well because they're required before you can Save.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+[[crowd-ssl-trust]]
+===== Adding the Crowd Server Certificate to the Truststore
+
+In order to add the server certificate of your Crowd server to the truststore, go to 'SSL Certificates', located 
+under 'Security' in the 'Administration' menu. In the 'SSL Certificates' panel click the 'Load Certificate' 
+button, prompts a dropdown menu with two options:
```
joedragons: This sentence structure seems off.  Couple ways to make it right I think:
* ", which prompts a dropdown menu with two options:"
* "; this prompts a dropdown menu with two options:"
* "In the 'SSL Certificates' panel clicking the 'Load Certificate' button prompts a dropdown menu with two options:"

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+. 'Load from server', where you can enter the full +https://+ URL from the Crowd server
+. 'Paste PEM', where you can enter an encoded, remote certificate generated from Crowd
```
joedragons: Comma drama: I personally would prefer : after the buttons rather than commas.  That is: "Load from server: where you can..." etc

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+NOTE: Read more about centralizing <<ssl-certificates,ssl certificates>> to the {nxrm} in the Security chapter.
+
+[[crowd-config-connection]]
+==== Configure Nexus Crowd Connection
+
+The Crowd Configuration screen displayed in <<fig-manage-crowd-config>> can be accessed by users with 
+administrative privileges in {pro} by selecting 'Atlassian Crowd' in the 'Security' section of the
```
joedragons: Minor: In case you want to be specific about privilege 'crowd-all' and 'crowd-update' (currently effectively the same) will allow edits made in this section.  'crowd-read' will allow you to see it, which is how it's phrased (but won't help get thru this section).  I am not sure how we feel about directing at certain permissions (not just a question here), but thought I'd mention since saying "admin" seemed broad.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+NOTE: Read more about centralizing <<ssl-certificates,ssl certificates>> to the {nxrm} in the Security chapter.
+
+[[crowd-config-connection]]
+==== Configure Nexus Crowd Connection
+
+The Crowd Configuration screen displayed in <<fig-manage-crowd-config>> can be accessed by users with 
+administrative privileges in {pro} by selecting 'Atlassian Crowd' in the 'Security' section of the 
+'Administration' menu.
+
+From this screen you can access your Crowd roles through a running instance of {nxrm}. To manage a Crowd
```
joedragons: Roles are actually done via Security-Roles unless I misunderstand what is trying to be said here.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+The Crowd Configuration screen displayed in <<fig-manage-crowd-config>> can be accessed by users with 
+administrative privileges in {pro} by selecting 'Atlassian Crowd' in the 'Security' section of the 
+'Administration' menu.
+
+From this screen you can access your Crowd roles through a running instance of {nxrm}. To manage a Crowd 
+server connection within the repository manager you can activate the following:
+
+. Check the 'Enable Crowd' box to activate the capability
+. Check 'Enable Crowd Realm for Authentication' to allow prioritization of you available security realms
```
joedragons: typo "your" (says you)

NOTE: Doing this also enables the realm, making the final section unneeded.  I will comment there as well.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+Application Password:: This field contains the application password of a Crowd application. This value should 
+match the value in the Password field of the form.
+
+Crowd Server URL:: This is the URL used to connect to the Crowd Server.  Both 'http://' and 'https://' URLs are 
+accepted. You may need to <<crowd-ssl-trust,trust the crowd server certificate>> if a 'https://' URL is used.
+
+HTTP Timeout:: The HTTP Timeout specifies the number of milliseconds the repository manager will wait for a
+response from Crowd. A value of zero indicates that there is no timeout limit. Leave the field blank to use the
+default HTTP timeout.
```
joedragons: 3 of 4 of these fields do not match the field names in the application.
3 of 4 of these fields are not in order as they are in the screen/application.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+. Check 'Enable Crowd Realm for Authentication' to allow prioritization of you available security realms
+
+[[fig-manage-crowd-config]]
+.Crowd Configuration Panel
+image::figs/web/manage-crowd-config.png[scale=50]
+
+This panel contains the following fields:
+
+Application Name:: This field contains the application name of a Crowd application. This value should match the 
+value in the Name field of the form.
```
joedragons: Minor: I assume "form" refers to the form used to configure Crowd.  If correct, it may be worth being more specific here and saying something like "...of the Crowd form from the Details section when adding the application to your Crowd server" ... or something=)

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+Application Password:: This field contains the application password of a Crowd application. This value should 
+match the value in the Password field of the form.
```
joedragons: (same as last comment) repasting = "...of the Crowd form from the Details section when adding the application to your Crowd server"

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+Crowd Server URL:: This is the URL used to connect to the Crowd Server.  Both 'http://' and 'https://' URLs are 
+accepted. You may need to <<crowd-ssl-trust,trust the crowd server certificate>> if a 'https://' URL is used.
+
+HTTP Timeout:: The HTTP Timeout specifies the number of milliseconds the repository manager will wait for a
+response from Crowd. A value of zero indicates that there is no timeout limit. Leave the field blank to use the
+default HTTP timeout.
+
+You can use the 'Verify Connection' button to validate if your connection to Crowd is working. Once you have a 
+working connection, press 'Save' to confirm the configuration. Use 'Cancel' to abort saving any changes.
```
joedragons: Interestingly, you can only discard to not save, however it doesn't actually abort when you have no info (on creation), it basically resets.  It aborts editing.  Not really sure what to suggest there that's brief, but Cancel doesn't appear anywhere that I see (that is different than many of our other forms).

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+[[crowd-sect-mapping]]
+=== Configure {pro} Crowd Security
+
+There are two approaches available to manage what privileges a Crowd user has when they login to {pro}.
+
+. Mapping Crowd Groups to Roles
+. Mapping Crowd Users to Roles
```
joedragons: Minor: Suggest any remaining indexy stuff be consistently linked (like the first two) or not linked (like this one).  As far as indexy, I like this the best of the 3 because you may not perform both these (as opposed to the above 2 which are steps) however, if removed makes my first comment obsolete=)

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+NOTE: Mapping Crowd Groups to {pro} Roles is preferable because there is less configuration is involved overall in
```
joedragons: Unnecessary word(s) here, I think.  Consider:
"...because there is less configuration involved..." vs "because less configuration is involved..." to fix

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+{pro} and assigning users to Crowd groups can be centrally managed inside of Crowd by your security team after the
+initial repository manager setup.
+
+[[crowd-sect-mapping-group]]
+==== Mapping a Crowd Group to Roles
+
+When mapping a Crowd group to a {pro} role, you are specifying the permissions ( via roles ) that users within the
```
joedragons: Minor: Spaces in parenthetical are weird to me "( via roles )" vs "(via roles)".

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+Crowd group will have after they authenticate.
+
+To map a Crowd group to a {pro} role, open the 'Roles' panel by clicking on the 'Roles' link under 'Security'
+in the 'Administration panel. Click on 'Create role' button, select 'External Role Mapping', then click 'Crowd'.
```
joedragons: Close ' after Administration is missing, and yet somehow (in the pdf) the italics correct themself after role'.  Not sure how that happened, may be worse on the website OR you may have a very deep extra ' (or maybe just smart app).

Minor: this also made me realize that above it's refered to as the "'Administration' menu" (not panel).  Not sure if it's inter-exchangable...menu seemed more familiar to me but not sure I would have noticed if not for the weird italics=)

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+==== Mapping a Crowd Group to Roles
+
+When mapping a Crowd group to a {pro} role, you are specifying the permissions ( via roles ) that users within the
+Crowd group will have after they authenticate.
+
+To map a Crowd group to a {pro} role, open the 'Roles' panel by clicking on the 'Roles' link under 'Security'
+in the 'Administration panel. Click on 'Create role' button, select 'External Role Mapping', then click 'Crowd'. 
+This will take you 'Create Role' panel, as mentioned in <<roles>>.
+
+After choosing the 'Crowd' realm, the 'Role' drop-down should list all the Crowd groups the 'demo' Crowd
```
joedragons: So if you're going to refer to your example as demo, I strongly suggest you change:
"Click Add application, shown in Figure 6.1, to display the form and create a new application with the following values in the Details tab of the Add Application form:
• Application Type: Generic Application
• Name: nexus" to demo as well (note in case you want to do the reverse, the screenshot shown in the figure says demo as the name).
However, I recommend making this generic and just removing 'demo' from here.  My original writeup was saying if you ever called the example demo I missed it...which I don't think you do but it's in the fig (as noted).
Let me know if this is unclear.  I'm referencing two above sections and not sure how well=)

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+When mapping a Crowd group to a {pro} role, you are specifying the permissions ( via roles ) that users within the
+Crowd group will have after they authenticate.
+
+To map a Crowd group to a {pro} role, open the 'Roles' panel by clicking on the 'Roles' link under 'Security'
+in the 'Administration panel. Click on 'Create role' button, select 'External Role Mapping', then click 'Crowd'. 
+This will take you 'Create Role' panel, as mentioned in <<roles>>.
+
+After choosing the 'Crowd' realm, the 'Role' drop-down should list all the Crowd groups the 'demo' Crowd 
+application has access to. Select the group you would like to map in the 'Role' field and click 'Create Mapping'.
```
joedragons: "and click 'Create Mapping'" can be struck.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+NOTE: If you have two or more groups in Crowd accessible to the 'demo' application with the same name but in
```
joedragons: ('demo' vs 'nexus' vs generic again)

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+different directories, the repository manager will only list the first one that Crowd finds. Therefore, Crowd
+administrators should avoid identically named groups in Crowd directories.
+
+Before saving the group-to-role mapping, *you 'must' add at least one {pro} role to the mapped group*. After you
```
joedragons: "role" can be "role or privilege" unless intentionally written thusly.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+have added the roles using the 'Add' button, click the 'Save' button.
```
joedragons: Minor: Suggest 'Add to Selected' rather than 'Add' if using literal markup (never been clear how to refer to those arrows.
(I suspect 'Add' might be hangover from NX2 but suggestion made as minor since I'm not sure)

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+Saved mappings will appear in the list of roles with a mapping value of 'Crowd'.
```
joedragons: Minor: Note this is currently untrue because of https://issues.sonatype.org/browse/NEXUS-9444.  As usual, I defer if we should keep bugged areas as they should be or as they are.

If we leave this in, I do however (minorly) suggest switching "a mapped value" to "a source value" since that's what the field header is (currently) called.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+[[fig-crowd-view-user-groups]]
+.Crowd Groups for User "johnsmith"
+image::figs/web/crowd-view-user-groups.png[scale=45]
+
+To add an external user, open the 'Users' panel in the repository manager by clicking 'Users' in the 'Security' 
+section of the sidebar menu.
+
+Click the 'Source' dropdown button and select 'Crowd'. This action will display a table where you can locate a 
+user by Crowd user ID.
```
joedragons: Guessing the table is leftover from NX2.  NX3 now just lets you fill in the filter.  The filter is there whether you select Crowd or not, so be careful on (re)wording here.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+Typing the Crowd user id - for example +johnsmith+ - into the 'ID' field and clicking the magnifying 
+glass icon, will prompt the repository manager to search for a user ID +johnsmith+ from the Crowd realm.
+
+When the name you entered appears, click the ID to grant roles to the Crowd user. This will take you
```
joedragons: Can change "ID" for "row".  Borderline minor.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+
+Click the 'Source' dropdown button and select 'Crowd'. This action will display a table where you can locate a 
+user by Crowd user ID.
+
+Typing the Crowd user id - for example +johnsmith+ - into the 'ID' field and clicking the magnifying 
+glass icon, will prompt the repository manager to search for a user ID +johnsmith+ from the Crowd realm.
+
+When the name you entered appears, click the ID to grant roles to the Crowd user. This will take you 
+to a form where you can assign available roles. *You must map at least one role
+to the Crowd managed user* in order to 'Save'. The external Crowd User Example displays the s+johnsmith+
```
joedragons: typo "+johnsmith+" (currently shows "s+johnsmith+" which seems to also be messing up markup)

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,233 @@
...
+Crowd realm user as a member of the 'dev' Crowd group and the mapped role called 'Nexus Administrator Role'. 
+External groups like +dev+ are bolded in the 'Roles' table.
+
+[[crowd-realm]]
+=== Activate {pro} Crowd Realm
```
joedragons: As noted above, this entire section is not needed if you configure Crowd per the above steps.  It can be done independantly however but with "Enable Crowd Realm for authentication" controlling this, you'd need to have it off to see anything to do here (instructions above specifically have it checked) OR do this as step 2 rather than 5.  My suggestion is to remove this section and if you feel some note should be made mention in step 2 that checking that box basically does this (and refer to the realms section of the doc).

book-nexus.asciidoc
dulanism: Saving this edit to the end after all improvements are complete.

chapter-crowd.asciidoc
dulanism: Updated.

chapter-crowd.asciidoc
dulanism: Updated.

chapter-crowd.asciidoc
dulanism: ~option~, ~3.1~

chapter-crowd.asciidoc
dulanism: I'll talk to you about how to work this into docs after 2nd review.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+=== Prepare Atlassian Crowd
+
+[[crowd-compatibility]]
+==== Compatibility
+
+Always use the latest version of Crowd available at the time your version of {pro} was installed or upgraded. 
+If upgrading to a newer Crowd server, carefully review the Crowd server release notes for REST API backwards 
+compatibility issues.
+
+Crowd support in {pro} and greater only works in Crowd versions that support the Crowd REST API. Older versions
```
joedragons: "and greater" I think is leftover from when we had version#s and can be struck.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+* Click on the 'Applications' tab.
+* Click 'Add application' to display a form, shown in <<fig-crowd-add-application>>
+
+Next, create the new application with the following values in the 'Details' section:
+
+* Application Type: Generic Application
+
+* Name: demo
+
+* Description: (optional)
```
joedragons: Minor: The way this looks to me is you're telling them they should put "(optional)" as their desc.  I am not sure a nice way to phrase this otherwise tho...maybe in text (not bulleted)?

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+
+Choose a password for the application. The repository manager will use this password to authenticate with the 
+Crowd server. Confirm the password, then click the 'Next' button.
+
+[[fig-crowd-add-application]]
+.Adding a New Crowd Application
+image::figs/web/crowd-add-application.png[scale=50]
+
+Click 'Next' to advance to the 'Connection' breadcrumb. On this screen you need to supply the URL of your
```
joedragons: Before the image, you also said "then click the 'Next' button" so here is repetitive (and potentially confusing if they try and follow instructions literally).  Easiest rephrase would probably be "Having clicked 'Next'" but don't feel strongly to that rephrase.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+
+Choose a password for the application. The repository manager will use this password to authenticate with the 
+Crowd server. Confirm the password, then click the 'Next' button.
+
+[[fig-crowd-add-application]]
+.Adding a New Crowd Application
+image::figs/web/crowd-add-application.png[scale=50]
+
+Click 'Next' to advance to the 'Connection' breadcrumb. On this screen you need to supply the URL of your 
+application instance and the remote IP address for {pro}. The form shows the 'Connection' configured for a local
```
joedragons: "The form shows..." - I commented on this last time and it's changed but I'm still not sure what this sentence is saying.  I suspect it's telling what 'Connection' should be but the sentence prior seems to do that.  Hard to give advice, so just relaying I find this sentence confusing (still).

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+Choose a password for the application. The repository manager will use this password to authenticate with the 
+Crowd server. Confirm the password, then click the 'Next' button.
+
+[[fig-crowd-add-application]]
+.Adding a New Crowd Application
+image::figs/web/crowd-add-application.png[scale=50]
+
+Click 'Next' to advance to the 'Connection' breadcrumb. On this screen you need to supply the URL of your 
+application instance and the remote IP address for {pro}. The form shows the 'Connection' configured for a local 
+instance of {pro}. If you were configuring Crowd and {pro} in a production environment, you would supply the URL
```
joedragons: "If you were configuring Crowd and {pro} in a production environment" - this reads weirdly since I think they are likely doing that and we don't give an alternative of what to do if they are not doing this.  I feel like this portion of the sentence could be cut or clarified.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+===== Enable the Crowd Capability
+
+To enable 'Crowd' locate the 'Create Crowd Capability' form to activate the connection between the external 
+server with the repository manager. Go to 'Capabilities' located under 'System' in the 'Administration' menu. 
+Perform the following steps:
+
+* Select 'Capabilities' to open the 'Capabilities' panel, located in the 'Administration' menu
+* Click the 'Create capability' button to get to the 'Select Capability Type' table
+* Select 'Crowd' to open the 'Create Crowd Capability' panel
+* Complete the form by checking the 'Enable this capability', entering the 'Crowd Server URL', and identifying
```
joedragons: Minor: Enable is checked by default so shouldn't need to say it here unless being thorough/cautious.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+* 'Load from server': where you can enter the full +https://+ URL from the Crowd server
+* 'Paste PEM': where you can enter an encoded, remote certificate generated from Crowd
+
+NOTE: Read more about centralizing <<ssl-certificates,ssl certificates>> to the {nxrm} in the Security chapter.
+
+[[crowd-config-connection]]
+==== Configure Crowd Connection
+
+The Crowd Configuration screen displayed in <<fig-manage-crowd-config>> can be accessed by users with 
+administrative privileges in {pro} by selecting 'Atlassian Crowd' in the 'Security' section of the
```
joedragons: @dulanism RE "can be accessed by users with administrative privileges": I believe in a previous comment you said you'd ping me regarding this, so I'm assuming unchanged you meant it'd be a bigger effort or you decided you're ok as is.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+NOTE: Read more about centralizing <<ssl-certificates,ssl certificates>> to the {nxrm} in the Security chapter.
+
+[[crowd-config-connection]]
+==== Configure Crowd Connection
+
+The Crowd Configuration screen displayed in <<fig-manage-crowd-config>> can be accessed by users with 
+administrative privileges in {pro} by selecting 'Atlassian Crowd' in the 'Security' section of the 
+'Administration' menu.
+
+From this screen you can access your Crowd roles through a running instance of the repository manager. To create
```
joedragons: "From this screen..." - I don't think this sentence is true.  You access roles via Security-Roles.  I feel like I said this before so sorry if you replied and I forgot the explaination.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+
+* Check the 'Enable Crowd' box to activate the capability
+* Check 'Enable Crowd Realm for Authentication' to prioritize your available security realms
+
+[[fig-manage-crowd-config]]
+.Crowd Configuration Panel
+image::figs/web/manage-crowd-config.png[scale=50]
+
+In this form copy the respective values you attributed to the 'Details' section of the Crowd menu you created in 
+<<crowd-setup>>. The fields are as follows:
```
joedragons: So, I think my comments last time were unclear, the capability and this screen input the exact same thing.  So if you setup the capability above, everything will be filled in.  They will need to do nothing here.  I suggest reviewing the entire sub-section for revamp with this in mind.  It may not even be necessary to have at all.  The only thing of non-redundant value it provides is what the fields are in more detail (which could be moved) and the fact that you can get to this section at all (which could also be moved, tho more awkwardly).

You cannot (currently) "Enable Crowd Realm for authentication" from the capability screen, however, you can do the work from the Realms screen.

Basically this whole page is redundant (intentionally).

chapter-crowd.asciidoc
dulanism: Yes, wanted to touch base with you about this @joedragons

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+You can use the 'Verify Connection' button to confirm your connection to Crowd is working. Once you have a 
+working connection, press 'Save'.
+
+[[crowd-sect-mapping]]
+=== Configure {pro} Crowd Security
+
+There are two approaches available to manage what privileges a Crowd user has when they login to {pro}. You can 
+map Crowd groups to roles or map Crowd users to roles.
+
+NOTE: Mapping Crowd Groups to {pro} Roles is preferable because there is less configuration involved overall in
```
joedragons: Minor: Maybe tip instead of note?  Seems logical to me the way this is written.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+initial repository manager setup.
+
+[[crowd-sect-mapping-group]]
+==== Mapping a Crowd Group to Roles
+
+When mapping a Crowd group to a {pro} role, you are specifying the permissions (via roles) that users within the
+Crowd group will have after they authenticate.
+
+To map a Crowd group to a {pro} role, open the 'Roles' panel by clicking on the 'Roles' link under 'Security'
+in the 'Administration panel. Click on 'Create role' button, select 'External Role Mapping', then click 'Crowd'.
```
joedragons: I believe a closing ' after 'Administration is still missing.  See attached.
screen shot 2016-07-11 at 10 34 58 am

I am still surprised the markup works so well.  I would have expected reverse italic thru the entire doc.  Good fortune=)

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+==== Mapping a Crowd Group to Roles
+
+When mapping a Crowd group to a {pro} role, you are specifying the permissions (via roles) that users within the
+Crowd group will have after they authenticate.
+
+To map a Crowd group to a {pro} role, open the 'Roles' panel by clicking on the 'Roles' link under 'Security'
+in the 'Administration panel. Click on 'Create role' button, select 'External Role Mapping', then click 'Crowd'. 
+This will take you 'Create Role' panel, as mentioned in <<roles>>.
+
+After choosing the 'Crowd' realm, the 'Role' drop-down should list all the Crowd groups to which the generic
```
joedragons: Minor: Am not sure what the word "generic" here means.  Might be able to be struck.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+Crowd group will have after they authenticate.
+
+To map a Crowd group to a {pro} role, open the 'Roles' panel by clicking on the 'Roles' link under 'Security'
+in the 'Administration panel. Click on 'Create role' button, select 'External Role Mapping', then click 'Crowd'. 
+This will take you 'Create Role' panel, as mentioned in <<roles>>.
+
+After choosing the 'Crowd' realm, the 'Role' drop-down should list all the Crowd groups to which the generic 
+Crowd application has access. Select the group you would like to map in the 'Role' field.
+
+NOTE: If you have two or more groups in Crowd accessible to the 'General Application' ('demo') with the
```
joedragons: If this is supposed to be matching the Crowd field, I believe it's 'Generic Application' (or that's what shows in the figure).  If not, I don't understand the markup.
Minor: Either way, I'm not convinced about the markup around demo.  I'd personally prefer "(in this example, demo)" but I see what you are doing here.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+This will take you 'Create Role' panel, as mentioned in <<roles>>.
+
+After choosing the 'Crowd' realm, the 'Role' drop-down should list all the Crowd groups to which the generic 
+Crowd application has access. Select the group you would like to map in the 'Role' field.
+
+NOTE: If you have two or more groups in Crowd accessible to the 'General Application' ('demo') with the 
+same name but in different directories, the repository manager will only list the first one that Crowd finds. 
+Therefore, Crowd administrators should avoid identically named groups in Crowd directories.
+
+Before you save, *you 'must' add at least one {pro} role or privilege to the mapped group*. After you
```
joedragons: Minor: Since the fields are validated, I'm not convinced the bold is needed here.  If you disagree, you might just bold "must" (and remove italic, which supports the italic used for literals "rule" anyway).

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+
+Consider the Crowd server user with an id of +johnsmith+. In the Crowd administrative interface, the 
+user is a member of the +dev+ group, as shown in <<fig-crowd-view-user-groups>>.
+
+[[fig-crowd-view-user-groups]]
+.Crowd Groups for User "johnsmith"
+image::figs/web/crowd-view-user-groups.png[scale=45]
+
+To add an external user, open the 'Users' panel in the repository manager by clicking 'Users' in the 'Security' 
+section of the sidebar menu.
```
joedragons: Minor: While true, this does not mention 'Administration' like most(all?) of the other documentation references do.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+
+Click the 'Source' dropdown button and select 'Crowd'. To search for users from the Crowd realm you can either 
+enter an individual username within the filter box, or click the magnifying glass icon to generate the list of 
+all users from the Crowd realm.
+
+When the name you entered appears, click the row to grant roles to the Crowd user. This will take you
```
joedragons: "When the name you entered appears, click the row to grant roles to the Crowd user."
Minor: Suggest rewrite to "Click on the row of the name you desire to create the mapping for."
Reasoning: A) "When the name you entered" speaks to a specific one of the two methods discussed in the previous paragraph B) and genericizes the end result (clicking will not grant roles, it's a means to).

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+
+To add an external user, open the 'Users' panel in the repository manager by clicking 'Users' in the 'Security' 
+section of the sidebar menu.
+
+Click the 'Source' dropdown button and select 'Crowd'. To search for users from the Crowd realm you can either 
+enter an individual username within the filter box, or click the magnifying glass icon to generate the list of 
+all users from the Crowd realm.
+
+When the name you entered appears, click the row to grant roles to the Crowd user. This will take you 
+to a form where you can assign available roles. *You must map at least one role to the Crowd managed user* in
```
joedragons: I again feel the bold is harsh here (with validation in place/mind).  I would replace with instructions or reference on how to add.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,206 @@
...
+section of the sidebar menu.
+
+Click the 'Source' dropdown button and select 'Crowd'. To search for users from the Crowd realm you can either 
+enter an individual username within the filter box, or click the magnifying glass icon to generate the list of 
+all users from the Crowd realm.
+
+When the name you entered appears, click the row to grant roles to the Crowd user. This will take you 
+to a form where you can assign available roles. *You must map at least one role to the Crowd managed user* in 
+order to 'Save'. The external Crowd user example displays the +johnsmith+ Crowd realm user as a member of both
+'dev' and 'crowd-administrators' groups.
```
joedragons: While true, I do not understand how this last sentence plays into this section OR I think a figure is missing.  Alternatively, maybe it should be further up?

chapter-security.asciidoc
```diff
@@ -582,12 +581,12 @@ Inbound client communication includes
 
 {inall}
 
-When the SSL certificate of a remote proxy repository is not trusted, the repository may be automatically blocked or
-outbound requests fail with a message similar to 'PKIX path building failed'.
+When the SSL certificate of a remote proxy repository is not trusted, the repository may be automatically blocked 
+oroutbound requests fail with a message similar to 'PKIX path building failed'.
```
joedragons: Looks like 114izing, you are now missing a space here (says "oroutbound" previously [correctly?] was "or outbound").

chapter-crowd.asciidoc
dulanism: Okay, this might at best merit a 'NOTE'. In the capability setup section. Image thus may be unnecessary.

chapter-crowd.asciidoc
dulanism: 86ing it.

chapter-crowd.asciidoc
dulanism: Forgot. Consider this update: 

> From this screen you can access your roles from 'Roles' in 'Security' within the 'Administration' menu. To create a Crowd server connection within the repository manager you can activate the following:
* Check the 'Enable Crowd' box to activate the capability
* Check 'Enable Crowd Realm for Authentication' to prioritize your available security realms

joedragons: "From this screen you can access your roles from 'Roles' in 'Security' within the 'Administration' menu."
The second part of the revamp is true 'you can access your roles from 'Roles' in 'Security' within the 'Administration' menu.' but unless I'm confused it's the not "From this screen".
Also seems tangential (or misplaced) information, as this entire section is talking about creating a capability.
The only reason I can think of you'd need to know to access Roles here is if you were trying to assign yourself permission to access Crowd (or Capabilities) and that's probably a redirect/link to the security section.

There is also a comment in the paragraph just above this one about potentially revamping to explain privileges needed to get here so maybe this sentence is just in the wrong place (and should be tied to that paragraph).

chapter-crowd.asciidoc
dulanism: Striking the 'Configure Crowd Connection' section altogether. Summarized in a final paragraph in the 'Enable... Crowd Capability' section. Consider this update: 

> After you enable Crowd, you can see the resolved 'Crowd server URL', 'Crowd application name', and 'Crowd application password' automatically filled out from panel located in 'Atlassian Crowd', under 'Security'.

chapter-crowd.asciidoc
```diff
@@ -83,8 +79,8 @@ settings. Click 'Add application' to confirm the settings.
 ===== Enable the Crowd Capability
 
 To enable 'Crowd' locate the 'Create Crowd Capability' form to activate the connection between the external 
-server with the repository manager. Go to 'Capabilities' located under 'System' in the 'Administration' menu. 
-Perform the following steps:
+server with the repository manager. Crowd will be enable by default. Go to 'Capabilities' located under 'System'
```
joedragons: typo "enabled" (says "enable")

chapter-crowd.asciidoc
```diff
@@ -95,6 +91,20 @@ the 'Application Name' and 'Application Password' that corresponds to your Crowd
...
 want to manage Crowd with the HTTPS protocol, mentioned in <<crowd-ssl>>.
 
+After you enable Crowd, you can see the resolved 'Crowd server URL', 'Crowd application name', and 'Crowd 
+application password' automatically filled out from panel located in 'Atlassian Crowd', under 'Security'. Also, 
+on this form your can configure 'Connection timeout', where you can enter a value that specifies the number of 
+milliseconds the repository manager will wait for a response from Crowd. A value of zero indicates that there is 
+no timeout limit. Leave the field blank to use the default timeout.
+
+You can use the 'Verify Connection' button to confirm your connection to Crowd is working. Once you have a 
+working connection, press 'Save'. An example of the Crowd configuration in {pro} can be seen in
```
joedragons: Minor: Since if you followed instructions in order you should have a working connection with not necessarily a need to save, recommend rewording "Once you have a working connection, press 'Save'." to "Pressing 'Save' will save any changes made to the Crowd configuration." or some such.

chapter-crowd.asciidoc
dulanism: Cool. I'd already made that fix before the last commit 
https://github.com/sonatype/nexus-book/pull/195/commits/99ab1e3a651aded637f4ad6267081f9499cab152

chapter-crowd.asciidoc
dulanism: Revised to: 

> You can use the 'Verify Connection' button to confirm your connection to Crowd is working. Pressing 'Save' will save any changes made to the Crowd configuration, as shown in ...

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+
+To connect {pro} to Crowd, you will need to configure {pro} as an application in Crowd.
+
+* Login to Crowd as a user with administrative rights.
+* Click on the 'Applications' tab.
+* Click 'Add application' to display a form, shown in <<fig-crowd-add-application>>
+
+Next, create the new application with the following values in the 'Details' section:
+
+* Application Type (e.g. Generic Application)
```
joedragons: Minor: I know I asked this to be generalized but I'm not sure this is e.g.  I think you may need to pick this.  If you have other experiences, I defer to them (thus marked minor).

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+Click 'Next' to advance to the final screen, 'Confirmation', which gives you a summary of your Crowd server 
+settings. Click 'Add application' to confirm the settings.
+
+[[crowd-configuration]]
+=== Configure {pro} Crowd Integration
+
+[[crowd-capability]]
+===== Enable the Crowd Capability
+
+To enable 'Crowd' locate the 'Create Crowd Capability' form to activate the connection between the external
```
joedragons: Suggest removing "locate the 'Create Crowd Capability' form to".  The subsequent sentences tell exactly how to do this and removing these words makes it more a summary.  If you prefer this text here, I think some of the subsequent steps are redundant.

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+=== Configure {pro} Crowd Integration
+
+[[crowd-capability]]
+===== Enable the Crowd Capability
+
+To enable 'Crowd' locate the 'Create Crowd Capability' form to activate the connection between the external 
+server with the repository manager. Crowd will be enabled by default. Go to 'Capabilities' located under 'System' 
+in the 'Administration' menu. Perform the following steps:
+
+* Select 'Capabilities' to open the 'Capabilities' panel, located in the 'Administration' menu
```
joedragons: Minor: This potentially reads like the capabilities panel is located in the administration menu.  Suggest either removing ", located in the 'Administration' menu" or rephrasing like "Select 'Capabilities' from the 'Administration' menu to open the 'Capabilities' panel".

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+.Crowd Configuration Panel
+image::figs/web/manage-crowd-config.png[scale=50]
+
+[[crowd-ssl]]
+==== Configure {pro} to Trust Crowd’s Secure URL (Optional)
+
+Although optional, we advise the connection from {pro} to your Crowd server to use the HTTPS protocol.
+
+If the Crowd certificate is not signed by a public certificate authority, you may have to explicitly trust
+the server certificate as explained in <<ssl-proxy-repo>>. A common symptom observed are +peer not authenticated+
```
joedragons: Minor: I'm not 100% sure (thus minor) but I think "are" should be "is".  symptom (singular) is (also singular)?

chapter-crowd.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+
+NOTE: Read more about centralizing <<ssl-certificates,ssl certificates>> to the {nxrm} in the Security chapter.
+
+[[crowd-sect-mapping]]
+=== Configure {pro} Crowd Security
+
+There are two approaches available to manage what privileges a Crowd user has when they login to {pro}. You can 
+map Crowd groups to roles or map Crowd users to roles.
+
+TIP: Mapping Crowd Groups to {pro} Roles is preferred because there is less configuration involved overall in
```
joedragons: Minor: This is the only place I see where Groups is caps.

joedragons:
+1

chapter-crowd.asciidoc
dulanism: I use e.g. the in same way we present examples/describe the forms/fields for setting up a repository. Example (Bower):

> To proxy a Bower registry, you simply create a new bower (proxy) as documented in Section 4.3, “Repository Management” in details. Minimal configuration steps are:

* Define Name
* Define URL for Remote storage e.g. http://bower.herokuapp.com
* Select a Blob store for Storage

Addendum: I do see you point, though. I could easily just scrap the "(e.g. blah-blah)" altogether for this chapter. Thoughts?

chapter-crowd.asciidoc
dulanism: Make sense.

chapter-crowd.asciidoc
joedragons: I'm mostly concerned that we (I believe) want them to pick Generic Application but they can name whatever they like.  I think it's obvious to us but we also know how it works.  I'm not convinced there's any way for them to tell the difference between Application Type and Name.  I see what you mean tho; I don't feel strongly to change or leave as is.  My private hope is that people know how to setup Crowd and will blaze by this section anyway=)

PR #196, opened 2016-06-29
NEXUS-10404-book-declarations

dulanism:
Temporarily updated to nexus book version - From 3.0.0/3.0.0-03 to 3.0.1/3.0.1-01, ahead of release


joedragons:
+1

PR #198, opened 2016-07-11
NEXUS-10303 Port and adapt user token security chapter to 3.1

dulanism:
JIRA: https://issues.sonatype.org/browse/NEXUS-10303
CI: http://bamboo.s/browse/BOOK-NEXUS160

### DIFF
--- a/chapter-security.asciidoc
+++ b/chapter-security.asciidoc
@@ -533,15 +533,13 @@ group. More details are available in <<roles>>.
 ==== Introduction
 
 When using Apache Maven with {pro} or {oss}, the user credentials for accessing the repository manager have to be 
-stored in the user’s +settings.xml+ file. Like a +pom.xml+ your +settings.xml+ is file that contains your user 
-preferences. The file lives in +$HOME/.m2+, your main Maven directory. This file contains listings for 
+stored in the user’s `settings.xml` file. Like a `pom.xml` your `settings.xml` is file that contains your user 
+preferences. The Maven framework has the ability to encrypt passwords within the +setting.xml+, but the need for 
+it to be reversible in order to be used limits its security.
+
+The +settings.xml+ lives in `$HOME/.m2`, your main Maven directory. This file contains listings for 
 personalized client or build-tool configurations such as repositories. This file is not exclusive to 
 Maven-specific repositories.
-
-The general setup and use is cumbersome, and the potential need for regular changes due to strong security 
-requirements e.g., with regular, required password changes triggers the need for a simpler, more secure solution. 
-The Maven framework has the ability to encrypt passwords within the +setting.xml+, but the need for it to be 
-reversible in order to be used, limits its security.
 
 Other build systems use similar approaches and can benefit from the usage of user tokens as well. {pro}'s user 
 token feature establishes a two-part token for the user. Usage of the token acts as a substitute method for 
@@ -549,7 +547,7 @@ authentication that would normally require passing your username and password in
 
 This is especially useful for scenarios where single sign-on solutions like LDAP are used for authentication
 against the repository manager and other systems and the plain text username and password cannot be stored in the
-+settings.xml+ following security policies. In this scenario the generated user tokens can be used instead.
+`settings.xml` following security policies. In this scenario the generated user tokens can be used instead.
 
 [[enable-reset-token]]
 ==== Enabling and Resetting User Tokens
@@ -566,6 +564,11 @@ write access for deployments from a build execution or a manual upload, but the 
 Besides resetting all user tokens, you can reset the token of an individual user by selecting the 'User Token' 
 tab in the 'Users' administration from the 'Security' menu. The password requested for the action to proceed is 
 the password for the authenticated administrator who resets the token(s).
+
+////
+currently a bug 
+Above true assuming we fix https://issues.sonatype.org/browse/NEXUS-10305 first.
+////
 
 WARNING: Resetting user tokens forces users to update the `settings.xml` with the newly created tokens, and 
 could potentially break any command line builds using the tokens until this change is carried out. This also 
@@ -604,7 +607,7 @@ the user token. The dialog will close automatically after one minute or simply c
 The user code and pass code values can be used as replacements for username and password in the login dialog. It
 is also possible to use the original username and the pass code to log in to the user interface.
 
-Upon activation of user tokens for repository authentication you can access the repository manager with the user 
+Upon activation of user tokens for repository authentication you must access the repository manager with the user 
 token, from the command line. For example, your username-password credentials access with
 
 ----
@@ -617,8 +620,11 @@ has to be replaced with the usage of user code and pass code separated by colon 
 curl -v --user N+ZBiTlF:76xSi+HAQvYHZH8kgyJldWD7aJnPgCrHG/Zu7mkpWmZZ http://localhost:2468/repository/bower-all/
 ----
 
+////
+Commented out because uncertain this feature exists in the app yet
 User token values can be accessed as part of the Maven settings template feature automating updates as mentioned 
 in <<maven-sect-single-group>>.
+////
 
 [[remote-user-token]]
 === Authentication via Remote User Token
--- a/chapter-security.asciidoc
+++ b/chapter-security.asciidoc
@@ -523,26 +523,15 @@ After your LDAP the successful configuration of your connection and user and gro
 configure external role mappings. This allows you to define the repository manager specific security for a LDAP
 group. More details are available in <<roles>>.
 	
-[[usertoken]]
+[[user-token]]
 === Security Setup with User Tokens
-
-////
-rewrite of this section in progress
-////
 
 {inrmonly}
 
+[[user-token-intro]]
 ==== Introduction
 
-When using Apache Maven with {pro} or {oss}, the user credentials for accessing the repository manager have to be
-stored in clear text in the user's settings.xml file. Maven has the ability to encrypt passwords in +setting.xml+,
-but the need for it to be reversible in order to be used, limits its security. In addition, the general setup and
-use is cumbersome, and the potential need for regular changes due to strong security requirements e.g., with
-regular, required password changes triggers the need for a simpler and more secure solution.
-
-Other build systems use similar approaches and can benefit from the usage of User Token as well.
-
-The User Token feature of {pro} fills that need for Apache Maven as well as other build systems and users. It
+{pro}'s user token feature fills the need for Apache Maven as well as other build systems and users. It
 introduces a two-part token for the user, replacing the username and password with a user code and a pass code
 that allows no way of recovering the username and password from the user code and pass code values; yet can be
 used for authentication with the repository manager from the command line via Maven as well as in the UI.
@@ -554,46 +543,43 @@ against the repository manager and other systems and the plain text username and
 User token usage is integrated in the Maven settings template feature of {pro} documented in 'maven-settings' to
 further simplify its use.
 
+[[enable-reset-token]]
 ==== Enabling and Resetting User Tokens
 
-The user token-based authentication can be activated by an administrator or user with the role +usertoken-admin+
-or +usertoken-all+ by accessing the 'User Token' item in the 'Security' submenu on the left-hand main menu.
+User token-based authentication can be activated by an administrator or user by clicking the 'User Token' menu 
+item under 'Security' in the 'Administration' menu. Check the 'Enable user tokens' box, then press 'Save' to 
+activate the feature. You can access the user token from the 'Profile' mode toggle, explained in <<access-token>>.
 
-Once user token is 'Enabled' by activating the checkbox in the administration tab and pressing 'Save',  the 
-feature is activated and the additional section to Reset All User Tokens is available as well.
+Additionally, you can check the 'Require user tokens for repository authentication' box to allow the repository 
+manager to require a user token for any access to the repository and group content URLs. This affects read and 
+write access for deployments from a build execution or a manual upload, but the user interface will not change.
 
-Selecting the 'Protect Content' feature configures the repository manager to require a user token for any access
-to the content URLs that includes all repositories and groups. This affects read access as well as write access
-e.g., for deployments from a build execution or a manual upload.
-
-'Activating User Token' as a feature automatically adds the 'User Token Realm' as a 'Selected Realm' in the 
-'Security Settings' section and available in the 'Server' section of the left-hand 'Administration' menu. If 
-desired, you can reorder the security realms used, although the default settings with the 'User Token Realm' as a 
-first realm is probably the desired setup. This realm is not removed when the User Token feature is disabled; 
-however, it will cleanly pass through to the next realm and with the realm remaining any order changes stay 
-persisted in case the feature is reactivated at a later stage.
+When you activate the user token, the feature automatically adds the 'User Token Realm' to the 'Active Realms' 
+list. To see the results, go to 'Realms' located under 'Security' in the 'Administration' menu. If desired, you 
+can reorder the security realms used, although the default settings with the 'User Token Realm' as a first realm 
+is probably the desired setup. This realm is not removed when the User Token feature is disabled; however, it 
+will cleanly pass through to the next realm. The realm remain in active bin in your 'Realms' in case the feature 
+is reactivated at a later stage.
 
 Besides resetting all user tokens, an administrator can reset the token of an individual user by selecting the 
-'User Token' tab in the 'Users' administration from the 'Security' menu in the left-hand navigation. The password 
-requested for this action to proceed is the password for the currently logged in administrator resetting the 
-token(s).
+'User Token' tab in the 'Users' administration from the 'Security' menu. The password requested for the action to 
+proceed is the password for the authenticated administrator who resets the token(s).
 
 WARNING: Resetting user tokens forces the users to update the `settings.xml` with the newly created tokens and 
 potentially breaks any command line builds using the tokens until this change is carried out. This specifically 
 also applies to continuous integration servers using user tokens or any other automated build executions.
 
+[[access-token]]
 ==== Accessing and Using Your User Tokens
 
-With user token enabled, any user can access his/her individual tokens via their 'Profile' panel. To access the
-panel, select 'Profile' when clicking on the user name in the top right-hand corner of the user interface. Then
-select 'User Token' in the drop-down to get access to the 'User Token' screen in the 'Profile panel'.
+With user token enabled, you can access your individual tokens in your 'Accounts' panel. To access the panel, 
+select the 'Profile' menu toggle, on the top right-hand corner of the user interface.
 
 In order to be able to see this 'User Token' panel the user has to have the +usertoken-basic+ role or the 
 +usertoken:current+ privilege. To access or reset the token you have to press the respective button in the panel
 and then provide your username and password in the dialog.
 
-Resetting the token will show and automatically hide a dialog with a success message and accessing the token will 
-show the dialog.
+Resetting the token will show a dialog with a success message and accessing the token will show the dialog.
 
 The User Token dialog displays the user code and pass code tokens in separate fields in the top level section as
 well as a server section ready to be used in a Maven settings.xml file. When using the server section you simply
--- a/chapter-security.asciidoc
+++ b/chapter-security.asciidoc
@@ -532,8 +532,20 @@ group. More details are available in <<roles>>.
 [[user-token-intro]]
 ==== Introduction
 
-{pro}'s user token feature establishes a two-part token for the user. Usage of the token acts as a substitute 
-method for authentication that would normally require passing your username and password in plain text.
+When using Apache Maven with {pro} or {oss}, the user credentials for accessing the repository manager have to be 
+stored in the user’s +settings.xml+ file. Like a +pom.xml+ your +settings.xml+ is file that contains your user 
+preferences. The file lives in +$HOME/.m2+, your main Maven directory. This file contains listings for 
+personalized client or build-tool configurations such as repositories. This file is not exclusive to 
+Maven-specific repositories.
+
+The general setup and use is cumbersome, and the potential need for regular changes due to strong security 
+requirements e.g., with regular, required password changes triggers the need for a simpler, more secure solution. 
+The Maven framework has the ability to encrypt passwords within the +setting.xml+, but the need for it to be 
+reversible in order to be used, limits its security.
+
+Other build systems use similar approaches and can benefit from the usage of user tokens as well. {pro}'s user 
+token feature establishes a two-part token for the user. Usage of the token acts as a substitute method for 
+authentication that would normally require passing your username and password in plain text.
 
 This is especially useful for scenarios where single sign-on solutions like LDAP are used for authentication
 against the repository manager and other systems and the plain text username and password cannot be stored in the
@@ -542,9 +554,10 @@ against the repository manager and other systems and the plain text username and
 [[enable-reset-token]]
 ==== Enabling and Resetting User Tokens
 
-User token-based authentication can be activated by an administrator or user by clicking the 'User Token' menu 
-item under 'Security' in the 'Administration' menu. Check the 'Enable user tokens' box, then press 'Save' to 
-activate the feature. You can access the user token from the 'Profile' mode toggle, explained in <<access-token>>.
+User token-based authentication can be activated by an administrator or user with the +nx-usertoken-settings+ 
+privilege. If given that privilege click the 'User Token' menu item under 'Security' in the 'Administration' 
+menu. Check the 'Enable user tokens' box, then press 'Save' to activate the feature. You can access the user 
+token from the 'Profile' mode toggle, explained in <<access-token>>.
 
 Additionally, you can check the 'Require user tokens for repository authentication' box to allow the repository 
 manager to require a user token for any access to the repository and group content URLs. This affects read and 
@@ -554,25 +567,25 @@ Besides resetting all user tokens, you can reset the token of an individual user
 tab in the 'Users' administration from the 'Security' menu. The password requested for the action to proceed is 
 the password for the authenticated administrator who resets the token(s).
 
-WARNING: Resetting user tokens forces user to update the `settings.xml` with the newly created tokens, and 
+WARNING: Resetting user tokens forces users to update the `settings.xml` with the newly created tokens, and 
 could potentially break any command line builds using the tokens until this change is carried out. This also 
 applies to continuous integration servers using user tokens or any other automated build executions.
 
 [[user-token-realms]]
 ==== Accessing User Tokens in Realms
 
-When you activate the user token, the feature automatically adds the 'User Token Realm' to the 'Active Realms' 
+When you activate user tokens, the feature automatically adds the 'User Token Realm' to the 'Active Realms' 
 list. To see the results, go to 'Realms' located under 'Security' in the 'Administration' menu. If desired, you 
 can re-order the security realms used, although the default settings with the 'User Token Realm' as the first 
-realm is probably the desired setup. This realm is not removed when the User Token feature is disabled; however, 
+realm is probably the desired setup. This realm is not removed when the user tokens are disabled; however, 
 it will cleanly pass through to the next realm. The realm will remain in the active bin in your 'Realms' in case 
 the feature is reactivated at a later stage.
 
 [[access-token]]
-==== Accessing and Using Your User Tokens
+==== Accessing and Using Your User Token
 
-With user token enabled, you can access your individual tokens on your 'Accounts' panel. To access the panel, 
-select the 'Profile' menu toggle, on the top right-hand corner of the user interface.
+With enabled, you can access your individual token on your 'Account' panel. To access the panel, select the 
+'admin' mode toggle, on the top right-hand corner of the user interface.
 
 In order to see the 'User Token' click the 'Access user token' button. This will prompt the 'Authenticate' dialog 
 where you are required to re-enter your credentials. After clicking 'Authenticate' in the completed dialog, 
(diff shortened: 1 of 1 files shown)


dulanism:
Also: 
* link to this chapter https://github.com/sonatype/nexus-book/pull/199#discussion_r71319061
* Ported chapter highlights Maven in intro but this feature should work for all formats

chapter-security.asciidoc
```diff
@@ -496,8 +497,8 @@ example value is 'uniqueMember'.
 Group member format:: This field captures the format of the Group Member Attribute, and is used by the repository 
 manager to extract a username from this attribute. An example values is `${dn}`.
 
-If your installation does not use static groups, you can configure the LDAP connection to refer to an attribute on 
-the user entry to derive group membership. To do this, select 'Dynamic Groups' in the 'Group type' drop down.
+If your installation does not use static groups, you can configure the LDAP connection to refer to an attribute 
+on the user entry to derive group membership. To do this, select 'Dynamic Groups' in the 'Group type' drop down.
```
bentmann: Ideally, all those reformatting changes are made on master and not mixed into a PR.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+
+[[user-token-intro]]
+==== Introduction
+
+{pro}'s user token feature establishes a two-part token for the user. Usage of the token acts as a substitute 
+method for authentication that would normally require passing your username and password in plain text.
+
+This is especially useful for scenarios where single sign-on solutions like LDAP are used for authentication
+against the repository manager and other systems and the plain text username and password cannot be stored in the
++settings.xml+ following security policies. In this scenario the generated user tokens can be used instead.
```
bentmann: What's a `settings.xml`? Unlike the NX2 docs, the NX3 docs fail to introduce/mention Apache Maven and hence lack context to resolve that filename. If we don't want to port over the Maven-specific intro/motiviation from the NX2 docs, we might want to drop the reference to the Maven-specific config file and use some more abstract lingo like "stored on clients" or "stored in build tool configurations".

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+
+[[enable-reset-token]]
+==== Enabling and Resetting User Tokens
+
+User token-based authentication can be activated by an administrator or user by clicking the 'User Token' menu
```
bentmann: The NX2 docs clarify "or user with (fancy role name)". Omitting that clarification and saying just "admin or user" is effectively saying "anybody" which is misleading.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
++settings.xml+ following security policies. In this scenario the generated user tokens can be used instead.
+
+[[enable-reset-token]]
+==== Enabling and Resetting User Tokens
+
+User token-based authentication can be activated by an administrator or user by clicking the 'User Token' menu 
+item under 'Security' in the 'Administration' menu. Check the 'Enable user tokens' box, then press 'Save' to 
+activate the feature. You can access the user token from the 'Profile' mode toggle, explained in <<access-token>>.
+
+Additionally, you can check the 'Require user tokens for repository authentication' box to allow the repository
```
bentmann: NX3 certainly offers that checkbox but from a code inspection I rather get the impression the functionality isn't implemented (cf. https://github.com/sonatype/nexus-internal/blob/5e34c8081fa48a0aeb5545c0c7f5aeb1d5d693a3/private/plugins/nexus-usertoken-plugin/src/main/java/com/sonatype/nexus/usertoken/plugin/internal/UserTokenServiceImpl.java#L398). Not sure what the plans here are, unless the feature is scheduled for 3.1 or I'm outright mistaken on its state, we might want to pull that paragraph (and at some point, the unimplemented UI option).

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+User token-based authentication can be activated by an administrator or user by clicking the 'User Token' menu 
+item under 'Security' in the 'Administration' menu. Check the 'Enable user tokens' box, then press 'Save' to 
+activate the feature. You can access the user token from the 'Profile' mode toggle, explained in <<access-token>>.
+
+Additionally, you can check the 'Require user tokens for repository authentication' box to allow the repository 
+manager to require a user token for any access to the repository and group content URLs. This affects read and 
+write access for deployments from a build execution or a manual upload, but the user interface will not change.
+
+Besides resetting all user tokens, you can reset the token of an individual user by selecting the 'User Token' 
+tab in the 'Users' administration from the 'Security' menu. The password requested for the action to proceed is
```
bentmann: ... assuming we fix https://issues.sonatype.org/browse/NEXUS-10305 first.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+
+Additionally, you can check the 'Require user tokens for repository authentication' box to allow the repository 
+manager to require a user token for any access to the repository and group content URLs. This affects read and 
+write access for deployments from a build execution or a manual upload, but the user interface will not change.
+
+Besides resetting all user tokens, you can reset the token of an individual user by selecting the 'User Token' 
+tab in the 'Users' administration from the 'Security' menu. The password requested for the action to proceed is 
+the password for the authenticated administrator who resets the token(s).
+
+WARNING: Resetting user tokens forces user to update the `settings.xml` with the newly created tokens, and
```
bentmann: "forces user**s**" (plural)

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+the password for the authenticated administrator who resets the token(s).
+
+WARNING: Resetting user tokens forces user to update the `settings.xml` with the newly created tokens, and 
+could potentially break any command line builds using the tokens until this change is carried out. This also 
+applies to continuous integration servers using user tokens or any other automated build executions.
+
+[[user-token-realms]]
+==== Accessing User Tokens in Realms
+
+When you activate the user token, the feature automatically adds the 'User Token Realm' to the 'Active Realms'
```
bentmann: The combination of "the user token" and "the feature" reads odd to me (single piece vs complete thing), "activate the user token feature" or "activate user token**s**" might be better.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+could potentially break any command line builds using the tokens until this change is carried out. This also 
+applies to continuous integration servers using user tokens or any other automated build executions.
+
+[[user-token-realms]]
+==== Accessing User Tokens in Realms
+
+When you activate the user token, the feature automatically adds the 'User Token Realm' to the 'Active Realms' 
+list. To see the results, go to 'Realms' located under 'Security' in the 'Administration' menu. If desired, you 
+can re-order the security realms used, although the default settings with the 'User Token Realm' as the first 
+realm is probably the desired setup. This realm is not removed when the User Token feature is disabled; however,
```
bentmann: "User Token feature", in line 535 it's all lower-case, which is it?

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+
+When you activate the user token, the feature automatically adds the 'User Token Realm' to the 'Active Realms' 
+list. To see the results, go to 'Realms' located under 'Security' in the 'Administration' menu. If desired, you 
+can re-order the security realms used, although the default settings with the 'User Token Realm' as the first 
+realm is probably the desired setup. This realm is not removed when the User Token feature is disabled; however, 
+it will cleanly pass through to the next realm. The realm will remain in the active bin in your 'Realms' in case 
+the feature is reactivated at a later stage.
+
+[[access-token]]
+==== Accessing and Using Your User Tokens
```
bentmann: Suggest to go with singular "Your User Token" to clarify that at any given time a given user has at most one user token.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+
+With user token enabled, you can access your individual tokens on your 'Accounts' panel. To access the panel,
```
bentmann: "user token**s** enabled"? Maybe I'm just not getting how you refer to the feature, to me "user token" is the single token an individual uses for access, not the feature itself.

bentmann: "your individual token~~s~~" (singular)

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+can re-order the security realms used, although the default settings with the 'User Token Realm' as the first 
+realm is probably the desired setup. This realm is not removed when the User Token feature is disabled; however, 
+it will cleanly pass through to the next realm. The realm will remain in the active bin in your 'Realms' in case 
+the feature is reactivated at a later stage.
+
+[[access-token]]
+==== Accessing and Using Your User Tokens
+
+With user token enabled, you can access your individual tokens on your 'Accounts' panel. To access the panel, 
+select the 'Profile' menu toggle, on the top right-hand corner of the user interface.
```
bentmann: The UI has changed in NX3 in this area and I suspect this hasn't been updated after the copy. E.g. I neither see anything reading "Profile" nor a menu toggle in the NX3 UI.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+
+In order to see the 'User Token' click the 'Access user token' button. This will prompt the 'Authenticate' dialog 
+where you are required to re-enter your credentials. After clicking 'Authenticate' in the completed dialog, 
+another dialog will appear with user token.
```
bentmann: "with **the/your** user token"

chapter-security.asciidoc
dulanism: Good point. Will expound.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+
+Below the 'Access your token' section is another section that allows you to reset your token. Click the 'Reset 
+user token' button, which prompts an 'Authenticate' dialog. Enter your credentials panel to complete the user
```
bentmann: "Enter your credentials ~~panel~~ to", or maybe "into the panel", but something there doesn't fit.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+With user token enabled, you can access your individual tokens on your 'Accounts' panel. To access the panel, 
+select the 'Profile' menu toggle, on the top right-hand corner of the user interface.
+
+In order to see the 'User Token' click the 'Access user token' button. This will prompt the 'Authenticate' dialog 
+where you are required to re-enter your credentials. After clicking 'Authenticate' in the completed dialog, 
+another dialog will appear with user token.
+
+Below the 'Access your token' section is another section that allows you to reset your token. Click the 'Reset 
+user token' button, which prompts an 'Authenticate' dialog. Enter your credentials panel to complete the user 
+token reset. Resetting the token will show a dialog with a success message and accessing the token will show the
```
bentmann: Do we really call those popups in the upper-right corner that automatically vanish after a few seconds dialogs? I would hope dialog is reserved for a window that I need to explicitly dismiss (via some OK/Cancel button). Maybe some non-updated wording from NX2?

bentmann: "accessing the token will show the dialog", what dialog and how is that related to the reset operation? I don't quite get what point this part of the sentence is trying to make.

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+In order to see the 'User Token' click the 'Access user token' button. This will prompt the 'Authenticate' dialog 
+where you are required to re-enter your credentials. After clicking 'Authenticate' in the completed dialog, 
+another dialog will appear with user token.
+
+Below the 'Access your token' section is another section that allows you to reset your token. Click the 'Reset 
+user token' button, which prompts an 'Authenticate' dialog. Enter your credentials panel to complete the user 
+token reset. Resetting the token will show a dialog with a success message and accessing the token will show the 
+dialog.
+
+The user code and pass code tokens in separate fields in the top level section as well as a server section ready
```
bentmann: Is that a proper sentence?

bentmann: Will we add a screenshot here as well? Otherwise I don't see how somebody is supposed to decipher those new terms "user code and pass code tokens" (the dialog has blurb to explain that those make up a user token).

dulanism: Yes, plan is to add screenshots

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+The user code and pass code tokens in separate fields in the top level section as well as a server section ready 
+to be used in a Maven `settings.xml` file. When using the server section you simply have to replace the 
+`${server}` placeholder with the repository id that references your repository manager you want to authenticate 
+against with the user token. The dialog will close automatically after one minute or simply click the 'Close' 
+button.
+
+The user code and pass code values can be used as replacements for username and password in the login dialog. It
+is also possible to use the original username and the pass code to log in to the user interface.
+
+Upon activation of user tokens for repository authentication you can access the repository manager with the user
```
bentmann: "you can", can expresses a possibility but as line 602 clarifies, it's a requirement to use the user token, so "you must" or "need to" seem more appropriate.

chapter-security.asciidoc
dulanism: Missing indicative plural verb - "are". Will fix

chapter-security.asciidoc
```diff
@@ -522,6 +523,90 @@ specific user account besides the global account used for the LDAP configuration
...
+curl -v --user admin:admin123 http://localhost:2468/repository/bower-all/
+----
+
+has to be replaced with the usage of user code and pass code separated by colon in the curl command line like this
+
+----
+curl -v --user N+ZBiTlF:76xSi+HAQvYHZH8kgyJldWD7aJnPgCrHG/Zu7mkpWmZZ http://localhost:2468/repository/bower-all/
+----
+
+User token values can be accessed as part of the Maven settings template feature automating updates as mentioned
```
bentmann: Does the "Maven settings template feature" already exist in NX3? I couldn't find it.

dulanism: Need to create an entry - based on https://books.sonatype.com/nexus-book/reference/settings-sect-install.html

chapter-security.asciidoc
```diff
@@ -522,6 +523,89 @@ specific user account besides the global account used for the LDAP configuration
...
+In order to see the 'User Token' click the 'Access user token' button. This will prompt the 'Authenticate' dialog 
+where you are required to re-enter your credentials. After clicking 'Authenticate' in the completed dialog, 
+another dialog will appear with the user token.
+
+Below the 'Access your token' section is another section that allows you to reset your token. Click the 'Reset 
+user token' button, which prompts an 'Authenticate' dialog. Enter your credentials panel to complete the user 
+token reset. Resetting the token will show a dialog with a success message, and accessing the token will show the 
+dialog.
+
+The 'User Token' dialog displays a user code and pass code tokens in separate fields. Below the token, is your
```
bentmann: "**a** user code and pass code token**s**", sounds like a mismatch of singular/plural to me

chapter-security.asciidoc
dulanism: @joedragons Made me aware of this. Adding a commented-out note.

chapter-security.asciidoc
```diff
@@ -432,8 +432,9 @@ specified a Search Base of +dc=sonatype,dc=com+, you use a value of +ou=users+.
 
 User subtree:: Check the box if 'True'. Uncheck if 'False'. Values are 'True' if there is a tree below the base 
 DN that can contain user entries and 'False' if all users are contain within the specified Base DN. For example,
```
joedragons: Possible later cleanup: Sentence says "base DN" and later "Base DN".  Seems inconsistent (or markup missing).

chapter-security.asciidoc
```diff
@@ -453,8 +454,8 @@ repository manager uses this attribute for the 'Email' attribute of the user. It
 of the user.
 
 Password attribute:: It can be used to configure the Object class, which supplies the password ("userPassword").
-If this field is blank the user will be authenticated against a bind with the LDAP server. The password attribute is
-optional. When not configured authentication will occur as a bind to the LDAP server.  Otherwise this is the
+If this field is blank the user will be authenticated against a bind with the LDAP server. The password attribute 
+is optional. When not configured authentication will occur as a bind to the LDAP server.  Otherwise this is the
 attribute of the Object class that supplies the password of the user. The repository manager uses this attribute
```
joedragons: Minor: While unrelated (I think this is just 114izing?) it seems to me we state the same thing here twice.  If this field is blank, bind is used...followed by attribute is optional (blank?), bind is used.  If there's a difference it's subtle and I'm missing it.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+	
+[[user-token]]
+=== Security Setup with User Tokens
+
+{inrmonly}
+
+[[user-token-intro]]
+==== Introduction
+
+When using Apache Maven with {pro} or {oss}, the user credentials for accessing the repository manager have to be
```
joedragons: Minor: I'm +0 saying {pro}/{oss} since this feature is just for PRO.  I think we can generalize here.  This is true as written so just minor comment.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+
+[[user-token-intro]]
+==== Introduction
+
+When using Apache Maven with {pro} or {oss}, the user credentials for accessing the repository manager have to be 
+stored in the user’s `settings.xml` file. Like a `pom.xml` your `settings.xml` is file that contains your user 
+preferences. The Maven framework has the ability to encrypt passwords within the +setting.xml+, but the need for 
+it to be reversible in order to be used limits its security.
+
+The +settings.xml+ lives in `$HOME/.m2`, your main Maven directory. This file contains listings for
```
joedragons: I'd be curious if people who know maven better than me think this is alright as written.
1) I am wondering if this is blanket true for all OS setups.
2) I am wondering if this is configurable.
3) Even with my individual setup, I am pretty sure that *A* settings.xml lives in .m2 but I think it can live elsewhere.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+
+This is especially useful for scenarios where single sign-on solutions like LDAP are used for authentication
+against the repository manager and other systems and the plain text username and password cannot be stored in the
+`settings.xml` following security policies. In this scenario the generated user tokens can be used instead.
+
+[[enable-reset-token]]
+==== Enabling and Resetting User Tokens
+
+User token-based authentication can be activated by an administrator or user with the +nx-usertoken-settings+ 
+privilege. If given that privilege click the 'User Token' menu item under 'Security' in the 'Administration'
```
joedragons: Minor: Suggest "Users with" rather than "If given".

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+list. To see the results, go to 'Realms' located under 'Security' in the 'Administration' menu. If desired, you 
+can re-order the security realms used, although the default settings with the 'User Token Realm' as the first 
+realm is probably the desired setup. This realm is not removed when the user tokens are disabled; however, 
+it will cleanly pass through to the next realm. The realm will remain in the active bin in your 'Realms' in case 
+the feature is reactivated at a later stage.
+
+[[access-token]]
+==== Accessing and Using Your User Token
+
+With enabled, you can access your individual token on your 'Account' panel. To access the panel, select the
```
joedragons: "When" (rather than "With")?  Or there's some text of what with is missing.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+against the repository manager and other systems and the plain text username and password cannot be stored in the
+`settings.xml` following security policies. In this scenario the generated user tokens can be used instead.
+
+[[enable-reset-token]]
+==== Enabling and Resetting User Tokens
+
+User token-based authentication can be activated by an administrator or user with the +nx-usertoken-settings+ 
+privilege. If given that privilege click the 'User Token' menu item under 'Security' in the 'Administration' 
+menu. Check the 'Enable user tokens' box, then press 'Save' to activate the feature. You can access the user 
+token from the 'Profile' mode toggle, explained in <<access-token>>.
```
joedragons: Minor: While true I find it awkward that this text sends someone off to the Profile to check something then continues to speak about configuration especially as the next sentence is started with "Additionally,".  Suggest moving this late or maybe to the end of this section or making this a NOTE:.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+could potentially break any command line builds using the tokens until this change is carried out. This also 
+applies to continuous integration servers using user tokens or any other automated build executions.
+
+[[user-token-realms]]
+==== Accessing User Tokens in Realms
+
+When you activate user tokens, the feature automatically adds the 'User Token Realm' to the 'Active Realms' 
+list. To see the results, go to 'Realms' located under 'Security' in the 'Administration' menu. If desired, you 
+can re-order the security realms used, although the default settings with the 'User Token Realm' as the first 
+realm is probably the desired setup. This realm is not removed when the user tokens are disabled; however,
```
joedragons: Minor: I feel like we've had support issues where we do not put the Nexus realms first but I cannot find any specifics.  Would be curious if @rseddon or @peterlynch have any thoughts here.  This seems safe, so marked minor, just a niggling doubt.

chapter-security.asciidoc
bentmann: The default location is as stated (cf. https://maven.apache.org/ref/3.3.9/maven-settings/settings.html) unless there are cases where `${user.home}` and `$HOME` resolve to different places.

The path to the settings file can be configured via the command line (cf. http://books.sonatype.com/mvnref-book/reference/running-sect-options.html#running-sect-custom-locations-option).

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+can re-order the security realms used, although the default settings with the 'User Token Realm' as the first 
+realm is probably the desired setup. This realm is not removed when the user tokens are disabled; however, 
+it will cleanly pass through to the next realm. The realm will remain in the active bin in your 'Realms' in case 
+the feature is reactivated at a later stage.
+
+[[access-token]]
+==== Accessing and Using Your User Token
+
+With enabled, you can access your individual token on your 'Account' panel. To access the panel, select the 
+'admin' mode toggle, on the top right-hand corner of the user interface.
```
joedragons: This is only 'admin' toggle, if they're 'admin'.  Probably should generalize here.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+
+In order to see the 'User Token' click the 'Access user token' button. This will prompt the 'Authenticate' dialog 
+where you are required to re-enter your credentials. After clicking 'Authenticate' in the completed dialog, 
+another dialog will appear with the user token.
+
+Below the 'Access your token' section is another section that allows you to reset your token. Click the 'Reset 
+user token' button, which prompts an 'Authenticate' dialog. Enter your credentials panel to complete the user
```
joedragons: "Enter your credentials panel" seems like at least two things condensed into one (entering credentials, panel appears).  Easiest suggestion would be remove "panel" but not sure all the intent here.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+'admin' mode toggle, on the top right-hand corner of the user interface.
+
+In order to see the 'User Token' click the 'Access user token' button. This will prompt the 'Authenticate' dialog 
+where you are required to re-enter your credentials. After clicking 'Authenticate' in the completed dialog, 
+another dialog will appear with the user token.
+
+Below the 'Access your token' section is another section that allows you to reset your token. Click the 'Reset 
+user token' button, which prompts an 'Authenticate' dialog. Enter your credentials panel to complete the user 
+token reset. Resetting the token will show a dialog with a success message, and accessing the token will show the 
+dialog.
```
joedragons: Multiple uses of dialog here and I think something might be lost.  Suggest "but you must access the user token again to see the new value." instead of " and accessing the token will show the dialog."

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+
+The 'User Token' dialog displays user code and pass code tokens in separate fields. Below the token, is your 
+Maven +settings.xml+ server file. When using the server section you simply have to replace the `${server}`
```
joedragons: Suggest "xml that can replace the server section in your Maven +settings.xml+" instead of "our +Maven +settings.xml+ server file".  Might be able to be improved, but the thought here is it's not the entire file, it's just a section of it.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+token reset. Resetting the token will show a dialog with a success message, and accessing the token will show the 
+dialog.
+
+The 'User Token' dialog displays user code and pass code tokens in separate fields. Below the token, is your 
+Maven +settings.xml+ server file. When using the server section you simply have to replace the `${server}` 
+placeholder with the repository id that references your repository manager you want to authenticate against with 
+the user token. The dialog will close automatically after one minute or simply click the 'Close' button.
+
+The user code and pass code values can be used as replacements for username and password in the login dialog. It
+is also possible to use the original username and the pass code to log in to the user interface.
```
joedragons: (Super)Minor: Suggest "still" instead of "also".  Idea is this was possible before so still actually speaks to the fact that if you have this setup, original stuff (still) works.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+
+Upon activation of user tokens for repository authentication you must access the repository manager with the user
```
joedragons: "must" should be "can" (or "can now") unless you have 'require user tokens' on.  I also originally read this as a setup instruction, I think changing the word also helps that.

chapter-security.asciidoc
```diff
@@ -522,6 +523,108 @@ specific user account besides the global account used for the LDAP configuration
...
+is also possible to use the original username and the pass code to log in to the user interface.
+
+Upon activation of user tokens for repository authentication you must access the repository manager with the user 
+token, from the command line. For example, your username-password credentials access with
+
+----
+curl -v --user admin:admin123 http://localhost:2468/repository/bower-all/
+----
+
+has to be replaced with the usage of user code and pass code separated by colon in the curl command line like this
```
joedragons: Again "has to" should be "can" or I'm missing the intent here.

joedragons:
+1 once non-minor edits are made.  Thanks!

chapter-security.asciidoc
dulanism: It was previously "can", then switched to "must". But your explanation makes sense. Reference: https://github.com/sonatype/nexus-book/pull/198#discussion_r72076789

joedragons: I see.  Well, you could rewrite and add "In order to utilize them, you must" then it's true=)

chapter-security.asciidoc
joedragons: I'd like to largely retract the above statement.  I just noticed that the User Token Realm is first automatically, which I was unaware of.  This largely mitigates my concern.  I am leaving the previous comment in case others are concerned.

chapter-security.asciidoc
dulanism: >When enabled, you can access your individual token on your 'Account' menu. To access the menu, select the username, on the top right-hand corner of the user interface.

dulanism: > When enabled, you can access your individual token on your 'Account' menu. To access the menu, select the username, on the top right-hand corner of the user interface. In the 'User' menu, to the left, you will see the 'User Token' icon appear. cc @joedragons

joedragons: My only hesitancy here is both with 'Account', there's nothing I see where it's labelled "Account" so I'm not sure if the markup is right.  If you tell me it is, I'll believe you=)  Also, this isn't a result of this text change, so probably I shouldn't even comment on it=)  +1

chapter-security.asciidoc
dulanism: Revised paragraph (with markup):

>The 'User Token' dialog displays user code and pass code tokens in separate fields. Below the token, is the server section of your +settings.xml+. When using the server section you can replace the `${server}` placeholder with the repository id that references your repository manager you want to authenticate against with the user token. The dialog will close automatically after one minute or simply click the 'Close' button.

cc: @joedragons

chapter-security.asciidoc
dulanism: Revised: 

> The user code and pass code values can be used as replacements for username and password in the login dialog. You can still to use the original username and the pass code to log in to the user interface.

cc @joedragons

chapter-security.asciidoc
dulanism: Rewrite:

> In order to utilize your user tokens for repository authentication you must access the repository manager with the user token, from the command line.

chapter-security.asciidoc
joedragons: +1

chapter-security.asciidoc
joedragons: +1

chapter-security.asciidoc
joedragons: I'm cool with this tho I'm curious what @bentmann thinks as he balked with "can" initially.

bentmann: require

I might have gotten confused by the wording here, mistaking it to refer to said option ("Protect Content" in NX2). With this option enabled, it the user token is "required". Otherwise, "can" is fine.

dulanism: Thanks @joedragons & @bentmann

chapter-security.asciidoc
joedragons: https://github.com/sonatype/nexus-book/commit/f86e919bbf241de061ef330c8223813762d3d7ba

PR #199, opened 2016-07-13
NEXUS-10375-update text to upgrade tool

dulanism:
CI: http://bamboo.s/browse/BOOK-NEXUS161


chapter-migration.asciidoc
```diff
@@ -1,52 +1,46 @@
-[[migration]]
-==  Migration
+[[upgrade]]
+==  Upgrade Tool
```
mosabua: I would NOT use Upgrade Tool ... if anything there are multiple Tools. But to keep it simple .. the chapter will probably contain 3.x to 3.y upgrade instructions later as well. So just call the chapter "Upgrade" or "Upgrading Nexus Repository Manager" .. or Upgrade Instructions .. but do not indicate a specific tool or so. imho anyway

chapter-migration.asciidoc
```diff
@@ -1,52 +1,46 @@
...
-////
-
-[[migration-introduction]]
+[[upgrade-introduction]]
 === Introduction
 
-The migration tool provides a built-in supported method for migrating components from {pro} 2.14 and higher to 
-{pro} {version} and higher. You will be able to initiate a migration into {pro} {version}, and later versions, 
+The upgrade tool provides a built-in supported method for migrating components from {pro} 2.14 and higher to 
+{pro} {version} and higher. You will be able to initiate an upgrade into {pro} {version}, and later versions,
```
mosabua: You can not use the version macro here... since later it would say e.g. 3.3 .. and that would be wrong.

mosabua:
Imho this chapter should be targeted at all upgrades 2.x to 3.x and 3.x to 3.y .. so it has to be a bit reworded and restructured to accommodate for that.

chapter-migration.asciidoc
dulanism: I hadn't started working on that yet but I was gonna ping you to follow up @mosabua. When I finish rewording, I'll get to that step. Thx in advance

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
+[[upgrading]]
+==  Upgrading Nexus Repository Manager
+{inall}
+
+[[upgrade-introduction]]
+=== Introduction
+
+The upgrade tool provides a built-in supported method for migrating components from {nxrm} 2.14 or later to 3.1
```
joedragons: Minor: I'm not sure how paranoid I am about any use of "migrating" now, but it's here=)  Alternative words: "copying", "moving", "transferring".

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+[[upgrade-introduction]]
+=== Introduction
+
+The upgrade tool provides a built-in supported method for migrating components from {nxrm} 2.14 or later to 3.1 
+or later. You will be able to initiate an upgrade into {nxrm} 3.1, and later versions, with a wizard available in 
+the application.
+
+This feature helps you build an upgrade plan which allows version 3 to talk to version 2 by way of a 
+capability that communicates with the upgrade tool. The plan you create with version 3 will provide a 
+wizard that guides you through an upgrade from the latest version of 2.x or higher to {nxrm} 3.1 or higher.
```
joedragons: Above you say 2.14 and higher, here you say 2.x.  Not sure if intentional or not but seems contrary?

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+During an upgrade, version 2 should remain functioning at high capacity during the majority of the upgrade 
+process. You or your organization should put strong emphasis on performance and stability when designing the 
+upgrade plan.
+
+NOTE: This chapter will cover a special feature used to upgrade from {nxrm} 2.14 to 3.1. However, the repository
```
joedragons: Minor: Suggest leaving out the point numbers and put "{nxrm} 2 to 3."  You've already said "or greater" a bunch of times above, hopefully they get it by now and it's tangential to this NOTE.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+capability that communicates with the upgrade tool. The plan you create with version 3 will provide a 
+wizard that guides you through an upgrade from the latest version of 2.x or higher to {nxrm} 3.1 or higher.
+
+During an upgrade, version 2 should remain functioning at high capacity during the majority of the upgrade 
+process. You or your organization should put strong emphasis on performance and stability when designing the 
+upgrade plan.
+
+NOTE: This chapter will cover a special feature used to upgrade from {nxrm} 2.14 to 3.1. However, the repository 
+manager provides alternative methods to upgrade if you are currently running older versions of 3.x; this is 
+limited to milestone 7. See our support knowledge base to walk you through the steps to upgrade from
```
joedragons: Minor: Suggest putting "back to milestone 7" rather than limited to.  It is obvious now, however, eventually when we have 3.1, 3.2, 3.3, 3.10 etc "limited to" will potentially be misleading.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+* where they want their components to be transferred to when the upgrade is complete
+* which of the two architectures is best for your upgrade plan
+* designating a maintenance window for version 2 so you can take it down for a period, and the upgrade can 
+proceed with interruption
+* how your system storage is structured and what ingestion methods are available to them
+
+The upgrade tool only allows one transfer at a time. After it starts you must wait for the upgrade 
+to complete if you decide to transfer more content. If you press 'Abort' during the 'Preparing' or 
+'Synchronizing' phases the contents will be deleted.
```
joedragons: Suggest saying something like "will not move" or "will not be transferred".  Be deleted is scary if you don't say from where.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+There are several ways to tune {nxrm} 2.14, or later, for a quicker upgrade. Review the current configuration in 
+your compatible version of 2.x to clean up content that may hinder performance.
+
+When considering upgrade time and speed, take into account all enabled scheduled tasks on your version 2 
+instance that you may not need. Depending on your the configuration of version 2 you could optimize the 
+performance of the application by:
+
+* disabling system feeds
+* disabling repair tasks
+* reviewing Custom Metadata
```
joedragons: Minor: I think I said this on earlier "update" review and it's still here but I don't think this bullet is helpful unless we tell them what to review for.  All the other ones are straightforward tasks (disabling something, removing something or increasing a size).  This is akin to saying "Review update on the internet".  Maybe it'll be clear what you are "reviewing" for when you do it...  Marked minor since this passed previous reviews.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+As mentioned in <<admin-repository-blobstores,Chapter 4>> you can create new ones as well as determine the amount 
+of space for each, if more than one exists. In version 3, consider how you want to organize this storage 
+mechanism. Components in {nxrm} 2.x are stored as individual files on disk. {nxrm} version 3 stores components as 
+blobs. The conversion process requires the repository manager to iterate every component in your storage. This 
+process will take the bulk of the time required for upgrading.
+
+When upgrading, make sure you have enough storage capacity in the destined file system(s). If you are using 
+the hard linking ingestion method, the bytes themselves are not duplicated (saving space), but you must ensure 
+there are enough free inodes for the content being upgraded.
```
joedragons: Minor: "content being upgraded" sounds weird to me since the component versioning is all the same (really it's the server not the content being upgraded).  This may be intentional to avoid switching words all the time.  Would suggest "moved" or "transferred" like I suggested above however for this case.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+has to serve the bytes. Copying and hard linking only have to be told where the file is, and they obtain it from 
+the file system, directly.
+
+[[upgrade-content-prep]]
+=== Preparing Content for Upgrade
+
+*Repository IDs*
+
+Before migrating repositories to version 3, find the repository IDs in version 2 and plan how you will 
+rename any conflicting repositories in version . Contact our https://support.sonatype.com/hc/en-us[support
```
joedragons: I think a 3 is missing here after "in version" and before ".".  Otherwise, I think "in version " can be struck.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+*Repository IDs*
+
+Before migrating repositories to version 3, find the repository IDs in version 2 and plan how you will 
+rename any conflicting repositories in version . Contact our https://support.sonatype.com/hc/en-us[support 
+team] to learn on how to do this safely.
+
+*User Token*
+
+The upgrade tool can only upgrade user tokens from version 2 if the 'Enabled' box is checked. From the
```
joedragons: Minor: Suggesting putting which version it needs to be checked in.  I believe it's 3 based around 2 sentences down but I'm not 100% sure.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+link to user token chapter
+////
+
+*NuGet API Key*
+
+Though your account has only one associated NuGet API key, they are visible per repository.
+
+The upgrade tool will add all keys to {nxrm} 3, that are present in {nxrm} 2.14 when asked even if the NuGet API 
+Key Realm is not active in version 2. This is because there is no explicit on or off setting for NuGet keys. You 
+can locate 'Selected Realms' under 'Administration' -> 'Server' -> 'Security Settings'.
```
joedragons: Minor: I am not sure the point to the last sentence.  While true, it seems like tangential or an unfinished thought.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+////
+
+[[upgrade-methods]]
+=== Upgrade Methods
+
+In order to start an upgrade, version 2 has to listen for version 3 to initiate communication. This is the core 
+property of the 'Migration: Agent'. When the upgrade wizard in version 3 helps you build an upgrade plan, it 
+talks to your version 2 instance thru the migration-agent to find out what repositories exist (e.g. to list them 
+in the wizard user interface). When you pick them, version 3 tells version 2 to start compiling a full list of 
+what it has that needs migrating in each of those repositories. Ultimately, this creates a scheduled task that
```
joedragons: "migrating" used here (unless intentional or you assume you'll be back over this section when we rename the names in the app)

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+NOTE: This chapter will cover a special feature used to upgrade from {nxrm} 2.14 to 3.1. However, the repository 
+manager provides alternative methods to upgrade if you are currently running older versions of 3.x; this is 
+limited to milestone 7. See our support knowledge base to walk you through the steps to upgrade from 
+https://support.sonatype.com/hc/en-us/articles/222159808[milestone 7 of version 3 to 3.0.0 final] or 
+https://support.sonatype.com/hc/en-us/articles/217967608[version 3.0.0 to a later version].
+
+[[upgrade-support]]
+=== Supported Architecture for Upgrading
+
+There are two available architectures that support upgrading. You can either upgrade content from version 2.14
```
bentmann: "from **a** version 2.14 server"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+'Synchronizing' phases the contents will be deleted.
+
+[[upgrade-performance]]
+==== Performance and Tuning for Upgrading
+
+There are several ways to tune {nxrm} 2.14, or later, for a quicker upgrade. Review the current configuration in 
+your compatible version of 2.x to clean up content that may hinder performance.
+
+When considering upgrade time and speed, take into account all enabled scheduled tasks on your version 2 
+instance that you may not need. Depending on your the configuration of version 2 you could optimize the
```
bentmann: "your ~~the~~ configuration"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+* disabling system feeds
+* disabling repair tasks
+* reviewing Custom Metadata
+* removing unused snapshots
+* removing old scheduled tasks not in use
+* removing repositories no longer supported by the repository manager (e.g. 
+https://support.sonatype.com/hc/en-us/articles/217611787-codehaus-org-Repositories-Should-Be-Removed-From-Your-Nexus-Instance[Codehaus repositories])
+* disabling 'Rebuild Maven Metadata Files' scheduled tasks
+* increasing your file size limit (e.g. using 'https://support.sonatype.com/hc/en-us/articles/213465218-The-nexus-log-file-is-full-of-too-many-open-files-exceptions-how-can-I-fix-this-[+nexus.log+]')
```
bentmann: "file size limit"? The linked KB isn't about file size, it's about open file handles.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+Another factor, the quality of your network, will arbitrarily impact the speed of ingestion methods 
+(e.g HTTP download).
+
+[[upgrade-file-systems]]
+==== File System Considerations
+
+As mentioned in <<admin-repository-blobstores,Chapter 4>> you can create new ones as well as determine the amount
```
bentmann: "new ones", from where does the reader get the context to resolve the meaning of "one"?

chapter-upgrading.asciidoc
bentmann: What's a "conflicting" repository? Have we detailed when a conflict in NX 3 would occur or have we just signed up our support team to clarify this for each and every customer?

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+[[upgrade-content-prep]]
+=== Preparing Content for Upgrade
+
+*Repository IDs*
+
+Before migrating repositories to version 3, find the repository IDs in version 2 and plan how you will 
+rename any conflicting repositories in version . Contact our https://support.sonatype.com/hc/en-us[support 
+team] to learn on how to do this safely.
+
+*User Token*
```
bentmann: The previous paragraph has a plural header, shouldn't this one as well, i.e. "User TokenS"?

chapter-upgrading.asciidoc
bentmann: >  I believe it's 3 

It needs to be enabled in NX 2 (as otherwise the tokens cannot be read out of their database)

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+The upgrade tool can only upgrade user tokens from version 2 if the 'Enabled' box is checked. From the 
+'Administration' panel click the 'User Token' tab and enable the setting. This automatically activates 'Realms' 
+in the version 3 panel, even if it is inactive.
```
bentmann: I assume others understand what this sentence is supposed to convey, I don't get it.

chapter-upgrading.asciidoc
bentmann: Also doesn't clarify which NX version is being talked about.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+TBD
+
+potential new section
+////
+
+[[upgrade-methods]]
+=== Upgrade Methods
+
+In order to start an upgrade, version 2 has to listen for version 3 to initiate communication. This is the core 
+property of the 'Migration: Agent'. When the upgrade wizard in version 3 helps you build an upgrade plan, it
```
bentmann: "Migration: Agent", I assume this is going to be changed when the UI will be reworded?

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+[[upgrade-methods]]
+=== Upgrade Methods
+
+In order to start an upgrade, version 2 has to listen for version 3 to initiate communication. This is the core 
+property of the 'Migration: Agent'. When the upgrade wizard in version 3 helps you build an upgrade plan, it 
+talks to your version 2 instance thru the migration-agent to find out what repositories exist (e.g. to list them 
+in the wizard user interface). When you pick them, version 3 tells version 2 to start compiling a full list of 
+what it has that needs migrating in each of those repositories. Ultimately, this creates a scheduled task that 
+walks though the repository storage folder, lists the files, then synchronizes them, storing the metadata in the
```
bentmann: Typo, "though" -> "through"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+If versions 2 and 3 are on the same machine (or share access to the same file systems) then version 3 will 
+copy the files from version 2. Version 2 will tell version 3 the location of the file and where to retrieve the 
+content. 
+
+Assuming versions 2 and 3 are on the same machine, configured in a way that the mounts are accessible by the same 
+path (from one machine to the other) this option will work. It is a slightly faster process than the download 
+method and has less impact on the performance of version 2.
+
+*Hard linking*
```
bentmann: Title-case, "Hard Linking"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+This method only works on the same file system. If you want to hard link, configure your instance of version 3 in 
+such a way that you have a blob store defined in an appropriate location where hard linking is possible.
+
+Hard linking repository metadata creates another reference to the same set of bytes somewhere else. If the 
+content set for upgrade in version 2 is too large for version 3, part of the data will be distributed to 
+cleared data blocks, or elsewhere.
```
bentmann: No clue what that means, I assume somebody with Unix background does.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+|Maven2   |yes        | 3.0*, 3.1 and greater
+|RubyGems |yes        | 3.1 and greater
+|Bower    |no         | 3.1 and greater
+|PyPI     |no         | 3.1 and greater
+|P2       |yes        | Not supported
+|OBR      |yes        | Not supported
+|Yum      |yes        | Not supported
+|============================================
+
+NOTE: The '*' in the list indicates missing advanced features.
```
bentmann: Upgrade is generally only available in 3.1+, as such I'm not sure there is much point in the added complexity/confusion around what "advanced features" 3.0 may or may not support.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+[[upgrade-start]]
+=== Starting the Upgrade
+
+After considerations around system performance and storage are taken into account, there are a few basic steps 
+to start the upgrade:
+
+* Upgrade the repository manager to the latest version of 2.x and configure the upgrade capability that allows 
+you to synchronize that instance with {nxrm} 3.1 or later.
+* Install 2.14 or greater on one server, and version 3.1 or later on another non-conflicting server.
```
bentmann: In step one I upgraded my 2.x instance, now I should "install" another 2.14 instance? That little word "install" seems to give rise to potential confusion, "run" might be an alternative.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+potential new section
+////
+
+[[upgrade-methods]]
+=== Upgrade Methods
+
+In order to start an upgrade, version 2 has to listen for version 3 to initiate communication. This is the core 
+property of the 'Migration: Agent'. When the upgrade wizard in version 3 helps you build an upgrade plan, it 
+talks to your version 2 instance thru the migration-agent to find out what repositories exist (e.g. to list them
```
bentmann: "thru", is this appropriate for the docs? Except here, you seem to use "through"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+* Upgrade the repository manager to the latest version of 2.x and configure the upgrade capability that allows 
+you to synchronize that instance with {nxrm} 3.1 or later.
+* Install 2.14 or greater on one server, and version 3.1 or later on another non-conflicting server.
+* Configure the migration-agent in version 2 and start the upgrade wizard in version 3.
+
+{nxrm} 3.1 provides a wizard to instruct you through upgrading in three phases:
+
+ * 'Preparing', the phase that prepares the transfer and creation of all components.
+ * 'Synchronizing', the phase that counts and processes all components set the upgrade.
```
bentmann: Did you mean "set **to** upgrade" by chance?

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+ * 'Finishing', the phase that performs final clean up, then closes the process.
+
+To execute the upgrade plan you must open the connection between version 2 and 3. The connection finds what 
+repositories exist and lists them in the upgrade wizard. It enables the port you configured to run version 2 
+remotely in order to communicate with version 3. The upgrade plan, as a whole, is two-part process where version
```
bentmann: "is **a** two-part process"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+In version 2, activate the 'Migration: Agent' capability to open the connection for the migration-agent. Follow 
+these steps:
+
+* Click 'Administration' in the left-hand panel
+* Open the 'Capabilities' screen
+* Select 'New' to prompt the 'Create new capability' modal
+* Select 'Migration: Agent' as your capability 'Type'
+
+In the lower section of 'Capabilities' interface, the repository manager acknowledges the migration-agent as
```
bentmann: "**the** Capabilities interface"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+[[upgrade-content]]
+==== Upgrading Content
+
+After you establish upgrade capabilities for versions 2 and 3, you will activate the wizard to start your 
+upgrade. In version 3 go to the 'Administration' menu and select 'Migration', located under 'System', to 
+open the wizard.
+
+*Wizard*
+
+Overview:: The wizard will provide and overview of what is allowed for an automatic upgrade as well as
```
bentmann: "**an** overview"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+location from version 2 and paste it to the 'URL' field so upgrading repositories will persist. An example input 
+would be: +http://localhost:8081/nexus/service/siesta/migrationagent+. The 'Access Token' will display the 
+security key from your version 2 'Migration: Agent' capability details.
+
+Content:: This screen presents checkboxes for security features ('Security'), server configuration ('System'), 
+and user-managed repositories ('Repositories') that can be upgraded. For 'Security' you have the option to choose 
+among 'Anonymous', 'LDAP Configuration', 'NuGet API-Key', 'Realms', 'Roles', 'SSL Certificates' 'Users', and 
+'User Tokens'. For 'System' you can upgrade 'Email' and 'HTTP Configuration' contents.
+
+NOTE: Some 'Content' items can only be accessed and upgraded by {nxrm} users.
```
bentmann: Did you mean `{pro} users`?

The note would probably become superfluous if the previous paragraph didn't list the specific options, end users can probably read the UI for themselves.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+'User Tokens'. For 'System' you can upgrade 'Email' and 'HTTP Configuration' contents.
+
+NOTE: Some 'Content' items can only be accessed and upgraded by {nxrm} users. 
+
+Repository Defaults:: If 'User-related repositories' is one of your selections from the 'Content' screen, the 
+'Repository Defaults' screen allows you to select directory destination and upgrade method. The first dropdown 
+menu, 'Destination' gives your option to pick a blob store name different than the default. The second dropdown
+menu, 'Method', allows you to choose among hard linking, copying local files or downloading.
+
+Repositories:: If 'User-related repositories' is one of your selections from the 'Content' screen, the
```
bentmann: "User-managed repositories"

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+Repository Defaults:: If 'User-related repositories' is one of your selections from the 'Content' screen, the 
+'Repository Defaults' screen allows you to select directory destination and upgrade method. The first dropdown 
+menu, 'Destination' gives your option to pick a blob store name different than the default. The second dropdown
+menu, 'Method', allows you to choose among hard linking, copying local files or downloading.
+
+Repositories:: If 'User-related repositories' is one of your selections from the 'Content' screen, the 
+'Repositories' screen allows you to select which repositories you want to upgrade. You can either select all 
+repositories with one click, at the top of the table. Alternatively, you can click each individual repository. In 
+addition to 'Repository', the table displays information around the status of the repository. The table includes 
+'Type', 'Format', 'Supported', 'Status', 'Destination', and 'Method'.
```
bentmann: Another case of the doc just stating what everybody can read themselves in the UI.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+Preview:: This table displays a preview of the content set for upgrade, selected in the previous screen. 
+Click 'Begin', then confirm from the modal, that you want to start the upgrade. After the preview 'Preparing', 
+'Synchronizing', and 'Finishing' will persist.
+
+When the final content upgrade ends, shut down version 2 and reboot to see your content, now replicated in
```
bentmann: "reboot" which version?

bentmann:
Did either @peterlynch or @rseddon had a chance to read the upgrade doc to see whether they have more feedback regarding the completeness or clarity of the process to avoid support interaction where feasible?

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+While version 2 is active repository content will be added, updated, and deleted as the upgrade continues 
+within version 3. Version 3 will pick up these changes. However, configuration changes such as new 
+repositories, realm settings, permissions, roles and role assignments, HTTP configuration, and SSL certificates 
+should not be changed. After the upgrade starts such changes will not be picked up by version 3 and may result 
+in failure.
+
+If you are an administrator who manages the upgrade plan consider updating your version 2 instance to read-only 
+mode so it is inaccessible to users without administrative privileges. This shuts off the flow of new 
+changes, and allows version 3 to catch up with any outstanding content changes made on version 2.
```
joedragons: This seems a good place for a link on how to do this if one such exists.  I am not sure how to do this so it makes sense to me.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+[[upgrade-configuration]]
+==== Enabling Upgrade in Version 2
+
+In version 2, activate the 'Migration: Agent' capability to open the connection for the migration-agent. Follow 
+these steps:
+
+* Click 'Administration' in the left-hand panel
+* Open the 'Capabilities' screen
+* Select 'New' to prompt the 'Create new capability' modal
+* Select 'Migration: Agent' as your capability 'Type'
```
joedragons: Minor: I find it slightly weird you're so specific here but yet you don't say click "Add".  Marked minor because I guess I didn't find this weird last time I reviewed=)

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+* Select 'Migration', then click 'Create capability' to enable the upgrade
+
+[[upgrade-content]]
+==== Upgrading Content
+
+After you establish upgrade capabilities for versions 2 and 3, you will activate the wizard to start your 
+upgrade. In version 3 go to the 'Administration' menu and select 'Migration', located under 'System', to 
+open the wizard.
+
+*Wizard*
```
joedragons: Minor: At least in the PDF, this looks weird to me because it's the same font as the bullets below.  This might just be a PDF thing (or an asciidoc thing); I didn't check the HTML.
screen shot 2016-07-19 at 9 46 11 am

chapter-upgrading.asciidoc
joedragons: Minor: I'm +0 on "an automatic".  I think "the" upgrade is just as accurate, less verbose and less prone to interpretation (such as, is there a non-automatic upgrade?).

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+Overview:: The wizard will provide and overview of what is allowed for an automatic upgrade as well as 
+warnings on what cannot be upgraded. Click the 'Next' buttons to proceed.
+
+Agent Connection:: This screen presents two fields, 'URL' and 'Access Token'. Copy over the server's service
```
joedragons: Unless something has changed, the service location is not available to copy in NX2 and must be extrapolated.  So I think the word "over" (maybe even "Copy over") is misleading.  Of more value here, I think would be how to craft it.

chapter-upgrading.asciidoc
joedragons: Honestly not sure what this note means.  It seems @bentmann understands, so I'll leave it at that=)

chapter-upgrading.asciidoc
joedragons: I believe you can also click and change each repository's individual method.  Unless we're intentionally not relaying that, it seems that would be useful to add.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+menu, 'Destination' gives your option to pick a blob store name different than the default. The second dropdown
+menu, 'Method', allows you to choose among hard linking, copying local files or downloading.
+
+Repositories:: If 'User-related repositories' is one of your selections from the 'Content' screen, the 
+'Repositories' screen allows you to select which repositories you want to upgrade. You can either select all 
+repositories with one click, at the top of the table. Alternatively, you can click each individual repository. In 
+addition to 'Repository', the table displays information around the status of the repository. The table includes 
+'Type', 'Format', 'Supported', 'Status', 'Destination', and 'Method'.
+
+Preview:: This table displays a preview of the content set for upgrade, selected in the previous screen.
```
joedragons: Minor: Should screen be "screens" in the likely case they have picked repository and non-repository items to upgrade?

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+Repositories:: If 'User-related repositories' is one of your selections from the 'Content' screen, the 
+'Repositories' screen allows you to select which repositories you want to upgrade. You can either select all 
+repositories with one click, at the top of the table. Alternatively, you can click each individual repository. In 
+addition to 'Repository', the table displays information around the status of the repository. The table includes 
+'Type', 'Format', 'Supported', 'Status', 'Destination', and 'Method'.
+
+Preview:: This table displays a preview of the content set for upgrade, selected in the previous screen. 
+Click 'Begin', then confirm from the modal, that you want to start the upgrade. After the preview 'Preparing', 
+'Synchronizing', and 'Finishing' will persist.
```
joedragons: Minor: I'm +0 on the word persist based around [websters simple definition](http://www.merriam-webster.com/dictionary/persist) which matches my understanding of the word.  Alternative word suggestion: "follow".

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+Agent Connection:: This screen presents two fields, 'URL' and 'Access Token'. Copy over the server's service
+location from version 2 and paste it to the 'URL' field so upgrading repositories will persist. An example input 
+would be: +http://localhost:8081/nexus/service/siesta/migrationagent+. The 'Access Token' will display the 
+security key from your version 2 'Migration: Agent' capability details.
+
+Content:: This screen presents checkboxes for security features ('Security'), server configuration ('System'), 
+and user-managed repositories ('Repositories') that can be upgraded. For 'Security' you have the option to choose 
+among 'Anonymous', 'LDAP Configuration', 'NuGet API-Key', 'Realms', 'Roles', 'SSL Certificates' 'Users', and 
+'User Tokens'. For 'System' you can upgrade 'Email' and 'HTTP Configuration' contents.
```
joedragons: 'Repository Targets' (content selectors) can not be migrated as of this sprint as well, under the 'Repositories' section.  Ref: https://issues.sonatype.org/browse/NEXUS-10379

chapter-upgrading.asciidoc
joedragons: As far as I know (just tested) you shouldn't need to reboot either version, the updated stuff just exists now on NX3.  I think this paragraph can be struck (or rewritten with intent) unless someone knows something I don't.

mrprescott: Yeah, no need for rebooting.

joedragons:
+1
Despite having so many comments, I am giving +1 because I believe very few of my comments had to do with the actual work done, just a happenstantial (re)review of the entire section because of the way it was adjusted.  Also most of them were minor.
Thanks for the continuted work.

chapter-upgrading.asciidoc
dulanism: Typo. Will fix.

chapter-upgrading.asciidoc
dulanism: "end users can probably read the UI for themselves", that being said the UI speaks for itself. Will remove.

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+The upgrade tool provides a built-in supported method for migrating components from {nxrm} 2.14 or later to 3.1 
+or later. You will be able to initiate an upgrade into {nxrm} 3.1, and later versions, with a wizard available in 
+the application.
+
+This feature helps you build an upgrade plan which allows version 3 to talk to version 2 by way of a 
+capability that communicates with the upgrade tool. The plan you create with version 3 will provide a 
+wizard that guides you through an upgrade from the latest version of 2.x or higher to {nxrm} 3.1 or higher.
+
+During an upgrade, version 2 should remain functioning at high capacity during the majority of the upgrade
```
dbradicich: the phrase 'during an upgrade' is effectively repeated here, maybe something like

```During the majority of the upgrade process, version 2 should remain functioning at high capacity```

File: chapter-upgrading.asciidoc (hunk 14 above)
Comment (dbradicich) [inline chapter-upgrading.asciidoc]: ```There are two architectures that can be used for upgrading.  You can have separate nexus 2 and nexus 3 servers, or you can have both nexus 2 and 3 running on the same server in different service locations```

That sounds a bit more clear.  The way 'either' was used and separating the options into different sentences just didn't sit well with me

File: chapter-upgrading.asciidoc [hunk 43]
```diff
@@ -0,0 +1,303 @@
...
+server to a version 3.1 server. The other option is you set both instances to run the upgrade on the same 
+server; this requires that you create two separate service locations for your server.
+ 
+[[upgrade-prep]]
+=== Preparing for an Upgrade
+
+If you decide to update versions 2.x and 3.x to the latest instances that allow you to upgrade, the most critical 
+decisions you or your organization make around the upgrade plan should include: 
+
+* where they want their components to be transferred to when the upgrade is complete
```
dbradicich: 'you' instead of 'they' and 'your' instead of 'their' ?

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+* which of the two architectures is best for your upgrade plan
+* designating a maintenance window for version 2 so you can take it down for a period, and the upgrade can 
+proceed with interruption
+* how your system storage is structured and what ingestion methods are available to them
```
dbradicich: 'you' instead of 'them' ?

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+
+When upgrading, make sure you have enough storage capacity in the destined file system(s). If you are using 
+the hard linking ingestion method, the bytes themselves are not duplicated (saving space), but you must ensure 
+there are enough free inodes for the content being upgraded.
+
+In the instance you have configured your repository manager HTTP context and path as a reverse proxy server
+the traffic coming from version 2 could be busy, as you start the upgrade steps. Version 3 starts pulling 
+over the content with the upgrade tool as content is added to version 2.
+
+NOTE: HTTP download puts more load on the version 2 instance than other ingestion methods, because this version
```
dbradicich: 'HTTP download ingestion puts more load on the 2.x instance that other ingestion methods, because this method requires 2.x to serve all of the content.  Copying and hard linking only have to be.....'

chapter-upgrading.asciidoc
dbradicich: what he said ^

chapter-upgrading.asciidoc
```diff
@@ -0,0 +1,303 @@
...
+walks though the repository storage folder, lists the files, then synchronizes them, storing the metadata in the 
+changelog database.
+
+On the version 3 side, components are stored as blobs. The conversion process requires version 3 to iterate over 
+every component in your storage.
+
+Once the 'Migration: Agent' capability, mentioned in <<upgrade-start>>, is enabled and both repository 
+manager instances are stood up and communicating, you can consider one of three ingestion methods: HTTP download, 
+file system copying, or hard linking. The choice of ingestion method can be made on a repository-by-repository 
+basis, but there is a upgrade-wide default.
```
dbradicich: a -> an

chapter-upgrading.asciidoc
dulanism: Revising intro to: 

> The upgrade tool provides a built-in supported method for transferring components from {nxrm} versions 2.14 to 3.1. Future iterations of this feature will be available beyond both 2.14 and 3.1, alike.

This steps up the narrative for me to just write "version 2" and "version 3" throughout the rest of the doc.

chapter-upgrading.asciidoc
dulanism: See https://github.com/sonatype/nexus-book/pull/199#discussion_r71361797

chapter-upgrading.asciidoc
dulanism: Though i don't like 'upgrade tool' out the gate; will revise.

chapter-upgrading.asciidoc
dulanism: Revised bullet: 
* Review Custom Metadata Capability Enabled

chapter-upgrading.asciidoc
dulanism: @bentmann short answer: yes. There's a document from written by Peter that anticipates this. Will contact him for more feedback.

chapter-upgrading.asciidoc
dulanism: If I understand your suggestion, does this read better?

> The upgrade tool can only upgrade user tokens from version 2 if the 'Enabled' box is checked. In version 3, click the 'User Token' tab, 'Administration' panel, and enable the setting. This automatically activates 'Realms' in the version 3 panel, even if it is inactive.

chapter-upgrading.asciidoc
dulanism: and the new word shall be....

> replicating

chapter-upgrading.asciidoc
bentmann: >  In version 3, click the 'User Token' tab

Is that version 3 a typo? As I tried to explain, the feature needs to be enabled in NX**2** if one expects any pre-existing user tokens from NX2 to make it over to NX3. The user most definitively does not need to enable the feature in NX3 prior to upgrade, it's the job of the upgrade tool to make NX3 match the configuration of NX2.

chapter-upgrading.asciidoc
dulanism: @dbradicich I like that.

chapter-upgrading.asciidoc
joedragons: @dulanism It does IMO.  I've been assuming that by articulating the second and third sentences that even if enabled in NX2 the capability is not enabled in NX3 thus you need to explain why.  Which actually is weird if they 'upgrade' capabilities as well, but maybe they didn't and better to explain it.
I also didn't test this right now, so my suppositions could be false.

chapter-upgrading.asciidoc
dulanism: Will remove.

chapter-upgrading.asciidoc
dulanism: Yes

chapter-upgrading.asciidoc
dulanism: Adapted from original notes taken in March/April. Probably speculative at the time. Removing to avoid confusion.

chapter-upgrading.asciidoc
dulanism: I'll rethink this section if needed. In the meantime, I'll remove it.

chapter-upgrading.asciidoc
dulanism: This alludes to a use case discussed when putting together the outline. I'll remove it, talk to Peter about it. Might be better for a KB article, blog.

chapter-upgrading.asciidoc
dulanism: I just removed *Wizard*. The last sentence about and the first line in *Overview* are adequate paragraph hooks.

chapter-upgrading.asciidoc
dulanism: Saving this for another ticket.

chapter-upgrading.asciidoc
dulanism: Second person, you and your.

chapter-upgrading.asciidoc
dulanism: Second person

PR #201, opened 2016-07-25
Adding tip to use SSL w/ pip

joedragons:
Based around recent experiences personal and with beta tester.

### DIFF
--- a/chapter-pypi.asciidoc
+++ b/chapter-pypi.asciidoc
@@ -18,8 +18,8 @@ Also, you can publish your own packages to a private index as a hosted repositor
 expose the remote and private repositories as a repository group, which is a repository that merges and exposes 
 the contents of multiple repositories in one convenient URL.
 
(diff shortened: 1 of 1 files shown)


chapter-pypi.asciidoc
```diff
@@ -18,6 +18,10 @@ Also, you can publish your own packages to a private index as a hosted repositor
 expose the remote and private repositories as a repository group, which is a repository that merges and exposes 
 the contents of multiple repositories in one convenient URL.
 
+TIP: If using pip with the repository manager, strongly consider setting up your repository manager to use SSL as
```
dulanism: The main clause is missing for this sentence. The conditional might sound better as "If using pip with the repository manager, we strongly consider..."  or "If using pip with the repository manager, you should consider..."

dulanism:
Otherwise +1

PR #202, opened 2016-07-26
Touch ups to repository config section

joedragons:
* User noticed a screen was out of date (incorrect tab).
* Subsequently I noticed two figures were named the same, so renamed
them.
* This effort drew me into a small sentence rewrite after trying to
figure out what exactly we were saying the two figures were referencing.

I made a PR specifically to assure the screens are good in various
formats before merging; will set to review after I check.


mrprescott:
+1

dulanism:
+1

mosabua:
@joedragons I assume you tested the scale value of 100 .. it affects the PDF only so you need to test it there. In the HTML only the original image size is used as is. fyi @dulanism @annrollo

joedragons:
@mosabua I did, thus so many commits=)  Thanks for the info; it will be invaluable (next time=D).

PR #205, opened 2016-08-01
NEXUS-10422-content-selector-chapter

dulanism:
JIRA https://issues.sonatype.org/browse/NEXUS-10422
CI: http://bamboo.s/browse/BOOK-NEXUS167


dulanism:
Moving this over for preliminary review, provided new changes to the UI in the works

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
 across multiple repositories will be handled by the repository manager and therefore be much faster than client
 side resolution done by Maven each time.
 
+[[content-selectors]]
+==== Content Selectors
+{inrmonly}
```
joedragons: At least generally, available in OSS as well.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+
+Content selectors is a set of regular expressions used to match characters on the path of components in a
```
joedragons: Minor: Might be some grammar rule/bad english on my part but "selectors is a set" sounds weird to me.  "selectors are a set" sounds better.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
 across multiple repositories will be handled by the repository manager and therefore be much faster than client
 side resolution done by Maven each time.
 
+[[content-selectors]]
+==== Content Selectors
+{inrmonly}
+
+Content selectors is a set of regular expressions used to match characters on the path of components in a 
+repository. Content selectors allow you to select specific content, with regular expressions that are defined by 
+you from all our your content. The regular expressions are implemented with
```
joedragons: * Typo "our" (I believe should be "of"
* Minor (comma drama): The first comma (after content) surprised me, but if you cull the rest following, it makes sense.  However, also suggestion an additional one after "by you" (ie: content, with regular expressions that are defined by you, from...)

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+{inrmonly}
+
+Content selectors is a set of regular expressions used to match characters on the path of components in a 
+repository. Content selectors allow you to select specific content, with regular expressions that are defined by 
+you from all our your content. The regular expressions are implemented with 
+https://commons.apache.org/proper/commons-jexl/[JEXL], a scripting language, that allows filtering and querying 
+on component patterns.
+
+Content selectors allow you to define, for example, a selector called Apache Maven with a pattern of 
+`^/org/apache/maven/.*`. This would match all components with the groupId `org.apache.maven` and any components
```
joedragons: Minor: Suggest matching terms with UI field names =
* "called" becoming "named" (field is "Name:")
* "pattern" becoming "search expression" (field is "Search expression:")

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+on component patterns.
+
+Content selectors allow you to define, for example, a selector called Apache Maven with a pattern of 
+`^/org/apache/maven/.*`. This would match all components with the groupId `org.apache.maven` and any components 
+within nested within the 'Group' such as `org.apache.maven.plugins`.
+
+[[selector-create]]
+===== Creating a New Content Selector
+
+To create a new selector query, click 'Content Selectors' located in 'Administration' menu to get to the form. In
```
joedragons: This intro sentence seems super simplified.  It's missing the typical "in the Repository section" as well as the fact you have to click the "Create selector" button.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+
+Content selectors allow you to define, for example, a selector called Apache Maven with a pattern of 
+`^/org/apache/maven/.*`. This would match all components with the groupId `org.apache.maven` and any components 
+within nested within the 'Group' such as `org.apache.maven.plugins`.
+
+[[selector-create]]
+===== Creating a New Content Selector
+
+To create a new selector query, click 'Content Selectors' located in 'Administration' menu to get to the form. In 
+the 'Selector ID' section enter a 'Name' and 'Description' (optional) of your selector in the corresponding
```
joedragons: Super Minor: I prefer "(optional) 'Description'" because it (verbally) reads better and also there's less potential confusion both fields may be optional.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+within nested within the 'Group' such as `org.apache.maven.plugins`.
+
+[[selector-create]]
+===== Creating a New Content Selector
+
+To create a new selector query, click 'Content Selectors' located in 'Administration' menu to get to the form. In 
+the 'Selector ID' section enter a 'Name' and 'Description' (optional) of your selector in the corresponding 
+fields. In the 'Specification' section use the 'Search expression' field to enter your JEXL query.
+
+Creating a new content selector allows you to identify what components or assets you want your user to have
```
joedragons: Until preview is done (this sprint?) I think "identify what" is not true.  Nothing about content selectors helps you figure out (identify) what you're selecting:P  It's mostly so you can "assign privilege to" (or "limit what").
We make this easier with preview but I don't think that's the point.  Let me know if not clear.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+Other JEXL examples may include:
+
+*Selecting all 'raw' format content*
+format == 'raw'
+
+*Selecting all 'maven2' content with a group that starts with 'org.sonatype.nexus'*
+format == 'maven' and coodinate.group =^'org.sonatype.nexus'
+
+*Selecting all 'nuget' content in the 'mycompany' namespace*
+format == 'nuget' and path =^'/MyCompany'.
```
joedragons: General notes: 
At least in the pdf, none of the single quotes show up.  Instead they are italic.

screen shot 2016-08-16 at 3 39 12 pm

I am also not sure if you intended these to have line breaks or not but (again at least in pdf) they don't.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+available in https://commons.apache.org/proper/commons-jexl/reference/syntax.html[Apache Commons] as a reference 
+to create your queries.
+
+Other JEXL examples may include:
+
+*Selecting all 'raw' format content*
+format == 'raw'
+
+*Selecting all 'maven2' content with a group that starts with 'org.sonatype.nexus'*
+format == 'maven' and coodinate.group =^'org.sonatype.nexus'
```
joedragons: coordinate typo (currently shows coodinate)

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+fields. In the 'Specification' section use the 'Search expression' field to enter your JEXL query.
+
+Creating a new content selector allows you to identify what components or assets you want your user to have 
+permission to access.
+
+[[selector-examples]]
+===== Content Selector Examples
+
+If, for example, you want to create a pattern that would capture all components with any part of the path 
+containing the format 'maven2' you can define it in place of the variable 'format'. You can use the syntax
```
joedragons: I think "any part of the path containing" is wrong and can be struck.  That or you'd be using the variable 'path'.  I also think "in place of" should be "utilizing".  It's not like you remove format and put maven2, which is how I read "in place of".

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+
+To create a new content selector privilege click 'Privileges' located in the 'Administration' menu, under 
+'Security'. From the 'Privileges' screen click the 'Create privileges' button to view a list of privilege types. 
+Click 'Repository Content Selector' to get to a form that displays the following:
+
+* 'Name':: Create a name for the content selector privilege.
+* 'Description':: Add a brief description for the privilege.
+* 'Content Selector':: Use the dropdown to select from a list of selectors you created in <<selector-create>>.
+* 'Repository':: Choose from a list repositories you created, as documented in <<admin-repositories>>.
+* 'Actions':: Grant read, edit, delete, or '*' (all) privileges for user access control.
```
joedragons: Minor: I'm a little wary of redundancy from 5.3 (Privileges) here, but if you're ok, I'm ok.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+ 
+After you create content selector privileges, regulating what permissible data you want the user to access, you 
+can assign a role as documented in <<roles>>. When roles are defined, assign a user, aa mentioned in <<users>>.
```
joedragons: "aa" typo (guessing should be 'as'?)

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+
+With this setup you can delegate all control of components in +org.apache.commons+ to a "Maven" team. This 
+way, you will not need to create separate repositories for each logical division of your components.
```
joedragons: Suggest saying this is an example and/or changing "can" to "could" and/or changing "setup" to something more generic, since indeed the setup above is generic not specific to this example.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+
+[[selector-reference]]
+===== Content Selector Reference
+
+Below are the allowable attributes for content selectors. The table below defines 'path', 'format', and 
+'coordinate' as values supported by {nxrm}.
```
joedragons: Seems like these sentences could be merged.  I almost said just strike the first but it does reference "content selectors" (which could be unneeded since that's the section you're in) which you could put in the second sentence.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+
+[width="60%",frame="topbot",options="header,footer"]
+|=====================================================================
+|Attribute  |Allowed Values
+|path       |The path under a repository where your content is exposed
```
joedragons: Minor: Might switch "under" with "of".  I'm not really sure that's an improvement but under doesn't seem right to me.  Might just be a terminology thing I'm unaware of.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+|format     |The format of the content for which you query
+|coordinate |A map of attributes that differ by content format
+|=====================================================================
+
+This table contains a list of formats containing variables for coordinate name, group name, and format version.
```
joedragons: It's weird to me this says coordinate, group and version but those arn't paritied below.  To be clear, this is a critique of this sentence describing the table not the table itself.  I think it'd be better suited to say something like "table contains a list of attributes that can affect the respective formats".
That or this is a critique of the table too because that's what I was expecting/think would be useful here;)

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+|Bower    |name, version
+|Docker   |name, version
+|Maven2   |group, name, version
+|npm      |group, name, version
+|NuGet    |name, version
+|PyPI     |name, version
+|Raw      |name
+|RubyGems |name, version
+|===============================
+
```
joedragons: Similar to above note, this doesn't seem to do justice to the available variables (or soon to be available) via https://docs.sonatype.com/display/Nexus/Nexus+3+Component+Selector+Normalized+Attributes.  I know it doesn't do justice to Maven2 (having tested that).  I think a full list will be valuable here as I'm not sure where they'd get it otherwise, without having to dig through code (if even available) or super blind (as there are no hints) trial and error (or maybe if we ever add a script api).

chapter-configuration.asciidoc
dulanism: It might be better to re-write the (correct) into the intro, imo

chapter-configuration.asciidoc
dulanism: Clarifies my question earlier on if in fact 'format' is a classic variable; it's necessary to add the attribute for association.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
 across multiple repositories will be handled by the repository manager and therefore be much faster than client
 side resolution done by Maven each time.
 
+[[content-selectors]]
+==== Content Selectors
+{inall}
+
+Content selectors are a set of regular expressions used to match characters on the path of components in a
```
dbradicich: not sure we want to label content selectors simply as s set of regular expressions (which is exactly what they were in nx2),  there are some uses of regular expressions in the jexl language, but its just a portion.  Content selectors are now a means to select specific content that evaluates against a jexl expression.   Its also not just against the path, but also against coordinates of the content as well

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+==== Content Selectors
+{inall}
+
+Content selectors are a set of regular expressions used to match characters on the path of components in a 
+repository. Content selectors allow you to select specific content, with regular expressions that are defined by 
+you, from all of your content. The regular expressions are implemented with 
+https://commons.apache.org/proper/commons-jexl/[JEXL], a scripting language, that allows filtering and querying 
+on component patterns.
+
+Creating content selectors allow you to identify what selector privilege you can assign (or limit) to a
```
dbradicich: I would remove ```(or limit)``` the perms in the system are only additive (i.e. you grant access via perms, not remove access via perms)

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+
+Content selectors are a set of regular expressions used to match characters on the path of components in a 
+repository. Content selectors allow you to select specific content, with regular expressions that are defined by 
+you, from all of your content. The regular expressions are implemented with 
+https://commons.apache.org/proper/commons-jexl/[JEXL], a scripting language, that allows filtering and querying 
+on component patterns.
+
+Creating content selectors allow you to identify what selector privilege you can assign (or limit) to a 
+user. You can define, in a simplified example, a selector named Apache Maven with a search expression of 
+`^/org/apache/maven/.*`. This would match all components with the groupId `org.apache.maven` and any components
```
dbradicich: the noted expressiion should probably be ```path =~ `^/org/apache/maven/.*` ```

Also, should note this will match all components in the path.  If you wanted to check for groupId you could have a simpler expression like ```coordinate.groupId == 'org.apache.maven' ```

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+Content selectors are a set of regular expressions used to match characters on the path of components in a 
+repository. Content selectors allow you to select specific content, with regular expressions that are defined by 
+you, from all of your content. The regular expressions are implemented with 
+https://commons.apache.org/proper/commons-jexl/[JEXL], a scripting language, that allows filtering and querying 
+on component patterns.
+
+Creating content selectors allow you to identify what selector privilege you can assign (or limit) to a 
+user. You can define, in a simplified example, a selector named Apache Maven with a search expression of 
+`^/org/apache/maven/.*`. This would match all components with the groupId `org.apache.maven` and any components 
+within nested within the 'Group' such as `org.apache.maven.plugins`.
```
dbradicich: ~~within~~ nested within

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+
+[[selector-create]]
+===== Creating a New Content Selector
+
+To create a new selector query, click 'Content Selectors' located in 'Repository', from the
```
dbradicich: create a new content selector (just thinking we dont need to introduce another way to refer to these things)

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+To create a new selector query, click 'Content Selectors' located in 'Repository', from the 
+'Administration' menu. Alternately, click the 'Content Selectors' button located within the 'Repository' section. 
+In the 'Selector ID' section enter a 'Name' and (optional) 'Description' of your selector in the corresponding 
+fields. In the 'Specification' section use the 'Search expression' field to enter your JEXL query.
+
+[[selector-security]]
+===== Managing Security with Content Selectors
+
+Once you have created a content selector, you can utilize it as part of your security setup. You can add a new 
+privilege that relates to the target and controls operations such as read, edit, delete, and '*' (all), for
```
dbradicich: relates to the ~~target~~ content selector

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+fields. In the 'Specification' section use the 'Search expression' field to enter your JEXL query.
+
+[[selector-security]]
+===== Managing Security with Content Selectors
+
+Once you have created a content selector, you can utilize it as part of your security setup. You can add a new 
+privilege that relates to the target and controls operations such as read, edit, delete, and '*' (all), for 
+components matching that path. The privilege can even span multiple repositories.
+
+As explained in <<privileges>>, create a new custom content selector privilege. Then choose 'Repository Content
```
dbradicich: i think `custom` can be removed

chapter-configuration.asciidoc
dbradicich: I would include add here as well, worth also noting that this is (currently) a comma separated list if you want to do multiple actions at once

dbradicich: actually, maybe not sure about add... i guess leave that out for now, since add (at least currently) doesn't get you much like joe noted

dulanism:
@damian tweaked the intro. Thoughts?

> Content selectors provide a means for you to select specific content from all of your content. 
You can create a content selector filter to match characters on the path of components in a repository. The content you select is evaluated against a JEXL expression. [JEXL](https://commons.apache.org/proper/commons-jexl/) is a expression library you will use to script queries along specific paths and coordinates available to your repository manager formats.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+define, in a simplified example, a selector named Apache Maven with a search expression of `path =~ 
+^/org/apache/maven/.*`. This would match all components within the designated component path. You can refer to 
+more examples in <<selector-examples>>.
+
+[[selector-create]]
+===== Creating a New Content Selector
+
+To identify a new selector privilege, create the query first. Click 'Content Selectors' located in 'Repository', 
+from the 'Administration' menu. Alternately, click the 'Content Selectors' button located within the 'Repository' 
+section. In the 'Selector ID' section enter a 'Name' and (optional) 'Description' of your selector in the
```
joedragons: * Minor: This is the first time I recall seeing any mention of the alternate way (tho it's true for all our buttons).
* Minor: No mention of clicking 'Create Selector' here, commenting in case not intentional.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+^/org/apache/maven/.*`. This would match all components within the designated component path. You can refer to 
+more examples in <<selector-examples>>.
+
+[[selector-create]]
+===== Creating a New Content Selector
+
+To identify a new selector privilege, create the query first. Click 'Content Selectors' located in 'Repository', 
+from the 'Administration' menu. Alternately, click the 'Content Selectors' button located within the 'Repository' 
+section. In the 'Selector ID' section enter a 'Name' and (optional) 'Description' of your selector in the 
+corresponding fields. In the 'Specification' section use the 'Search expression' field to enter your JEXL query.
```
joedragons: Minor: Similarly, here, no mention of 'Create selector' button to create.  This whole section seems a little fast=)  Tho everything is true.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+
+[[selector-security]]
+===== Managing Security with Content Selectors
+
+Once you have created a content selector, you can utilize it as part of your security setup. You can add a new 
+privilege that controls operations such as read, edit, delete, and '*' (all), for components matching that path. 
+The privilege can even span multiple repositories.
+
+As explained in <<privileges>>, create a new content selector privilege. Then choose 'Repository Content
```
joedragons: I know I commented that this section was redundant but I think this is oversimplified as written now.  Basically the only thing "As explained in <<privileges>>, create a new content selector privilege." shortens is where they go to create it.  Which means they have to click somewhere else to find where to go then come back here to find out what to do.
Additionally some details I felt redundant (all the field names) are still here (below).
Despite my previous comment, comparitively, I prefer the old version (though I still think there's redundancy).

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+Once you have created a content selector, you can utilize it as part of your security setup. You can add a new 
+privilege that controls operations such as read, edit, delete, and '*' (all), for components matching that path. 
+The privilege can even span multiple repositories.
+
+As explained in <<privileges>>, create a new content selector privilege. Then choose 'Repository Content 
+Selector' to get to a form that displays the following:
+
+* 'Name':: Create a name for the content selector privilege.
+* 'Description':: Add a brief description for the privilege.
+* 'Content Selector':: Use the dropdown to select from a list of selectors you created in <<selector-create>>.
```
joedragons: Minor: Not convinced the value of the link (or maybe the phrasing); note that it's possible to get selectors from upgrade or that they're reading this without creating any just trying to figure out this section.  +0 on it.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+* 'Repository':: Choose from a list repositories you created, as documented in <<admin-repositories>>.
```
joedragons: Minor: Similar to above comment, I see what's being done here but "you created" isn't necessarily true (others or defaults or upgrade).  +0 on phrasing.

chapter-configuration.asciidoc
joedragons: Matching now, thanks=)

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+
+JEXL examples may include:
+
+*Selecting all 'raw' format content*
+
+'format == "raw"'
+
+*Selecting all 'maven2' content with a groupId that starts with 'org.apache.commons'*
+
+'format == "maven2" && coordinate.groupId =^ 'org.apache.commons.''
```
joedragons: As I believe we discussed via HC, I am not sure if single quotes will work.  If you've tested or got someone elses vet, no worries.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+
+*Selecting all 'nuget' package versions of EntityFramework*
+
+'format == "nuget" && path =~ '/packages/EntityFramework/*.''
```
joedragons: As I believe we discussed via HC, I am not sure if single quotes will work.  If you've tested or got someone elses vet, no worries.

Minor: I am also skeptical of the period here but I may be wrong.  Minor note.

joedragons:
+1.  Comments almost all minor.  Thanks!

chapter-configuration.asciidoc
dulanism: Fleshed this out to better define each step. I left the list of field on the "Create Repository Content Selector" (with minor rewrites) in order to call out the requirements and motivation of each. I agree with your comments above: adding the list instead of the linking away only to come back is basically a better reader experience.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
 across multiple repositories will be handled by the repository manager and therefore be much faster than client
 side resolution done by Maven each time.
 
+[[content-selectors]]
+==== Content Selectors
+{inall}
+
+Content selectors provide a means for you to select specific content from all of your content. You can 
+create a content selector filter to match characters on the path of components in a repository. The content you
```
bentmann: There's "content" and then "content selector" but what is a "content selector filter"?

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
 across multiple repositories will be handled by the repository manager and therefore be much faster than client
 side resolution done by Maven each time.
 
+[[content-selectors]]
+==== Content Selectors
+{inall}
+
+Content selectors provide a means for you to select specific content from all of your content. You can 
+create a content selector filter to match characters on the path of components in a repository. The content you 
+select is evaluated against a JEXL expression. https://commons.apache.org/proper/commons-jexl/[JEXL] is a
```
bentmann: "a**n** expression"

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+==== Content Selectors
+{inall}
+
+Content selectors provide a means for you to select specific content from all of your content. You can 
+create a content selector filter to match characters on the path of components in a repository. The content you 
+select is evaluated against a JEXL expression. https://commons.apache.org/proper/commons-jexl/[JEXL] is a 
+expression library you will use to script queries along specific paths and coordinates available to your 
+repository manager formats.
+
+Creating content selectors allow you to identify what selector privilege you can assign to a user. You can
```
bentmann: "Creating content selectors allow", missing "s" or superfluous "Creating", something doesn't fit.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+Creating content selectors allow you to identify what selector privilege you can assign to a user. You can 
+define, in a simplified example, a selector named Apache Maven with a search expression of `path =~ 
+^/org/apache/maven/.*`. This would match all components within the designated component path. You can refer to 
+more examples in <<selector-examples>>.
+
+[[selector-create]]
+===== Creating a New Content Selector
+
+To identify a new selector privilege, create the query first. Click 'Content Selectors' located in 'Repository', 
+from the 'Administration' menu. Alternately, click the 'Content Selectors' button located within the 'Repository'
```
bentmann: Probably meant "Alternatively"

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+* 'Name':: Create a name for the content selector privilege.
+* 'Description':: Add a brief description for the privilege.
+* 'Content Selector':: Use the dropdown to select from a list of selectors you created in <<selector-create>>.
+* 'Repository':: Choose from a list repositories you created, as documented in <<admin-repositories>>.
+* 'Actions':: Grant read, edit, delete, or '*' (all) privileges for user access control.
+ 
+After you create content selector privileges, regulating what permissible data you want the user to access, you 
+can assign a role as documented in <<roles>>. When roles are defined, assign a user, as mentioned in <<users>>.
+
+A practical example might be where you delegate all control of components in `org.apache.commons` to a "Maven"
```
bentmann: "org.apache.commons" vs "Maven", unless we're out to confuse, that probably means "org.apache.maven" instead as line 896 refers to.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+===== Content Selector Reference
+
+Below are the allowable attributes for content selectors that define 'path', 'format', and 'coordinate' as values 
+supported by {nxrm}.
+
+[width="60%",frame="topbot",options="header,footer"]
+|=============================================================
+|Attribute  |Allowed Values
+|path       |The path of your repository content
+|format     |The format of the content for which you query
```
bentmann: How does the end user know which format strings are valid/supported?

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+|coordinate |A map of attributes that differ by content format
+|=============================================================
+
+This table contains a description of attributes that can affect the respective formats.
+
+[width="60%",frame="topbot",options="header,footer"]
+|===============================================================================================
+|Format   |Coordinate Attributes
+|Maven2   |coordinate.groupId, coordinate.artifactId, coordinate.version, coordinate.classifier, coordinate.extension
```
bentmann: If that table is supposed to answer my above question about formats, it's misleading in presenting the formats with wrong case compared to what the JEXL expressions need.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,96 @@ repositories are made available to your development team. In addition the perfor
...
+|NuGet    |coordinate.id, coordinate.version
+|Raw      |(No coordinates)
+|===============================================================================================
+
+[[selector-examples]]
+===== Content Selector Examples
+
+If, for example, you want to create a pattern that would capture all components within the format 'maven2' you 
+can define it utilizing the variable 'format'. You can use the syntax available in 
+https://commons.apache.org/proper/commons-jexl/reference/syntax.html[Apache Commons] as a reference to create
```
bentmann: Why do we label the link "Apache Commons"? Apache Commons is an umbrella project (https://commons.apache.org/), the link here is for the "JEXL Syntax"

chapter-configuration.asciidoc
bentmann: https://commons.apache.org/proper/commons-jexl/reference/syntax.html#Literals
> Can start and end with either ' or " delimiters

chapter-configuration.asciidoc
bentmann: > I am also skeptical of the period here but I may be wrong

This using a regexp, he probably means ".*".

More generally though, I don't think `path`-based expressions are good examples for NuGet. When talking about packages/components, it's more robust to use a `coordinate`-based expression. NuGet is evolving its protocol and the example here assume their v2 API and the v3 API uses different paths (e.g. https://api.nuget.org/v3-flatcontainer/newtonsoft.json/6.0.4/newtonsoft.json.6.0.4.nupkg)

chapter-configuration.asciidoc
joedragons: I wanted to disagree, but apparently this grammar rule is a lot more confusing than I remembered=\

chapter-configuration.asciidoc
dulanism: Yes, you're right. I struck the sentence beginning with "Alternately" because I felt it was going down an unnecessary path of detail.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,103 @@ repositories are made available to your development team. In addition the perfor
...
+
+JEXL examples may include:
+
+*Select all 'raw' format content*
+
+'format == "raw"'
+
+*Select all 'maven2' content along a path that starts with 'org.apache.commons'*
+
+'format == "maven2" && path =^ "org/apache/commons/*"'
```
bentmann: The asterisk should not be here given the prefix isn't a regexp.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,103 @@ repositories are made available to your development team. In addition the perfor
...
+{inall}
+
+Content selectors provide a means for you to select specific content from all of your content. You can 
+create a selector to match characters on the path of components in a repository. The content you select is 
+evaluated against a JEXL expression. JEXL is an expression library used to script queries along specific 
+paths and coordinates available to your repository manager formats.
+
+Content selectors allow you to identify what selector privilege you can assign to a user. You can 
+define, in a simplified example, a selector named Apache Maven with a search expression of 'path =^ 
+"/org/apache/maven/*"'. This would match all components that 'start with' the designated component path.
```
dbradicich: as benjamin noted in example further down, the * isn't necessary in the example as =^ is 'starts with' comparison

dulanism: Noted this as an update... will fix. Thx

chapter-configuration.asciidoc
dulanism: Fixing

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,103 @@ repositories are made available to your development team. In addition the perfor
...
+[width="60%",frame="topbot",options="header,footer"]
+|=============================================================
+|Attribute  |Allowed Values
+|path       |The path of your repository content
+|format     |The format of the content for which you query
+|coordinate |A map of attributes that differ by content format
+|=============================================================
+
+This table contains a description of attributes that can affect the respective formats. You can review some 
+simple JEXL queries that call the listed attributes, against specific formats in <<selector-examples>>.
```
dulanism: Needs re-wording.

chapter-configuration.asciidoc
dulanism: Updated.

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
 across multiple repositories will be handled by the repository manager and therefore be much faster than client
 side resolution done by Maven each time.
 
+[[content-selectors]]
+==== Content Selectors
+{inall}
+
+Content selectors provide a means for you to select specific content from all of your content. You can 
+create a selector to match characters on the path of components in a repository. The content you select is
```
bentmann: "You can create a selector ...", do we really need this sentence given the next two ones (specifically line 891) also (and more completely) describe that "paths and coordiantes" are available for expressions?

chapter-configuration.asciidoc
```diff
@@ -881,6 +881,102 @@ repositories are made available to your development team. In addition the perfor
...
+////
+Add a section to talk about the selector preview feature in https://issues.sonatype.org/browse/NEXUS-10377
+////
+
+[[selector-permissions]]
+===== Managing Selector Permissions
+
+As part of your security setup, you can create user permissions to manage the filters you built in the 'Create 
+Selector' form. You can add a new privilege that controls operations such as read, edit, delete, and '*' 
+(all), for components matching that path. The privilege can even span multiple repositories.
```
bentmann: "components matching that **selector**", might be more accurate given it's not just about path.

chapter-configuration.asciidoc
bentmann: Hm, and pushed? Still says "Apache Commons" here.

chapter-configuration.asciidoc
dulanism: True. Struck!

bentmann:
+1

chapter-configuration.asciidoc
dulanism: Updated the first column in the table second table to elicit "available formats" to date.

chapter-configuration.asciidoc
dulanism: Updated to "JEXL guide".

PR #207, opened 2016-08-09
Correction to instruction that user needs both URL and nexus section in bowerrc file

dulanism:
Updated.
TODO: Cherrypick to 3.1


joedragons:
+1 (sorry for the delay; I wasn't confident so wanted to test).

PR #217, opened 2016-08-24
NEXUS-10659-iq-server-upgrade

dulanism:
JIRA: https://issues.sonatype.org/browse/NEXUS-10659

Notes:
- Cherry pick these changes into @jwayman's branch
- Added 'Health Check' to the list of Repository content that can be upgraded (see 'Upgrading Content')

- Added 'Crowd' config as s 'Content' option

### DIFF
--- a/chapter-upgrading.asciidoc
+++ b/chapter-upgrading.asciidoc
@@ -105,7 +105,7 @@ the file is, and they obtain it from the file system, directly.
 
 If upgrading {iq}, ensure that your licenses include the integration for both versions 2 and 3. Your 
 configuration for 'IQ Server URL', 'Username', 'Password', and 'Request Timeout' will be included in the upgrade. 
-Additionally, if configured, scanner properties, trust store usage, and the enabled {iq} connection itself will 
+Additionally, if configured, analysis properties, trust store usage, and the enabled {iq} connection itself will 
 be replicated from versions 2 to 3.
 
 *NuGet API Key*
--- a/chapter-upgrading.asciidoc
+++ b/chapter-upgrading.asciidoc
@@ -90,6 +90,21 @@ the file is, and they obtain it from the file system, directly.
 [[upgrade-content-prep]]
 === Preparing Content for Upgrade
 
+*IQ Server*
+
+{inrmonly}
+
+If upgrading your {iq} integration, ensure that your licenses for both versions 2 and 3 have the capability 
+enabled. Your configuration for 'IQ Server URL', 'Username', 'Password', and 'Request Timeout' will be included 
+in the upgrade. Additionally, scanner properties, trust store usage, and enabled {iq} connection can be 
+replicated from versions 2 to 3. As as best practice, be sure to encrypt your {iq} password before initiating 
+the upgrade.
+
+*NuGet API Key*
+
+The upgrade tool will add all keys to {nxrm} 3 that are present in {nxrm} 2.14 when asked, even if the version 2 
+NuGet API Key Realm is not active. This is because there is no explicit on or off setting for NuGet keys.
+
 *Repository IDs*
 
 Before migrating repositories to version 3, find the repository IDs in version 2 and plan how you will 
@@ -110,10 +125,6 @@ The upgrade tool can only replicate pre-existing user tokens from version 2 to 3
 link to user token chapter
 ////
 
-*NuGet API Key*
-
-The upgrade tool will add all keys to {nxrm} 3 that are present in {nxrm} 2.14 when asked, even if the version 2 
-NuGet API Key Realm is not active. This is because there is no explicit on or off setting for NuGet keys.
 
 *Repository Health Check and SSL Health Check*
 
--- a/chapter-upgrading.asciidoc
+++ b/chapter-upgrading.asciidoc
@@ -5,8 +5,9 @@
 [[upgrade-introduction]]
 === Introduction
 
-The upgrade feature provides a built-in supported method for replicating components from {nxrm} versions 2.14 to 
-3.1. Future iterations of this feature will be available for versions beyond both 2.14 and 3.1, alike.
+The upgrade feature provides a built-in supported method for replicating components and configurations from 
+{nxrm} versions 2.14 to 3.1. Future iterations of this feature will be available for versions beyond both 2.14 
+and 3.1, alike.
 
 This tool helps you build an upgrade plan which allows version 3 to talk to version 2 by way of a 
 capability that communicates with the upgrade tool. The plan you create with version 3 will provide a 
@@ -20,6 +21,10 @@ manager provides alternative methods to upgrade if you are currently running old
 back to milestone 7. See our support knowledge base to walk you through the steps to upgrade from 
 https://support.sonatype.com/hc/en-us/articles/222159808[milestone 7 of version 3 to 3.0.0 final] or 
 https://support.sonatype.com/hc/en-us/articles/217967608[version 3.0.0 to a later version].
+
+////
+Expand this intro to give readers a 10-foot view of the upgrade tool
+////
 
 [[upgrade-architecture]]
 === Supported Architecture for Upgrading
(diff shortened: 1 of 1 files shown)


dulanism:
Thinking, retrospectively, that I should add a note that the upgrade of this feature will not work if the user only has an NXRM3 trial license. Thoughts?

joedragons:
@dulanism Are there NX3 trial licenses yet?  I thought that wasn't done.  Maybe I'm misremembering.
Also (without building) looks like the PR does not add to the list of items that you can migrate further down in the guide.  Just adds detail about the area.  I can quote better once I build if this is unclear.

dulanism:
@joedragons I'm the mis-rememberer, as I now looking back as some notes. Thanks for the heads up, about the content list. I'll update in the next commit.

chapter-upgrading.asciidoc
```diff
@@ -90,6 +90,21 @@ the file is, and they obtain it from the file system, directly.
 [[upgrade-content-prep]]
 === Preparing Content for Upgrade
 
+*IQ Server*
+
+{inrmonly}
+
+If upgrading your {iq} integration, ensure that your licenses for both versions 2 and 3 have the capability 
+enabled. Your configuration for 'IQ Server URL', 'Username', 'Password', and 'Request Timeout' will be included
```
joedragons: +0 to first sentence as I cannot remember if this is true.  I am pretty sure other config just migrates over and no pre-treatment is needed.  Looks like @bentmann tested this maybe he remembers.  I can check if you like, let me know.  If you've experienced this then you can ignore this comment=)

chapter-upgrading.asciidoc
```diff
@@ -90,6 +90,21 @@ the file is, and they obtain it from the file system, directly.
 [[upgrade-content-prep]]
 === Preparing Content for Upgrade
 
+*IQ Server*
+
+{inrmonly}
+
+If upgrading your {iq} integration, ensure that your licenses for both versions 2 and 3 have the capability 
+enabled. Your configuration for 'IQ Server URL', 'Username', 'Password', and 'Request Timeout' will be included 
+in the upgrade. Additionally, scanner properties, trust store usage, and enabled {iq} connection can be
```
joedragons: You say "can" here is it optional?  I don't recall an option.  Alternatively, you might be saying/inferring these fields may be blank and thus arn't technically migrated as blank.  If that's so, might be worth a rephrase to avoid the misinterpretation stated above.

chapter-upgrading.asciidoc
```diff
@@ -90,6 +90,21 @@ the file is, and they obtain it from the file system, directly.
...
 === Preparing Content for Upgrade
 
+*IQ Server*
+
+{inrmonly}
+
+If upgrading your {iq} integration, ensure that your licenses for both versions 2 and 3 have the capability 
+enabled. Your configuration for 'IQ Server URL', 'Username', 'Password', and 'Request Timeout' will be included 
+in the upgrade. Additionally, scanner properties, trust store usage, and enabled {iq} connection can be 
+replicated from versions 2 to 3. As as best practice, be sure to encrypt your {iq} password before initiating
```
joedragons: Should be encrypted by us already, I think.  Maybe dev (or you) know more than me.

bentmann: What Joe said. Really not sure what message we try to get across, pretty much any piece of Nexus UI that I remember expects end users to enter credentials as is, i.e. in clear, it's up to the code to encrypt/obfuscate those passwords when persisting them or as here when transmitting them.

chapter-upgrading.asciidoc
bentmann: One cannot even select the IQ config for upgrade if the license in NX3 doesn't enable it (cf. https://github.com/sonatype/nexus-internal/commit/ebf37c20517d1211e6480c8b324e60ab7804ba40#diff-274d35362e4e7f6e304ff7af4e518668R77). If the license in NX2 doesn't enable IQ it will just look like IQ isn't configured (cf. https://github.com/sonatype/nexus2-internal/commit/acf91a42695c92f203e26f8ac1d9300d5c1dee38#diff-2a8caff70f970cf2a042c36d004b91a1R75)

joedragons: Ok, cool, then I think this sentence is fine.  Thanks @bentmann .

dulanism: Ah, good to know. I think that should be noted. In the process of testing. Thanks.

bentmann: What I described is the implementation, I don't see why an end user would care about that level of detail. I find your current sentence rather sufficient.

chapter-upgrading.asciidoc
```diff
@@ -90,6 +90,21 @@ the file is, and they obtain it from the file system, directly.
 [[upgrade-content-prep]]
 === Preparing Content for Upgrade
 
+*IQ Server*
+
+{inrmonly}
+
+If upgrading your {iq} integration, ensure that your licenses for both versions 2 and 3 have the capability
```
bentmann: "capability" already has a certain and different meaning in NX. For licenses, we typically speak of features AFAIK.

chapter-upgrading.asciidoc
dulanism: Consider this update: 

> Additionally, if configured, scanner properties, trust store usage, and the enabled {iq} 
connection itself will be replicated from versions 2 to 3.

chapter-upgrading.asciidoc
```diff
@@ -90,6 +99,20 @@ the file is, and they obtain it from the file system, directly.
 [[upgrade-content-prep]]
 === Preparing Content for Upgrade
 
+*IQ Server*
+
+{inrmonly}
+
+If upgrading {iq}, ensure that your licenses include the integration for both versions 2 and 3. Your
```
bentmann: "If upgrading Nexus IQ Server", we might want to add "settings/configuration" or the like to make it clear that this is about the configuration in NX, not the IQ Server itself.

chapter-upgrading.asciidoc
```diff
@@ -90,6 +99,20 @@ the file is, and they obtain it from the file system, directly.
 [[upgrade-content-prep]]
 === Preparing Content for Upgrade
 
+*IQ Server*
+
+{inrmonly}
+
+If upgrading {iq}, ensure that your licenses include the integration for both versions 2 and 3. Your 
+configuration for 'IQ Server URL', 'Username', 'Password', and 'Request Timeout' will be included in the upgrade. 
+Additionally, if configured, scanner properties, trust store usage, and the enabled {iq} connection itself will
```
bentmann: Last I heard, "scan" or "scanner" are frowned upon to describe our tooling to the public. Potential alternatives could be "analysis properties" or "custom configuration properties".

bentmann:
+1

chapter-upgrading.asciidoc
joedragons: Looks good.

chapter-upgrading.asciidoc
```diff
@@ -90,6 +99,20 @@ the file is, and they obtain it from the file system, directly.
 [[upgrade-content-prep]]
 === Preparing Content for Upgrade
 
+*IQ Server*
+
+{inrmonly}
+
+If upgrading {iq} settings and configuration, ensure that your licenses include the integration for both versions 
+2 and 3. Your configuration for 'IQ Server URL', 'Username', 'Password', and 'Request Timeout' will be included 
+in the upgrade. Additionally, if configured, analysis properties, trust store usage, and the enabled {iq}
```
joedragons: Minor (comma drama): I'm not sure about "proper usage" but visually, removing the comma before if I think makes the conditionals look less like a part of the immediately following comma deliniated list.

chapter-upgrading.asciidoc
```diff
@@ -265,17 +284,13 @@ base URL (including context path) of your version 2 server (e.g. +http://localho
...
 
 Content:: This screen displays checkboxes for compatible component formats ('Repositories'), security features 
-('Security'), and server configuration ('System'). For 'Repositories' you can select 'User-Managed Repositories'  
-and 'Repository Targets', the latter of which are regular expressions that upgrade to JEXL format, used for 
-content selectors. For 'Security' you can choose from 'Anonymous', 'LDAP Configuration', 'NuGet API-Key', 
-'Realms', 'Roles', 'SSL Certificates', 'Target Privileges', 'Users', and 'User Tokens'. For 'System' you can 
-select 'Email' and 'HTTP Configuration' contents.
+('Security'), and server configuration ('System'). For 'Repositories' you can select 'User-Managed Repositories', 
+'Repository Targets', and 'Health Check'. For 'Security' you can choose from 'Anonymous', 'LDAP Configuration', 
+'NuGet API-Key', 'Realms', 'Roles', 'SSL Certificates', 'Target Privileges', 'Users', and 'User Tokens'. For
```
joedragons: Crowd is missing here, but I believe there's a distinct ticket for that.  If you're not going to do more than add to list, this could be easily done here saving you another PR (tho maybe that's taboo=)).

joedragons:
+1.  It is interesting to note that your list of content does not 1-1 match the app because yours is alphabetical and the app is not.  If anything, I think that's a fault of the app tho.

PR #223, opened 2016-08-30
Unticketed search improvements

joedragons:
Noticed one while testing and more on review, none of these have
tickets associated that I know of.

* Added PyPI search specifics
* Removed npm redundancy: it listed generics as specifics where there
was no different from the generics.  If this cull is -1ed for any
reason, note that it is also incorrect.  It lists 'group' but npm shows
'scope'.
* Made npm search description consistent with others: adding format
* Removed extra newline
* Hiding incorrect age and popularity information and updated
screenshot until restored (ref
https://issues.sonatype.org/browse/NEXUS-10770)

Bonus:
* Adjusted some screenshot (non-search) sizing to make pdf seeable
* Made a 2 line comment onto 1 line (still 114ized)


joedragons:
Two things I did not change but am checking with @dulanism about:
* It struck me as a little odd that the search examples in 3.3.1 were not alphabetized.  I left my addition on the end but would alphabetize if decided worth.
* I think the format example is incomplete and feel like it should be struck.  It could be updated or left incomplete alternatively.
This is complete to my satisifaction otherwise.

joedragons:
Saw @dulanism 's thumbs up, going to make those final changes.

dulanism:
+1 officially

PR #227, opened 2016-09-07
NEXUS-10664-systemd-correction

dulanism:
Also, line break fixes
JIRA: https://issues.sonatype.org/browse/NEXUS-10664


joedragons:
+1 esp since you tested it (I did not)

PR #229, opened 2016-09-11
[NEXUS-10617] ha-config-startup

dulanism:
JIRA: https://issues.sonatype.org/browse/NEXUS-10617
CI: http://bamboo.s/browse/BOOK-NEXUS193

### DIFF
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -22,7 +22,13 @@ and components utilized by globally distributed teams.
 [[why-component-fabric]]
 === Why Enable Component Fabric?
 
-Section TBD - In short component fabric provides/enables the backend that realizes high availability
+Component Fabric is, essentially, a storage backend for multiple {nxrm} instances.
+////
+expand later.
+Today organizations are using multiple repository manager installations in order to scale the required 
+infrastructure and optimize the performance and availability of components for all consumers and producers.
+////
+
 
 [[characteristics-component-fabric]]
 === Characteristics of Component Fabric
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -7,9 +7,9 @@
 
 Using {pro}'s high availability feature is an integral part of the repository manager as it reduces the 
 complexity of deployment and infrastructure, while simultaneously improving scalability, user experience, and 
-performance. Active-active high availability allows you to create a performant backend with built-in support for active-active 
-high availability configurations, enhanced security, 
-and license analysis for open source components.
+performance. Active-active high availability allows you to create a performant backend with built-in support for 
+active-active high availability configurations, enhanced security, and license analysis for open source 
+components.
 
 With high availability brought to light this feature sets the standard for a continuously operational application 
 to reduce points of failure among content within your repository manager. Ultimately, this particular feature 
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -6,16 +6,18 @@
 Update chapter name, sections to component fabric after sorting out sections
 ////
 
-[[ha-introduction]]
+[[fabric-introduction]]
 === Introduction
 
-High availability for {nxrm} sets the standard for a continuously operational application to reduce points of 
-failure among features and components within your repository manager. The component fabric in {pro}, which 
-enables high availability, acts as a private binary cloud storage and backend for all your components.
+Component Fabric is an integral part of {pro} that reduces the complexity of deployment and infrastructure, while 
+simultaneously improving scalability, user experience, and performance. Component fabric allows you to create a 
+performant backend with built-in support for active-active high availability configurations, enhanced security, 
+and license analysis for open source components.
 
-Component Fabric is an integral part of {pro} reduces the complexity of deployment and infrastructure, while 
-simultaneously improving performance, scalability, and user experience. You can activate as many repository 
-managers connected to your component fabric as you like and move your infrastructure to the next level.
+With high availability brought to light, Component Fabric will set the standard for a continuously operational 
+application to reduce points of failure among content within your repository manager. Ultimately, this particular 
+feature acts as a private binary cloud storage and backend for the entire suite of repository manager features 
+and components utilized by globally distributed teams.
 
 [[why-component-fabric]]
 === Why Enable Component Fabric?
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -27,9 +27,6 @@ line, you can expand the amount of clustered servers along your overall system b
 [[how-high-availability]]
 === How is High Availability Implemented?
 
-The high availability feature provides a storage backend for two {nxrm} instances colocated in the same data 
-center.
-
 High availability implementations are established between two nodes, joined together by a common blob store 
 directory. The second node that you initiate is synchronized with the first. When configured, the nodes 
 communicate with each other to create continuous uptime and the assurance that if one goes down, the second node 
@@ -55,9 +52,12 @@ of components even before a client tool requests it, leading to greater level of
 [[high-availability-expectations]]
 === Requirements for Clustering in {pro}
 
-{pro} is flexible and can lend itself to a variety of different system architectures. So, before enabling this 
-specific feature, determine which option is best for your existing architecture and system configuration before 
-installing. There are three supported infrastructures you can consider to facilitate high availability between 
+To make high availability work, you must have an existing repository manager already configured to start up. 
+Then, ensure that your existing repository manager has the same network configuration as the second repository 
+manager that you'll need to install. The two repository managers must run on different ports. Enabling the second 
+repository manager is the minimum requirement to allow redundancy between the nodes.
+
+There are three supported infrastructures you can consider to facilitate high availability between 
 your repository managers. You can install them across a mix of two:
 
 - physical servers
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -20,9 +20,9 @@ The biggest draw to a high performing, well-synchronized backend is a self-maint
 disaster recovery features. Downtime can be costly to organizations that host a multitude of applications.
 Therefore, an automated backup plan with distributed clusters decrease the chances of data loss. 
 
-With backup support, you can use high availability methodologies to resolve load balancing issues and implement
-failover initiatives for physical and virtual servers alike. Currently, you can only synchronize two nodes. But 
-down the line, you can expand the amount of clustered servers along your overall system backup plan.
+With backup support, you can use high availability methodologies to adopt load balancing and implement failover 
+initiatives for physical and virtual servers alike. Currently, you can only synchronize two nodes. But down the 
+line, you can expand the amount of clustered servers along your overall system backup plan.
 
 [[how-high-availability]]
 === How is High Availability Implemented?
@@ -30,17 +30,15 @@ down the line, you can expand the amount of clustered servers along your overall
 The high availability feature provides a storage backend for two {nxrm} instances colocated in the same data 
 center.
 
-High availability implementations are made among two nodes, joined together by a common blob store directory. The 
-second node that you initiate is synchronized with the first. When configured, the nodes communicate with each 
-other to provide uptime and the assurance that if one goes down a duplicated node can be used.
+High availability implementations are established between two nodes, joined together by a common blob store 
+directory. The second node that you initiate is synchronized with the first. When configured, the nodes 
+communicate with each other to create continuous uptime and the assurance that if one goes down, the second node 
+can be used.
 
-Instead of manual adjustments to your infrastructure, processes, and tools associated with your repository 
-manager installations, the high availability feature provides a replicated storage system. This eliminates 
-complexity and offers a simpler way to provision, configure, manage, and maintain multiple repositories.
-
-This feature is enabled across your nodes via link:https://hazelcast.com/[Hazelcast]. Hazelcast provides an 
-active/active approach to server clustering. It comes equipped with an interface for distributed data structures 
-and other aspects of in-memory computing.
+This feature is enabled across your nodes via link:https://hazelcast.com/[Hazelcast], which provides in-memory 
+computing for active data and active backup. So, instead of making manual adjustments to your servers' 
+infrastructure, processes, and tools, high availability via Hazelcast simplifies the configuration needed to 
+maintain multiple repositories.
 
 When enabling active/active high availability in your two repository manager instances, they act as individual 
 nodes, fully-aware of their setup parameters such as:
@@ -107,7 +105,7 @@ repository manager port.
 Follow these steps to enable the high availability feature:
 
 . In your existing repository manager, open the `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file.
-. Remove the `#` before `nexus.clustered = true` to enable Hazelcast usage on start. 
+. Remove the `#` before +nexus.clustered=true+ to enable Hazelcast usage on start. 
 . Go to the new repository manager and repeat steps 1 and 2, to enable the new instance.
 
 Any further work is done automatically by the interactions of the new repository manager with the data and 
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -91,8 +91,8 @@ pre-fetching of components from the fabric even before a client tool requests it
 move the section above to the intro
 ////
 
-[[ha-config]]
-=== HA Configuration and Startup
+[[fabric-configuration]]
+=== Configuring Component Fabric
 
 {nxrm} installation consists of very few simple steps:
 
@@ -103,13 +103,16 @@ move the section above to the intro
 Any further work is done automatically by the interactions of the new repository manager with the data and 
 configuration in the component fabric.
 
+[[fabric-start]]
+=== Starting Component Fabric
+
 To start your configuration, download a binary as explained in <<install-sect-downloading>>. Create two distinct 
 folders to store duplicate instances of the repository manager. For example, you can create two directories 
 `nexus` and `nexus2`. However, you have not yet installed {pro}, review the installation steps from the 
 beginning, as explained in <<install-introduction>>.
 
-In each distinct folders, unpackage the `tar.gz`. In the `nexus` folder, edit the 
-`$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file. Change the `application-port={port number}` in your first instance 
+In each distinct folders, unpackage the `tar.gz`.In the `nexus` folder, edit the 
+`$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file. Change the `application-port={port-number}` in your first instance 
 of {nxrm}. Then remove the `#` before `nexus.clustered = true`, enabling Hazelcast usage on start. Go to you the 
 second `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file and repeat the same steps.
 
@@ -126,10 +129,6 @@ Members [2] {
     Member [192.168.99.1]:5702 this
 }
 ----
-
-////
-Review Brad's demo on Docker startup, add new section 
-////
 
 [[after-fabric-startup]]
 === After Starting Up Component Fabric
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -17,72 +17,64 @@ creating a binary storage for the entire suite of features and components availa
 === Why Cluster?
 
 The biggest draw to a high performing, well-synchronized backend is a self-maintaining system with improved 
-disaster recovery features. Downtime can be costly to organizations that hosts a multitude of applications.
-Therefore, an automated backup plan with distributed clusters is preventative of any worse case scenarios around 
-data loss. 
+disaster recovery features. Downtime can be costly to organizations that host a multitude of applications.
+Therefore, an automated backup plan with distributed clusters decrease the chances of data loss. 
 
-With backup support, you can apply high availability methodologies that solve load balancing issues and implement
+With backup support, you can use high availability methodologies to resolve load balancing issues and implement
 failover initiatives for physical and virtual servers, alike. Currently, you can only synchronize two nodes. But 
-down the line, you can expand the amount of clustered servers you like to introduce to your overall system backup 
-plan.
+down the line, you can expand the amount of clustered servers along your overall system backup plan.
 
-////
-... or some better way to hint at component fabric
-////
+[[how-high-availability]]
+=== How is High Availability Implemented?
 
-[[high-availability-purpose]]
-=== What is the Purpose of High Availability?
-
-The high availability feature provides a storage backend for two {nxrm} instances co-located in the same data 
+The high availability feature provides a storage backend for two {nxrm} instances colocated in the same data 
 center.
 
-High availability implementations are made of servers, joined together but a common directory. The second cluster 
-is implemented to achieve synchronization between both enabled clusters. The clusters (or servers, or nodes) 
-communicate with each other to provide uptime and the assurance that if one goes down a duplicated node can be 
-used.
+High availability implementations are made among two nodes, joined together by a common blob store directory. The 
+second node that you initiate is synchronized with the first. When configured, the nodes communicate with each 
+other to provide uptime and the assurance that if one goes down a duplicated node can be used.
 
-Instead of manually managing your infrastructure, processes, and tools associated with numerous repository 
-manager installations, the repository manager's high availability feature provides a replicated storage system 
-that eliminates complexity and offers a simple and cost effective way to provision, configure, manage, and 
-maintain multiple repositories.
+Instead of manual adjustments to your infrastructure, processes, and tools associated with your repository 
+manager installations, the high availability feature provides a replicated storage system. This eliminates 
+complexity and offers a simpler way to provision, configure, manage, and maintain multiple repositories.
 
-This feature is enabled across your nodes via Hazelcast. Hazelcast provides an active/active approach, with an 
-interface for distributed data structures and other aspects of in-memory computing.
+This feature is enabled across your nodes via link:https://hazelcast.com/[Hazelcast]. Hazelcast provides an 
+active/active approach to server clustering. It comes equipped with an interface for distributed data structures 
+and other aspects of in-memory computing.
 
-When enabling active/active high availability in your two repository manager instances, they act as 
-individual nodes, fully-aware of their setup parameters such as:
+When enabling active/active high availability in your two repository manager instances, they act as individual 
+nodes, fully-aware of their setup parameters such as:
 
 - Memory availability and processor performance
 - Network connectivity performance
 - Characteristics of local storage in terms of size and performance
 
-This awareness also translates to increased efficiency, scalability, and control, allowing nodes to automatically 
-choose the fastest route to intercept component data between the two nodes. With additional customization, these 
-purpose-driven nodes can configure prefetching of components even before a client tool requests it, leading to 
-greater level of performance.
+This node-driven awareness within your clustered repository manager translates to increased efficiency, 
+scalability, and control, allowing the nodes to automatically choose the fastest route to intercept component 
+data between the two nodes. With additional customization, these purpose-driven nodes can configure prefetching 
+of components even before a client tool requests it, leading to greater level of performance.
 
 [[high-availability-expectations]]
 === Requirements for Clustering in {pro}
 
-{pro} is flexible, and can lend itself to a variety of different system architectures. So, before enabling this 
+{pro} is flexible and can lend itself to a variety of different system architectures. So, before enabling this 
 specific feature, determine which option is best for your existing architecture and system configuration before 
 installing. There are three supported infrastructures you can consider to facilitate high availability between 
-your repository managers. You can install them on two:
+your repository managers. You can install them across a mix of two:
 
-- separate physical servers
+- physical servers
 - virtual machines
 - Docker instances
 
-Your should recognize any challenges you may encounter when planning high availability implementation. You also 
-need to be mindful of cluster performance, so your repository manager can remain at its most optimal speed.
+You should to be mindful of server performance, so your repository manager nodes can remain at its most optimal 
+speed. Ultimately, the rule of thumb is you ensure the two nodes utilized share a file system for blob storage 
+and that they exist on the same network. Subsequently, to enable your nodes for high availability follow these 
+steps:
 
-Ultimately, the rule of thumb is you ensure the two nodes utilized share a file system for blob storage and that 
-they exist on the same network. Subsequently, the key goals for enabling high availability are as follows:
-
-- <<high-availability-install,Installing a secondary version>> of {pro}.
-- <<high-availability-configuration,Configuring the second binary>> as its own independent server.
-- <<high-availability-enable,Activating the clusters>> from the file `org.sonatype.nexus.cfg`.
-- <<high-availability-startup,Starting>> and running the service upon activation.
+- <<high-availability-install,Install a second instance>> of {pro}.
+- <<high-availability-configuration,Configure the second binary>> as its own independent server.
+- <<high-availability-enable,Activate the clusters>> from the file `org.sonatype.nexus.cfg`.
+- <<high-availability-startup,Start>> and run the service after configuration.
 
 [[high-availability-install]]
 ==== Installing a New Binary
@@ -110,7 +102,7 @@ nodes. To do this, you must distinguish the `application-port` of your existing 
 repository manager port.
 
 [[high-availability-enable]]
-==== Enabling the Clusters
+==== Enabling the Nodes
 
 Follow these steps to enable the high availability feature:
 
@@ -138,9 +130,17 @@ Members [2] {
 }
 ----
 
-[[high-availability-storage]]
-==== Managing Nodes and Storage
+[[high-availability-verify]]
+=== Verifying Synchronization
 
-With your two instances running, you can refer to the repository manager's user interface to view the nodes 
-activated by the high availability feature. Go to the 'Nodes' screen under 'System' located in the 
-'Administration' menu. Each machine should have a different node enabled as "true".
+At runtime, when you're log in as an administrator on one node, the other node automatically pairs the 
+administrator to the second node. The {nxrm} interface allows you to see the nodes paired.
+
+To verify this connection, go to the 'Nodes' screen, under 'System' located in the 'Administration' menu. This 
+screen provides details of the node, detected as the backup server. 
+
+For example, when you look at the row representing "node-a", you can see its 'UUID' (a unique identifier) 
+as well as the label 'true'. The 'true' statement, in the 'Local' column, indicates the "node-b" is synchronized 
+for backup. So, the moment you create and save a new repository called 'npm-internal' in "node-a", 'npm-internal' 
+will appears in "node-b".
+
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -5,20 +5,16 @@
 [[high-availability-introduction]]
 === Introduction
 
-{nxrm}'s high availability feature provides clustering for two separate instances of the repository manager. This 
-is used to create shared storage for a more fail-safe backend in your repository manager, minimizing downtime and 
+{nxrm}'s high availability feature provides a more fail-safe repository manager, minimizing downtime, and 
 reducing points of system failure.
 
-This chapter covers the benefits and steps needed to implement high availability. Usage of clusters set the 
+This chapter covers the benefits and steps needed to implement high availability. Usage of clusters sets the 
 standard for automating the process of continuous uptime, relieving teams of manual recovery techniques, and 
 creating a binary storage for the entire suite of features and components available in your repository managers.
 
-[[why-high-availability]]
-=== Why Cluster?
-
 The biggest draw to a high performing, well-synchronized backend is a self-maintaining system with improved 
 disaster recovery features. Downtime can be costly to organizations that host a multitude of applications.
-Therefore, an automated backup plan with distributed clusters decrease the chances of data loss. 
+Therefore, an automated backup plan with distributed nodes decreases the chances of data loss. 
 
 With backup support, you can use high availability methodologies to adopt load balancing and implement failover 
 initiatives for physical and virtual servers alike. Currently, you can only synchronize two nodes. But down the 
@@ -37,17 +33,15 @@ computing for active data and active backup. So, instead of making manual adjust
 infrastructure, processes, and tools, high availability via Hazelcast simplifies the configuration needed to 
 maintain multiple repositories.
 
-When enabling active/active high availability in your two repository manager instances, they act as individual 
-nodes, fully-aware of their setup parameters such as:
+With active/active high availability in your two repository manager instances, you enable two interchangeable
+nodes that act transparently as one instance. This makes them fully-aware of their setup parameters such as:
 
 - Memory availability and processor performance
 - Network connectivity performance
 - Characteristics of local storage in terms of size and performance
 
-This node-driven awareness within your clustered repository manager translates to increased efficiency, 
-scalability, and control, allowing the nodes to automatically choose the fastest route to intercept component 
-data between the two nodes. With additional customization, these purpose-driven nodes can configure prefetching 
-of components even before a client tool requests it, leading to greater level of performance.
+This node-driven awareness makes your clustered repository manager more scalable, allowing the nodes to 
+automatically intercept component data between the two nodes.
 
 [[high-availability-expectations]]
 === Requirements for Clustering in {pro}
@@ -87,8 +81,8 @@ same shared storage. You need to share the nodes the directory location of your
 <<admin-repository-blobstores,blob store>>. You shouldn't configure your nodes to share an entire data directory. 
 Each node should have its own independent non-conflicting OrientDB database.
 
-For example, if you have two nodes -- a default called "node-a" and a new node called "node-b" -- and "node-a's"  
-blob store directory located at `/foo/bar/blobs/foobar`. "Node-b" should be mounted to the same file system 
+For example, if you have two nodes -- a default called "node-a" and a new node called "node-b" -- and blob store 
+directory for "node-a" is located at `/foo/bar/blobs/foobar`. "Node-b" should be mounted to the same blob store
 location.
 
 [[high-availability-configuration]]
@@ -102,22 +96,22 @@ repository manager port.
 [[high-availability-enable]]
 ==== Enabling the Nodes
 
-Follow these steps to enable the high availability feature:
+Follow these steps to enable high availability:
 
 . In your existing repository manager, open the `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file.
-. Remove the `#` before +nexus.clustered=true+ to enable Hazelcast usage on start. 
+. Remove the `#` before +nexus.clustered=true+ to enable high availability in the new instance. 
 . Go to the new repository manager and repeat steps 1 and 2, to enable the new instance.
 
 Any further work is done automatically by the interactions of the new repository manager with the data and 
-configuration synchronized with the high availability feature.
+configuration synchronized with the high availability.
 
 [[high-availability-startup]]
 ==== Starting the High Availability Service
 
 At this point, you have already configured a new, distinct repository manager to run as a backup for the original 
-instance. So, start both repository managers, by entering and executing `./bin/nexus run` in the command line. On 
-startup you can review the log, in the command line. You should see connection messages via Hazelcast similar 
-to this, below:
+instance. To start, bring the nodes to full service by starting the repository managers in the command line. On 
+startup you can review the log, in the command line. You should see connection messages via Hazelcast similar to 
+this, below:
 
 ----
 2016-06-28 17:34:26,577-0400 INFO  [hz.nexus.generic-operation.thread-1] *SYSTEM com.hazelcast.cluster.ClusterService - [192.168.99.1]:5702 [nexus] [3.5.3]
--- a/chapter-high-availability.asciidoc
+++ b/chapter-high-availability.asciidoc
@@ -17,7 +17,7 @@ and license analysis for open source components.
 With high availability brought to light, Component Fabric sets the standard for a continuously operational 
 application to reduce points of failure among content within your repository manager. Ultimately, this particular 
 feature acts as a private binary cloud storage and backend for the entire suite of repository manager features 
-and components utilized by globally distributed teams.
+and components utilized by globally distributed teams. 
 
 [[why-fabric]]
 === Why Enable Component Fabric?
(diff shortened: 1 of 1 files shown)


chapter-high-availability.asciidoc
```diff
@@ -22,6 +22,15 @@ managers connected to your component fabric as you like and move your infrastruc
 
 Section TBD - In short component fabric provides/enables the backend that realizes high availability
 
+[[characteristics-component-fabric]]
+=== Characteristics of Component Fabric
+
```
mosabua: You can just link to the HA done right whitepaper for starters and not add much content here. Keep it very simple ... its a storage backend for multiple Nexus Repository Manager instances..

dulanism: Will do. Still in progress, watching demos for language/continuity.

joedragons:
Perhaps this has already been discussed, but I expressed surprise this was 3.1 branch.  I thought this was 3.1 EA then 3.2.  I suppose we could 3.1 branch then comment out after EA if I'm right.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
+{inrmonly}
 
-[[ha-introduction]]
+[[high-availability-introduction]]
 === Introduction
 
-High availability for {nxrm} sets the standard for a continuously operational application to reduce points of 
-failure among features and components within your repository manager. The component fabric in {pro}, which 
-enables high availability, acts as a private binary cloud storage and backend for all your components.
+{pro}'s high availability feature provides clustering for two separate instances of the repository manager. This
```
joedragons: Minor: Personally I think "Pro's" looks weird, I think {nxrm} is preferable to {pro} here.

dulanism:
I'll move it after review, @joedragons.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
 
-[[ha-introduction]]
+[[high-availability-introduction]]
 === Introduction
 
-High availability for {nxrm} sets the standard for a continuously operational application to reduce points of 
-failure among features and components within your repository manager. The component fabric in {pro}, which 
-enables high availability, acts as a private binary cloud storage and backend for all your components.
+{pro}'s high availability feature provides clustering for two separate instances of the repository manager. This 
+feature enables a quicker backend among hosts, minimizing downtime and reducing points of system failure.
```
joedragons: "quicker backend among hosts" - not sure what this means, maybe others do.  If I strictly interpret it, my understanding is each server back end will be as fast as they were before connected by HA.  Maybe some other meaning.

Minor: You might say "overall system" since technically one system (server) can still fail.  It might be implied that the system is the collection of servers, in which case this makes sense tho.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
 
-Component Fabric is an integral part of {pro} reduces the complexity of deployment and infrastructure, while 
-simultaneously improving performance, scalability, and user experience. You can activate as many repository 
-managers connected to your component fabric as you like and move your infrastructure to the next level.
+This chapter covers the benefits and steps needed to implement high availability in your repository manager.
```
joedragons: Minor: "repository manager cluster"?

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
-failure among features and components within your repository manager. The component fabric in {pro}, which 
-enables high availability, acts as a private binary cloud storage and backend for all your components.
+{pro}'s high availability feature provides clustering for two separate instances of the repository manager. This 
+feature enables a quicker backend among hosts, minimizing downtime and reducing points of system failure.
 
-Component Fabric is an integral part of {pro} reduces the complexity of deployment and infrastructure, while 
-simultaneously improving performance, scalability, and user experience. You can activate as many repository 
-managers connected to your component fabric as you like and move your infrastructure to the next level.
+This chapter covers the benefits and steps needed to implement high availability in your repository manager. 
+Usage of high availability clusters set the standard for automating the process, relieving teams of manual
```
joedragons: Comment: Not sure "the process" of what.  Maybe technical detail over my head.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
-simultaneously improving performance, scalability, and user experience. You can activate as many repository 
-managers connected to your component fabric as you like and move your infrastructure to the next level.
+This chapter covers the benefits and steps needed to implement high availability in your repository manager. 
+Usage of high availability clusters set the standard for automating the process, relieving teams of manual 
+recovery techniques, and creating a binary storage for the entire suite of repository manager features and 
+components.
 
-[[ha-node-performance]]
-=== Node Performance Considerations
+IMPORTANT: {pro} customers can take advantage of the assistance of the support team.
```
joedragons: Comment: I see we're saying this every pro feature now?=)

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
 
-The individual nodes are aware of their setup conditions such as:
+[[why-high-availability]]
+=== Why Cluster?
 
-* Memory availability and processor performance
-* Network connectivity performance
-* Characteristics of local storage in terms of size and performance
+The biggest draw to a high performing repository manager is of a self-maintaining system with improved disaster
```
joedragons: Minor: I THINK "of" is not needed here but not confident so made minor.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
-The individual nodes are aware of their setup conditions such as:
+[[why-high-availability]]
+=== Why Cluster?
 
-* Memory availability and processor performance
-* Network connectivity performance
-* Characteristics of local storage in terms of size and performance
+The biggest draw to a high performing repository manager is of a self-maintaining system with improved disaster 
+recovery features. Additionally backup support, high availability clusters can be used for load balancing and 
+failover initiative. Currently, you can only implement two nodes. But down the line, can expand the amount of
```
joedragons: "Additionally backup support, high availability clusters can be used for load balancing and failover initiative."

This sentence reads strangely to me.  Are backup support, 'load balancing' and failover initiatives draws?  Are you describing backup support as load balancing and failover initiative?  It's hard to even critique right without understanding what's being conveyed.
My guess is a word may be missing, or a comma...

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
-* Slow backup node for potential disaster recovery
-* Failover-node sitting in active standby mode
-* Load-balanced node
+[[high-availability-purpose]]
+=== What is the Purpose of High Availability?
 
-The knowledge of the component fabric about deployment topology and performance allows nodes to choose the 
-fastest route to fetch the component data from the fabric. The specified purpose of a node can configure 
-pre-fetching of components from the fabric even before a client tool requests it as a performance enhancement.
+The high availability feature provides a storage backend for two {nxrm} instances co-located in the same data
```
joedragons: Minor: "co-located" may be appropriate but colocate is a word, so I don't think the - is needed.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
-* Failover-node sitting in active standby mode
-* Load-balanced node
+[[high-availability-purpose]]
+=== What is the Purpose of High Availability?
 
-The knowledge of the component fabric about deployment topology and performance allows nodes to choose the 
-fastest route to fetch the component data from the fabric. The specified purpose of a node can configure 
-pre-fetching of components from the fabric even before a client tool requests it as a performance enhancement.
+The high availability feature provides a storage backend for two {nxrm} instances co-located in the same data 
+center.
```
joedragons: Q: Does it need to be the same center?  So NY and Asia cannot connect like Brian's chat?  (yet)

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
 
-[[ha-config]]
-=== HA Configuration and Startup
+High availability implementations are made of servers, joined together but a common directory. The second cluster
```
joedragons: I think a word may be missing before common directory...like "but using a common directory"?

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
-The knowledge of the component fabric about deployment topology and performance allows nodes to choose the 
-fastest route to fetch the component data from the fabric. The specified purpose of a node can configure 
-pre-fetching of components from the fabric even before a client tool requests it as a performance enhancement.
+The high availability feature provides a storage backend for two {nxrm} instances co-located in the same data 
+center.
 
-[[ha-config]]
-=== HA Configuration and Startup
+High availability implementations are made of servers, joined together but a common directory. The second cluster 
+is implemented to achieve synchronization between both enabled clusters. The clusters (or servers, or nodes)
```
joedragons: Ok, so this answers my Q above, however, new Q is are clusters, servers and nodes all synonymous as you say=)

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
-=== HA Configuration and Startup
+High availability implementations are made of servers, joined together but a common directory. The second cluster 
+is implemented to achieve synchronization between both enabled clusters. The clusters (or servers, or nodes) 
+communicate with each other to provide uptime and the assurance that if one goes down a duplicated node can be 
+used.
 
-{nxrm} installation consists of very few simple steps:
+Instead of manually managing your infrastructure, processes, and tools associated with numerous repository 
+manager installations, {pro} high availability feature provides a replicated storage system that eliminates 
+complexity and offers a simple and cost effective way to provision, configure, manage, and maintain multiple
```
joedragons: Minor: I assume by cost effective you mean resources not money.  You might just say that (resource effective).  Don't feel super strongly but in my social circles cost effective usually means money.  If this does mean money I think it's awkward.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
+
+This awareness also translates to increased efficiency, scalability, and control, allowing nodes to automatically 
+choose the fastest route to intercept component data between the two nodes. With additional customization, these 
+purpose-driven nodes can configure prefetching of components even before a client tool requests it, leading to 
+greater level of performance.
+
+[[high-availability-expectations]]
+=== Requirements for Clustering in {pro}
+
+{pro} is flexible, and can lend itself to a variety of different system architectures. So, before enabling this
```
joedragons: Minor (comma drama): "flexible, and" I don't feel commas are necessary with 2 options...just an additional char.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
+purpose-driven nodes can configure prefetching of components even before a client tool requests it, leading to 
+greater level of performance.
+
+[[high-availability-expectations]]
+=== Requirements for Clustering in {pro}
+
+{pro} is flexible, and can lend itself to a variety of different system architectures. So, before enabling this 
+specific feature, determine which option is best for your existing architecture and system configuration before 
+installing. There are three supported infrastructures you can consider to facilitate high availability between 
+your repository managers. You can install them on two:
```
joedragons: Q: Can this mix and match or does it need to be 2 of the same.  I believe it's mix and match but honestly not sure.  If so, you may want to rephrase slightly to say that or even make a seperate (folllowing) sentence.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
+
+- separate physical servers
+- virtual machines
+- Docker instances
+
+Your should recognize any challenges you may encounter when planning high availability implementation. You also
```
joedragons: Typo: "Your" -> You

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
+- virtual machines
+- Docker instances
+
+Your should recognize any challenges you may encounter when planning high availability implementation. You also 
+need to be mindful of cluster performance, so your repository manager can remain at its most optimal speed.
+
+Ultimately, the rule of thumb is you ensure the two nodes utilized share a file system for blob storage and that 
+they exist on the same network. Subsequently, the key goals for enabling high availability are as follows:
+
+- <<high-availability-install,Installing a secondary version>> of {pro}.
```
joedragons: To me "version" is like 2.13 vs 2.14.  Suggest "instance".

I am curious if anyone else feels "secondary" is weird.  I can't explain why, just doesn't seem totally right to me.  I would suggest "second" instead but it's pretty much saying the same thing.  Maybe the inference that one is primary (vs secondary) bothers me?  I wouldn't change anything if it's just me (or that's whats specifically trying to be said=)).

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,130 @@
...
+- <<high-availability-configuration,Configuring the second binary>> as its own independent server.
+- <<high-availability-enable,Activating the clusters>> from the file `org.sonatype.nexus.cfg`.
+- <<high-availability-startup,Starting>> and running the service upon activation.
```
joedragons: Minor: While this is true, I am not feeling strongly this is a "key goal".  I see there is a 19.4.5 that is not included in the key goals, I think this can be left off as well.

chapter-high-availability.asciidoc
dulanism: Agreed. Updated.

chapter-high-availability.asciidoc
dulanism: Clarifying with this revision with something like:

> Usage of clusters set the standard for automating the process of continuous uptime, relieving teams of manual recovery techniques...

chapter-high-availability.asciidoc
dulanism: I guess it'll be obvious, from the the {inrmonly} macro. Removing (if I understand your message).

chapter-high-availability.asciidoc
dulanism: A: I wanted to distinguish the clustering aspect (HA) from the subsequent feature in development (CF). But if that's not conveyed properly, please advise.

chapter-high-availability.asciidoc
dulanism: Re-writing to distinguish... since they are clusters, servers and nodes are more taxonomical than synonymous.

chapter-high-availability.asciidoc
dulanism: Re-wrote the section above (currently Why Cluster, but subject to change). Saving money and resources can imply the same goal when talking about the "why". I struck the wording, and worked into the "why" section because it's a better fit, imo.

chapter-high-availability.asciidoc
dulanism: Breaking out Managing Nodes and Storage to its own section

chapter-high-availability.asciidoc
joedragons: Ah okie=)  Thanks for the clarify.

chapter-high-availability.asciidoc
joedragons: No sorry, what I meant was I believe we put this in the Upgrade documentation as well.  Basically it seems like we're constantly reminding people that pro people can get support.  I think someone (else) balked at it there.  It seemed apt there to me but with that fresh in mind now seems like we're putting it everywhere.

chapter-high-availability.asciidoc
joedragons: Honestly, the lines are grey for me.  So it could just be me not understanding the difference that's confusing and not what you wrote.  Maybe see what someone else says on review.

chapter-high-availability.asciidoc
dulanism: I took it out.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,125 @@
...
-[[ha-config]]
-=== HA Configuration and Startup
+Instead of manual adjustments to your infrastructure, processes, and tools associated with your repository 
+manager installations, the high availability feature provides a replicated storage system. This eliminates 
+complexity and offers a simpler way to provision, configure, manage, and maintain multiple repositories.
 
-{nxrm} installation consists of very few simple steps:
+This feature is enabled across your nodes via link:https://hazelcast.com/[Hazelcast]. Hazelcast provides an 
+active/active approach to server clustering. It comes equipped with an interface for distributed data structures 
+and other aspects of in-memory computing.
```
joedragons: Q: Is the Hazelcast interface something we're touting to our users?  I feel like that's an internal detail, this might be worth checking on.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,125 @@
...
+scalability, and control, allowing the nodes to automatically choose the fastest route to intercept component 
+data between the two nodes. With additional customization, these purpose-driven nodes can configure prefetching 
+of components even before a client tool requests it, leading to greater level of performance.
+
+[[high-availability-expectations]]
+=== Requirements for Clustering in {pro}
+
+{pro} is flexible and can lend itself to a variety of different system architectures. So, before enabling this 
+specific feature, determine which option is best for your existing architecture and system configuration before 
+installing. There are three supported infrastructures you can consider to facilitate high availability between
```
joedragons: Minor: To me, "consider" makes it sound like this is instruction for someone who is setting up HA but has no servers up already.  Maybe this can be generalized?

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,125 @@
...
+Each node should have its own independent non-conflicting OrientDB database.
 
-In each distinct folders, unpackage the `tar.gz`.
-In the `nexus` folder, edit the `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file. Change the `application-port={port 
-number}` in your first instance of {nxrm}. Then remove the `#` before `nexus.clustered = true`, enabling 
-Hazelcast usage on start. Go to you the second `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file and repeat the same 
-steps.
+For example, if you have two nodes -- a default called "node-a" and a new node called "node-b" -- and "node-a's"  
+blob store directory located at `/foo/bar/blobs/foobar`. "Node-b" should be mounted to the same file system 
+location.
```
joedragons: Maybe I am confused by terminology (or misreading) but in the previous paragraph you say don't configure nodes to share a data directory but this last sentence seems to infer to do it.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,125 @@
...
+repository manager port.
+
+[[high-availability-enable]]
+==== Enabling the Nodes
+
+Follow these steps to enable the high availability feature:
+
+. In your existing repository manager, open the `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file.
+. Remove the `#` before `nexus.clustered = true` to enable Hazelcast usage on start. 
+. Go to the new repository manager and repeat steps 1 and 2, to enable the new instance.
```
joedragons: > to enable the new instance.

Suggest "to enable (High Availablility) in the new instance." (High Availability being at your discression but definiately suggest adding "in")

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,125 @@
...
-On startup in review your log. You should see connection messages via Hazelcast similar to this, below:
+After you extract your newly downloaded `tar.gz` of {pro}, you need to establish communication between your two 
+nodes. To do this, you must distinguish the `application-port` of your existing repository manager's 
+`org.sonatype.nexus.cfg` file, from the new instance you download. Go to <<config-http-port>> to identify the new 
+repository manager port.
+
+[[high-availability-enable]]
+==== Enabling the Nodes
+
+Follow these steps to enable the high availability feature:
```
joedragons: Since it is inferred that one of the managers exists, should you call out that it should be shut down and/or have a mention that they connect on restart?

chapter-high-availability.asciidoc
dulanism: I rewrote the paragraph to mention your options with HA.

Addendum: I see new comments below to the same section. I'll work thru.

joedragons:
This document speaks to folks that are setting up a new server to add to the cluster.  Unless we don't want to speak of or support the scenario, I think it's worth a note somewhere (probably before the end) that if you have an existing server with "conflicting" configurations that the first node settings will be copied over (the second nodes settings will be gone).
In a new server, this doesn't matter and is cool=)  With two existing servers I think it's very important to decide which is "master" or make it such that you make "master" not overwrite anything in the addition nodes.

chapter-high-availability.asciidoc
dulanism: From the chat room: seems reasonable to mention... but the language is somewhat marketing-ish. I'll downplay the words that upsell.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,125 @@
...
+same shared storage. You need to share the nodes the directory location of your 
+<<admin-repository-blobstores,blob store>>. You shouldn't configure your nodes to share an entire data directory. 
+Each node should have its own independent non-conflicting OrientDB database.
 
-In each distinct folders, unpackage the `tar.gz`.
-In the `nexus` folder, edit the `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file. Change the `application-port={port 
-number}` in your first instance of {nxrm}. Then remove the `#` before `nexus.clustered = true`, enabling 
-Hazelcast usage on start. Go to you the second `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file and repeat the same 
-steps.
+For example, if you have two nodes -- a default called "node-a" and a new node called "node-b" -- and "node-a's"
```
dulanism: Originally not sure, but I referred to this http://www.grammarunderground.com/plurals-and-possessives-of-movie-titles-in-quote-marks.html

joedragons: "The bad news is that the style guides don’t say. The good news is that the style guides don’t say." thanks for the info (and laugh)

chapter-high-availability.asciidoc
dulanism: Updated the language.

chapter-high-availability.asciidoc
dulanism: I re-worked the first paragraph to speak more to the assumption that the user has one server already set up. He's the variation

>To make high availability work, you must have an existing repository manager already 
configured to start up. Then, ensure that both network configuration and blob storage are identical to the second repository manager that you'll need to install. Enabling the second repository manager is the minimum requirement to allow redundancy between the nodes.

joedragons: Thoughts (can also add to later review, if desired):
* "and blob storage are identical" - I thought later you mentioned don't share the same blob storage; if I'm remembering right I'd be careful with wording here.
* Minor: "to start up" caused me pause initially but on reread, I think I'm ok.  My initial thought FYI was "how else would you have NXRM configured" but it could mean for HA or something else so I think ok.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
 
-[[ha-introduction]]
+[[high-availability-introduction]]
 === Introduction
 
-High availability for {nxrm} sets the standard for a continuously operational application to reduce points of 
-failure among features and components within your repository manager. The component fabric in {pro}, which 
-enables high availability, acts as a private binary cloud storage and backend for all your components.
+{nxrm}'s high availability feature provides clustering for two separate instances of the repository manager. This 
+is used to create shared storage for a more fail-safe backend in your repository manager, minimizing downtime and
```
bentmann: "used to create shared storage", I can't wrap my head around this. Our impl currently requires shared storage for the blob store, I fail to see how we "create" it. What exactly are we trying to convey here?

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
-failure among features and components within your repository manager. The component fabric in {pro}, which 
-enables high availability, acts as a private binary cloud storage and backend for all your components.
+{nxrm}'s high availability feature provides clustering for two separate instances of the repository manager. This 
+is used to create shared storage for a more fail-safe backend in your repository manager, minimizing downtime and 
+reducing points of system failure.
 
-Component Fabric is an integral part of {pro} reduces the complexity of deployment and infrastructure, while 
-simultaneously improving performance, scalability, and user experience. You can activate as many repository 
-managers connected to your component fabric as you like and move your infrastructure to the next level.
+This chapter covers the benefits and steps needed to implement high availability. Usage of clusters set the
```
bentmann: "Usage ... set**s** the standard", no?

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
-simultaneously improving performance, scalability, and user experience. You can activate as many repository 
-managers connected to your component fabric as you like and move your infrastructure to the next level.
+This chapter covers the benefits and steps needed to implement high availability. Usage of clusters set the 
+standard for automating the process of continuous uptime, relieving teams of manual recovery techniques, and 
+creating a binary storage for the entire suite of features and components available in your repository managers.
 
-[[ha-node-performance]]
-=== Node Performance Considerations
+[[why-high-availability]]
+=== Why Cluster?
```
bentmann: Confused why we have to spend words on this... More seriously, I wouldn't mind a refresher what goal our documentation is meant to serve. If to assist marketing, then sure, bring on the yada yada, if to help use our product, this strikes me as a distriction because the only question there is to answer "How do I this or that?".

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
 
-The individual nodes are aware of their setup conditions such as:
+The biggest draw to a high performing, well-synchronized backend is a self-maintaining system with improved 
+disaster recovery features. Downtime can be costly to organizations that host a multitude of applications.
+Therefore, an automated backup plan with distributed clusters decrease the chances of data loss.
```
bentmann: "backup plan ... decrease**s**"

"disributed clusters", two nodes form a one cluster, those nodes are distributed (on different hosts) but there's just on cluster, so "distributed nodes" seems more appropriate use of the terminology to me.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
-fastest route to fetch the component data from the fabric. The specified purpose of a node can configure 
-pre-fetching of components from the fabric even before a client tool requests it as a performance enhancement.
+This feature is enabled across your nodes via link:https://hazelcast.com/[Hazelcast], which provides in-memory 
+computing for active data and active backup. So, instead of making manual adjustments to your servers' 
+infrastructure, processes, and tools, high availability via Hazelcast simplifies the configuration needed to 
+maintain multiple repositories.
 
-[[ha-config]]
-=== HA Configuration and Startup
+When enabling active/active high availability in your two repository manager instances, they act as individual
```
bentmann: "individual nodes", another thing I don't get. Usually, I would think that an active/active HA setup is to have two nodes act transparently as one instance, i.e. the nodes are indistinguishable/interchangeable to the end user.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
 
-{nxrm} installation consists of very few simple steps:
+- Memory availability and processor performance
+- Network connectivity performance
+- Characteristics of local storage in terms of size and performance
 
-* Install binary on server
-* Register and authenticate with the component fabric feature
-* Declare minimal local node configuration such as storage and node performance (e.g. prefetch rules)
+This node-driven awareness within your clustered repository manager translates to increased efficiency,
```
bentmann: "increased efficiency", define "efficiency". HA is using twice the nodes/resources for the same task of managing components. It helps scalability and availability but efficiency?

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
-{nxrm} installation consists of very few simple steps:
+- Memory availability and processor performance
+- Network connectivity performance
+- Characteristics of local storage in terms of size and performance
 
-* Install binary on server
-* Register and authenticate with the component fabric feature
-* Declare minimal local node configuration such as storage and node performance (e.g. prefetch rules)
+This node-driven awareness within your clustered repository manager translates to increased efficiency, 
+scalability, and control, allowing the nodes to automatically choose the fastest route to intercept component
```
bentmann: "choose the fastest route to intercept component data", the fabric in the future maybe, our current HA impl, not really.

You put a load balancer in front of your two NX nodes, the balancer forwards requests to either of them on his discretion and the chosen NX node carries out the request. Given the blob store is to be shared storage and the component metadata replicated, the "route" to components and their metadata is of the same length for both nodes, no fancy "fatest route" selection involved/implemented

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+- Memory availability and processor performance
+- Network connectivity performance
+- Characteristics of local storage in terms of size and performance
 
-* Install binary on server
-* Register and authenticate with the component fabric feature
-* Declare minimal local node configuration such as storage and node performance (e.g. prefetch rules)
+This node-driven awareness within your clustered repository manager translates to increased efficiency, 
+scalability, and control, allowing the nodes to automatically choose the fastest route to intercept component 
+data between the two nodes. With additional customization, these purpose-driven nodes can configure prefetching
```
bentmann: There's no prefetching in our current HA impl, components are to be stored in a shared filesytem and hence don't need to be transferred across.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+of components even before a client tool requests it, leading to greater level of performance.
 
-Any further work is done automatically by the interactions of the new repository manager with the data and 
-configuration in the component fabric.
+[[high-availability-expectations]]
+=== Requirements for Clustering in {pro}
+
+To make high availability work, you must have an existing repository manager already configured to start up. 
+Then, ensure that your existing repository manager has the same network configuration as the second repository 
+manager that you'll need to install. The two repository managers must run on different ports. Enabling the second
```
bentmann: "different ports", generally no, it depends. If one runs the RM instances on different hosts (which I would assume is the preferred setup to get the most out of HA for uptime), they can use the same port.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+repository manager is the minimum requirement to allow redundancy between the nodes.
+
+There are three supported infrastructures you can consider to facilitate high availability between 
+your repository managers. You can install them across a mix of two:
+
+- physical servers
+- virtual machines
+- Docker instances
+
+You should to be mindful of server performance, so your repository manager nodes can remain at its most optimal
```
bentmann: "should ~~to~~ be mindful"

chapter-high-availability.asciidoc
dulanism: @bentmann You're absolutely right about the intent. Here's the refresher: https://docs.sonatype.com/display/ProdMgmt/A+guide+to+writing+functional+user+documentation. I think with the heavy push to get EA products out the door, the basic steps around _simply writing the steps_ has fallen a little by the wayside. I appreciate your honestly. I'll scale this back in tone.

chapter-high-availability.asciidoc
dulanism: The direct object was mistaken for the subject ("plan"). I'll update, along with the proper terminology.

chapter-high-availability.asciidoc
dulanism: Ironically, I will _efficiently_ throw away that word. I retrospect, it's an empty term.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+
+There are three supported infrastructures you can consider to facilitate high availability between 
+your repository managers. You can install them across a mix of two:
+
+- physical servers
+- virtual machines
+- Docker instances
+
+You should to be mindful of server performance, so your repository manager nodes can remain at its most optimal 
+speed. Ultimately, the rule of thumb is you ensure the two nodes utilized share a file system for blob storage
```
bentmann: "rule of thumb", this implies there are other rules/ways to do this. There aren't. Our HA impl currently requires a shared filesystem for the blob storage, period.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+- Docker instances
+
+You should to be mindful of server performance, so your repository manager nodes can remain at its most optimal 
+speed. Ultimately, the rule of thumb is you ensure the two nodes utilized share a file system for blob storage 
+and that they exist on the same network. Subsequently, to enable your nodes for high availability follow these 
+steps:
+
+- <<high-availability-install,Install a second instance>> of {pro}.
+- <<high-availability-configuration,Configure the second binary>> as its own independent server.
+- <<high-availability-enable,Activate the clusters>> from the file `org.sonatype.nexus.cfg`.
```
bentmann: "Activate the cluster" (singular) or "Activate the clustering".

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+- <<high-availability-startup,Start>> and run the service after configuration.
+
+[[high-availability-install]]
+==== Installing a New Binary
+
+Assuming you have an existing repository manager in place, download a brand new version of {pro}.
```
bentmann: "download a brand new version", this can mislead users to download a different/incompatible version. They should get the binary for the version they currently run.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+[[high-availability-install]]
+==== Installing a New Binary
+
+Assuming you have an existing repository manager in place, download a brand new version of {pro}. 
+Install this new binary in the same directory as your existing repository manager. You can refer to 
+<<installation-archive>> for details on how to download and install.
+
+As you install your new repository manager, be sure to mount it in the same directory as the existing repository 
+manager. To properly configure a high-availability cluster the hosts in the cluster must all have access to the 
+same shared storage. You need to share the nodes the directory location of your
```
bentmann: "share the nodes the directory", typo or a construct I can't decipher.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+- <<high-availability-startup,Start>> and run the service after configuration.
+
+[[high-availability-install]]
+==== Installing a New Binary
+
+Assuming you have an existing repository manager in place, download a brand new version of {pro}. 
+Install this new binary in the same directory as your existing repository manager. You can refer to 
+<<installation-archive>> for details on how to download and install.
+
+As you install your new repository manager, be sure to mount it in the same directory as the existing repository
```
bentmann: I vote to strike the first two sentence of this paragraph, they are too vague and give rise to misinterpretation. "same directory", which directory? "shared storage", sharing what? The 3rd sentence finally gets to the point.

chapter-high-availability.asciidoc
joedragons: I suspect we're trying to convey that we're sharing (clustering) between the nodes.  That's how I read it when I reviewed.  Is "used to create storage sharing" any more clear?=)

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+
+Assuming you have an existing repository manager in place, download a brand new version of {pro}. 
+Install this new binary in the same directory as your existing repository manager. You can refer to 
+<<installation-archive>> for details on how to download and install.
+
+As you install your new repository manager, be sure to mount it in the same directory as the existing repository 
+manager. To properly configure a high-availability cluster the hosts in the cluster must all have access to the 
+same shared storage. You need to share the nodes the directory location of your 
+<<admin-repository-blobstores,blob store>>. You shouldn't configure your nodes to share an entire data directory. 
+Each node should have its own independent non-conflicting OrientDB database.
```
bentmann: Independent data dirs (with the exception of blob store) are crucial for proper operation. So crucial that I feel this deserves a clearer callout by explicitly listing it above among the steps in line 72 ff., e..g "Configure shared blob storage".

Apart from being crucial, this step could also be non-trivial for users with existing RM instances that now have to figure out how to share whatever blobs currently exist, i.e. users might need to re-organize their filesystem in preparation for HA.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
 
-To start your configuration, download a binary as explained in <<install-sect-downloading>>. Create two distinct 
-folders to store duplicate instances of the repository manager. For example, you can create two directories 
-`nexus` and `nexus2`. However, you have not yet installed {pro}, review the installation steps from the 
-beginning, as explained in <<install-introduction>>.
+For example, if you have two nodes -- a default called "node-a" and a new node called "node-b" -- and "node-a's"  
+blob store directory located at `/foo/bar/blobs/foobar`. "Node-b" should be mounted to the same file system
```
bentmann: "should be mounted to the same file system location", I argue this is vague enough to have folks get it wrong and show up at support (in part due to "/foo/bar/" having no resemblence with an actual installation and hence not supporting comprehension).

The two nodes need to share the directory `sonatype-work/nexus3/data/blobs` (such that any blob store created beneath that directory is visible to both nodes)

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
-In the `nexus` folder, edit the `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file. Change the `application-port={port 
-number}` in your first instance of {nxrm}. Then remove the `#` before `nexus.clustered = true`, enabling 
-Hazelcast usage on start. Go to you the second `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file and repeat the same 
-steps.
+[[high-availability-configuration]]
+==== Configuring the Port of the Second Binary
 
-TIP:: Log in to your accouunt and go to the 'Administration' menu. Click 'Nodes' under the 'System' category.  
-You should see two nodes. Each machine should have a different node enabled as "true".
+After you extract your newly downloaded `tar.gz` of {pro}, you need to establish communication between your two
```
bentmann: "tar.gz", maybe I downloaded the "zip". Let's not confuse people by implying only one of the two archive formats has to be downloaded for HA, "newly downloaded archive/package/binary/distribution of"

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
-number}` in your first instance of {nxrm}. Then remove the `#` before `nexus.clustered = true`, enabling 
-Hazelcast usage on start. Go to you the second `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file and repeat the same 
-steps.
+[[high-availability-configuration]]
+==== Configuring the Port of the Second Binary
 
-TIP:: Log in to your accouunt and go to the 'Administration' menu. Click 'Nodes' under the 'System' category.  
-You should see two nodes. Each machine should have a different node enabled as "true".
+After you extract your newly downloaded `tar.gz` of {pro}, you need to establish communication between your two 
+nodes. To do this, you must distinguish the `application-port` of your existing repository manager's
```
bentmann: As mentioned above, this is not a requirement and depends on the setup. Ports are per host, so if one runs servers on different hosts, they can use the same port on each host.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+repository manager port.
 
-On startup in review your log. You should see connection messages via Hazelcast similar to this, below:
+[[high-availability-enable]]
+==== Enabling the Nodes
+
+Follow these steps to enable the high availability feature:
+
+. In your existing repository manager, open the `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file.
+. Remove the `#` before +nexus.clustered=true+ to enable Hazelcast usage on start.
```
bentmann: Hazelcast is part of the implementation, let's keep references to it to a minimum. "to enable clustering/high availability"

chapter-high-availability.asciidoc
bradbeck: How about "{nxrm}'s high availability feature provides a more fail-safe repository manager, minimizing downtime and reducing points of system failure."

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,123 @@
...
+[[high-availability-enable]]
+==== Enabling the Nodes
+
+Follow these steps to enable the high availability feature:
+
+. In your existing repository manager, open the `$NEXUS_HOME/etc/org.sonatype.nexus.cfg` file.
+. Remove the `#` before +nexus.clustered=true+ to enable Hazelcast usage on start. 
+. Go to the new repository manager and repeat steps 1 and 2, to enable the new instance.
+
+Any further work is done automatically by the interactions of the new repository manager with the data and
```
bentmann: We might to mention that auto detection of cluster members relies on multicast. In networks where multicast fails (e.g. blocked by firewall), users will need to customize `etc/hazelcast.xml`. Unsure what exactly we want to document about that, e.g. just link out to http://docs.hazelcast.org/docs/3.6/manual/html-single/index.html#discovering-cluster-members or selectively document something ourselves. Suggest consulting with support and PO teams.

chapter-high-availability.asciidoc
```diff
@@ -1,66 +1,125 @@
...
+. Go to the new repository manager and repeat steps 1 and 2, to enable the new instance.
+
+Any further work is done automatically by the interactions of the new repository manager with the data and 
+configuration synchronized with the high availability feature.
+
+[[high-availability-startup]]
+==== Starting the High Availability Service
+
+At this point, you have already configured a new, distinct repository manager to run as a backup for the original 
+instance. So, start both repository managers, by entering and executing `./bin/nexus run` in the command line. On
```
bentmann: At this point in the doc, it should not be necessary to tell the reader how to start their instance, that should have been covered elsewhere already.

bentmann: "start both repository managers", we should point out that the instances should be started one after the other, i.e. bring the first node to full service, then start the second one. AFAIR, there can be issues when both are started concurrently.

chapter-high-availability.asciidoc
```diff
@@ -71,11 +128,17 @@ Members [2] {
 }
 ----
 
-[[ha-scheduled-tasks]]
-=== Scheduled Tasks
+[[high-availability-verify]]
+=== Verifying Synchronization
+
+At runtime, when you're log in as an administrator on one node, the other node automatically pairs the
```
bentmann: "log**ged** in"

"automatically pairs the administrator to the second node", what does that mean?

chapter-high-availability.asciidoc
```diff
@@ -71,11 +128,17 @@ Members [2] {
...
-[[ha-scheduled-tasks]]
-=== Scheduled Tasks
+[[high-availability-verify]]
+=== Verifying Synchronization
+
+At runtime, when you're log in as an administrator on one node, the other node automatically pairs the 
+administrator to the second node. The {nxrm} interface allows you to see the nodes paired.
+
+To verify this connection, go to the 'Nodes' screen, under 'System' located in the 'Administration' menu. This 
+screen provides details of the node, detected as the backup server.
```
bentmann: "This screen provides details of the nodes in the cluster". There is no backup server, all nodes are equal, i.e. active/active.

chapter-high-availability.asciidoc
```diff
@@ -71,11 +128,17 @@ Members [2] {
...
+
+At runtime, when you're log in as an administrator on one node, the other node automatically pairs the 
+administrator to the second node. The {nxrm} interface allows you to see the nodes paired.
+
+To verify this connection, go to the 'Nodes' screen, under 'System' located in the 'Administration' menu. This 
+screen provides details of the node, detected as the backup server. 
 
-TBD
+For example, when you look at the row representing "node-a", you can see its 'UUID' (a unique identifier) 
+as well as the label 'true'. The 'true' statement, in the 'Local' column, indicates the "node-b" is synchronized
```
bentmann: local=true means the request backing the UI came from that node. That stuff about "synchronized for backup" doesn't make any sense to me.

I'd even go further and stroke any mention of the local column. If you hit the cluster via a load balancer that sends requests to either of the two nodes, one can see the local column fluctuate, i.e. sometimes node-a and sometimes node-b will be shown as "local", all depending on where the load balancer sent the request to. I don't think we want to describe that technicality at this point.

chapter-high-availability.asciidoc
bentmann: > Is "used to create storage sharing" any more clear?=)

Maybe, I wonder whether just "This is used to create a more fail-safe backend ..." misses the point.

chapter-high-availability.asciidoc
joedragons: I haven't had any issues with concurrent starting FWIW.  I believe I had an issue early on which was resolved.  I'm neutral if you want to change tho, I see no harm in telling them to wait.

bentmann: Maybe I misremember but I thought @bradbeck once reported issues when setting up the Docker-based HA test framework that bad things happened when the nodes boot concurrently.

joedragons: @bentmann I recall this as well, but I think he (or someone) may have fixed the issues.  I cannot find any tickets so I'd be interested into his input.  I do not wait til one node is up before starting another in my testing however. cc/ @bradbeck

chapter-high-availability.asciidoc
dulanism: I think there is a wholesome validity to stating "Why Cluster" (if that's the right terminology). The fact that a reader could start reading anywhere in this book, adding a prelude to motivate them to use the feature is important. No knowledge is wasted. That's my motto. Of course, the language can be massaged to remove the fluff. But I there goal here is to draw readers in, not the reverse. My 2 cents.

dulanism:
I think there is a wholesome validity to declaring such topics as "Why Cluster" (if that's the right terminology). The fact that a reader could start reading anywhere in this book, adding a prelude to motivate them to use the feature is important. No knowledge is wasted. That's my motto. Of course, the language can be massaged to remove the fluff. But I there goal here is to draw readers in, not the reverse. My 2 cents.

chapter-high-availability.asciidoc
dulanism: Now that I've had time to better think this thru, I'll strike Why Cluster.

bentmann: For me, the key reason to provide some the-why/motivation for a feature is whether it's unique to the product or not. For stuff that product-specific and mostly unknown to the user/reader, a short motivation to help their comprehension makes sense. But HA isn't specific to our product, it's a general feature found in many enterprise products and the reader already knows about it and why they want to use it doesn't need explaining.

chapter-high-availability.asciidoc
dulanism: Thanks. On the strength of your feedback, a better way to get this across is:

>With active/active high availability in your two repository manager instances, you enable two interchangeable nodes that act transparently as one instance.

chapter-high-availability.asciidoc
dulanism: Better to removed the last sentence entirely.

chapter-high-availability.asciidoc
dulanism: Removing. But the tip about getting the most out HA for uptime might be worth a little note in the configuration section.

chapter-high-availability.asciidoc
dulanism: May require a little bit of reading. I might just create new ticket for a new section/line

PR #233, opened 2016-09-19
Removed documentation about installer

mosabua:
Since its deprecated until further changes occur, NEXUS-10914

Actually deleted content instead of commenting out since we dont know when (and if) it will come back. Worst case we pull this out of the git history or rewrite as agreed with @dulanism .


mosabua:
Merging and publishing this. Will also push to 3.1 branch.

PR #234, opened 2016-10-19
Use $data-dir/etc/nexus.properties for configuration

bradbeck:


bradbeck:
Sorry, wrong repo. See: https://github.com/sonatype/nexus-book-internal/pull/14

PR #235, opened 2016-10-24
Fixed path to Nexus binary in the systemd service

wikier:


dulanism:
@wikier Just making sure that I follow you. Did you create the `nexus.service` file and add it to `/etc/systemd/system` before running the service from `/opt/nexus/bin`?

wikier:
Yes, i did @dulanism. But then I got the error that {{ExecStart}} was invalid, that's the purpose of this fix to the documentation. The change has been used, and it's working, on my dev environment.

dulanism:
Thank you, @wikier. The book is updated.

wikier:
Perfect. Thanks, @dulanism.

PR #236, opened 2016-10-25
Scripting - Fix Typo in Sample pom.xml

busches:
In the sample `pom.xml` file the `nx.version` property is defined with a period, but referenced as `nx-version`. Make them consistent so the sample `pom.xml` file works without any modifications.

The master branch appeared to still be Nexus Book version 2. If this needs to be applied to a different branch, let me know.

### DIFF
--- a/chapter-scripting.asciidoc
+++ b/chapter-scripting.asciidoc
@@ -37,7 +37,7 @@ available in the https://github.com/sonatype/nexus-book-examples/tree/nexus-3.0.
   <version>1.0-SNAPSHOT</version>
 
   <properties>
-    <nx.version>3.0.0-03</nx.version>
+    <nx-version>3.0.0-03</nx-version>
   </properties>
   <dependencies>
     <dependency>


joedragons:
+1

mosabua:
-1 ... the example uses nx-version. See https://github.com/sonatype/nexus-book-examples/blob/nexus-3.0.x/scripting/nexus-script-example/pom.xml

busches:
@mosabua embedded in the book itself: https://books.sonatype.com/nexus-book/3.0/reference/scripting.html No where in the book is there a link to the actual pom file. I can update it to match the pom file to be `nx-version` to be consistent. I'm care about having the example actually work.

busches:
@mosabua FYI I repushed the changes to make it `nx-version` yesterday.

mosabua:
The example does work... thats how I demoed it in the video and use it regularly when scripting. I will update this info in the branches and republish the docs.

mosabua:
I fixed this now internally in all branches applicable and pushed it out. Will push the docs to the site in a sec too. fyi @dulanism @joedragons

mosabua:
Fix is now live at https://books.sonatype.com/nexus-book/reference3/scripting.html

PR #239, opened 2017-07-21
Nexus 2.9.x

yuvamjain:


joedragons:
Can you explain the intention here?  Our different books are distinct on purpose in case folks want to review how the older versions worked.

